PR #4, opened 2015-05-18
Address cursor and editor context deprecations.

rgbkrk:
Knocked out a few of the deprecations from the Atom API I noticed with the deprecation cop.


willwhitney:
Looks great — I hadn't noticed. Thanks!

PR #14, opened 2015-05-21
Bump zmq.

rgbkrk:
Most importantly, this also bumps nan, to ensure Hydrogen builds correctly with iojs (ahem, the future of nodejs), which includes related apps like Electron and the coming release of Atom.


willwhitney:
Good call, thanks!

PR #19, opened 2015-05-22
Paths ported from jupyter_core

rgbkrk:
This is a port of https://github.com/jupyter/jupyter_core while also including `~/.ipython/` as part of the path search. This ensures that the default directories that the Jupyter consoles and notebook searches through are the same for Hydrogen.

The core function here is `jupyterPath`, which gets used like so:

```
coffee> jupyterPath('kernels')
[ '/Users/kyle6475/Library/Jupyter/kernels',
  '/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/share/jupyter/kernels',
  '/usr/local/share/jupyter/kernels',
  '/usr/share/jupyter/kernels',
  '/Users/kyle6475/.ipython/kernels' ]
coffee> jupyterPath('runtime')
[ '/Users/kyle6475/Library/Jupyter/runtime',
  '/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/share/jupyter/runtime',
  '/usr/local/share/jupyter/runtime',
  '/usr/share/jupyter/runtime',
  '/Users/kyle6475/.ipython/runtime' ]
```

This will also be useful for picking up what running kernels are available for the notebook, another console (use Hydrogen on the side of a console session!), etc.


rgbkrk:
Nevermind, it seems to be hanging off current master too. Not sure what's up.

rgbkrk:
Totally my fault. There was a problem on a branch I was running in dev mode that the IPython kernel was depending on. Working well. That's what I get for straddling two projects at once.

screenshot 2015-05-22 13 16 25

rgbkrk:
I'd also be open to making a package like this that can be used across packages, as I'd use this in side car.

rgbkrk:
I'm starting to add tests. In the process I tore out the boilerplate Atom specs for a new package (caused failures). Once this is done, it would be nice to enable Travis CI.

lib/paths.coffee
```diff
@@ -0,0 +1,144 @@
...
+
+  # Windows first for easier testing
+  if process.platform == 'win32'
+    appData = process.env['APPDATA']
+    if appData
+      return path.join appData, 'jupyter'
+    else
+      return path.join jupyterConfigDir(), 'data'
+
+  home = userHome()
```
willwhitney: Let's make this an `if... else if... else` instead of relying on the `return`s to be more explicit about the control flow

willwhitney:
This looks good, and it runs great for me! Have you been able to test it on Windows?

lib/paths.coffee
rgbkrk: Done.

rgbkrk:
Haven't been able to test this on Windows. I did a minimal amount of adding some mocking specs in (while tossing the skeleton code), but that doesn't say for sure how Windows goes.

willwhitney:
Given that our Windows support is already weak at best, I'll just ship it and see how it goes.

willwhitney:
@jzthree Just published this live. I know you've got Hydrogen running on Windows, so please let me know if this broke anything for you!

jzthree:
Hey I just tested. Unfortunately this broke Hydrogen:Run on Windows. The execSync call in paths.coffee failed.  I pasted the first few lines of error message here.

Error: EBADF: bad file descriptor, write
  at Error (native)
  at Object.fs.writeSync (fs.js:657:20)
  at SyncWriteStream.write (fs.js:1906:6)
  at execSync (child_process.js:1372:20)
  at Object.<anonymous> (C:\Users\Jian.WIN-MSIJZ\.atom\packages\Hydrogen\lib\paths.coffee:10:12)
  at Object.<anonymous> (C:\Users\Jian.WIN-MSIJZ\.atom\packages\Hydrogen\lib\paths.coffee:3:1)

rgbkrk:
Well now I wish I had a Windows box quite a bit... I'm surprised readSync doesn't work but perhaps it's not finding `python` on the path.

rgbkrk:
Hey @zooba, any ideas?

willwhitney:
@jzthree Just pushed an update that replaces `execSync` calls with `spawnSync`, which seems to not cause this "bad file descriptor, write" error in my tests. Mind running it when you get a chance?

jzthree:
Yes just updated and it all works now. Thanks!!

rgbkrk:
Thank you @willwhitney, sorry about that.

Issue #13, opened 2015-05-21
Windows installation fails


bilderbuchi:
I am on Windows 8, Atom 0.200.0, and the installation fails with
```
Installing “Hydrogen@0.3.3” failed.

> zmq@2.11.1 install C:\Users\buchner\AppData\Local\Temp\apm-install-dir-115421-4460-cnflq2\node_modules\Hydrogen\node_modules\zmq
> node-gyp rebuild


C:\Users\buchner\AppData\Local\Temp\apm-install-dir-115421-4460-cnflq2\node_modules\Hydrogen\node_modules\zmq>node "C:\Users\buchner\AppData\Local\atom\app-0.200.0\resources\app\apm\node_modules\npm\bin\node-gyp-bin\\..\..\node_modules\node-gyp\bin\node-gyp.js" rebuild 

gypnpm ERR! Windows_NT 6.2.9200
npm ERR! argv "C:\\Users\\buchner\\AppData\\Local\\atom\\app-0.200.0\\resources\\app\\apm\\bin\\\\node.exe" "C:\\Users\\buchner\\AppData\\Local\\atom\\app-0.200.0\\resources\\app\\apm\\node_modules\\npm\\bin\\npm-cli.js" "--globalconfig" "C:\\Users\\buchner\\.atom\\.apm\\.apmrc" "--userconfig" "C:\\Users\\buchner\\.atom\\.apmrc" "install" "C:\\Users\\buchner\\AppData\\Local\\Temp\\d-115421-4460-c5n6l2\\package.tgz" "--target=0.22.0" "--arch=ia32"
npm ERR! node v0.10.35
npm ERR! npm  v2.5.1
npm ERR! code ELIFECYCLE

npm ERR! zmq@2.11.1 install: `node-gyp rebuild`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the zmq@2.11.1 install script 'node-gyp rebuild'.
npm ERR! This is most likely a problem with the zmq package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     node-gyp rebuild
npm ERR! You can get their info via:
npm ERR!     npm owner ls zmq
npm ERR! There is likely additional logging output above.
```
Python 3.4.2, Ipython notebook 3.1.0, and zeromq 4.0.5 are installed via Anaconda/Miniconda.

speg:
I get the same error on OS X 10.10.3

```
[...]

PR #31, opened 2015-05-25
Level up windows

rgbkrk:
This adds new installation instructions in light of #13.

It also adds some keybindings. Note that I didn't add ctrl-alt-backspace because that's the hotkey for killing the X server in Linux. I wish we could make the hotkeys be `cmd-enter` and `ctrl-enter`, but hey.


rgbkrk:
Pictures for posterity:

screenshot 2015-05-25 09 31 04

screenshot 2015-05-25 09 41 15

Issue #28, opened 2015-05-24
investigate using fix-path


willwhitney:
[`fix-path`](https://github.com/sindresorhus/fix-path) may be able to make the path work when Atom is launched as an app. It won't fix problems with `node-gyp` during installation or update, though.

rgbkrk:
The one thing I'd want to be sure of is that fix-path won't override what's currently in the path (if someone is being explicit).

willwhitney:
Per my comment in #34, I think this is going to lead to bugs down the road and should be corrected in Atom instead.

PR #34, opened 2015-05-25
Use fix-path in kernel.coffee

rgbkrk:
Though I have an objection to forcing a new `PATH` on users who set an explicit one in launching Atom directly, this does make it easier for new users.

Addresses #28.

Couple notes:

* this will overwrite any PATH the user set before launching Atom at the command line `PATH=~/myvirtualenv`. That makes virtualenvs and conda environments a no go.
* fix-path forces some node modules on the front unconditionally (on OSX and Linux) if shell-path doesn't return a path back

However, I like the `sh -c echo $PATH` trick here. Wondering if there's a way to find out if Atom was launched via `launchctl`...

### DIFF
--- a/lib/kernel.coffee
+++ b/lib/kernel.coffee
@@ -5,6 +5,10 @@ _ = require 'lodash'
 child_process = require 'child_process'
 uuid = require 'uuid'
 
+fixPath = require 'fix-path'
+
+fixPath()
+
 StatusView = require './status-view'
 
 module.exports =
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
   },
   "dependencies": {
     "atom-space-pen-views": "^2.0.5",
+    "fix-path": "^1.1.0",
     "lodash": "^3.8.0",
     "uuid": "^2.0.1",
     "zmq": "^2.11.1"


rgbkrk:
With the realization of the PATH being overwritten for virtualenvs (or anything else that relies on PATH), I'm against adding this directly. With some `launchctl` checking we could roll with this.

rgbkrk:
Well, there's a way but it's hacky.

screenshot 2015-05-25 15 45 11

```node
cp.exec('launchctl list | grep ' + process.pid, function(err,stdout,stderr){console.log(stdout);})
```

willwhitney:
I'm not comfortable shipping a solution this hacky — it feels like a bug waiting to happen. Thanks for spending the time figuring this out!

I'll write this up as a bug against Atom. Doing this properly once seems like a clear win for the platform.

rgbkrk:
Definitely appropriate. I'll bet someone has posted the bug - there's got to be a way for Atom to detect being launched by launchctl and performing its own tomfoolery to get the `PATH`. There are *lots* of packages that need this (go-plus, atom-script, travis, terminal, the list goes on).

willwhitney:
Couldn't find the issue already opened, so I submitted #6956.

rgbkrk:
Went ahead and finished up the hacky solution earlier, figured I should post it anyway:

```CoffeeScript
if process.platform == "darwin"

  childProcess = require('child_process')
  shell = process.env.SHELL || '/bin/sh'

  childProcess.exec("launchctl list", launchCtlPathFix)

  launchCtlPathFix = (error, stdout) ->
    if error and error.code != 0
      # Could not find the Atom process within the launchctl list
      # Likely launched using the terminal. Do nothing
      console.log("Using your $PATH rather than trying to guess, based on launchctl error:")
      console.log(stderr)
      console.error(error)
      return

    if stdout
      # process.pid was in the list, make sure that it was a reasonable match
      output = stdout.trim()
      lines = output.split('\n')

      # Verify that at least one of the lines matches the exact pattern of
      # process.pid\t[0\-]\t.*\.Atom
      launchRegex = ///^#{process.pid}\t[0\-]\t.*\.Atom///
      matches = (line for line in lines when launchRegex.exec(line))

      if matches.length > 0
        # Now we think we can fix the path since launcctl was clearly running
        console.error("Atom was launched from Launchctl, attempting to shim PATH")
        childProcess.execFile(shell, ['-c', 'echo $PATH'], patchPath)
      else
        console.log("Your Atom process wasn't in the launchctl list. Using your $PATH as is.")

  patchPath = (error, stdout, stderr) ->
    if (error)
      console.error("Unable to get $PATH")
      console.error("stderr: #{stderr}")
      return

    console.log("Old $PATH: #{process.env.PATH}")
    process.env.PATH = stdout.trim()
    console.log("New $PATH: #{process.env.PATH}")
```

Should probably accept a callback on finish, emit an event, or use promises, but hey it's just some throwaway code.

rgbkrk:
More `launchctl` tomfoolery:

```
$ launchctl getenv PATH
/usr/bin:/bin:/usr/sbin:/sbin
$ launchctl setenv PATH $PATH
$ launchctl getenv PATH
./bin:/usr/local/bin:/usr/local/sbin:/Users/kyle6475/.dotfiles/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin:/opt/X11/bin:/usr/local/opt/go/libexec/bin:/Users/kyle6475/.rbenv/bin:/Users/kyle6475/.rbenv/shims:/Users/kyle6475/go/bin
```

rgbkrk:
Now for my new init script.

```
try
  childProcess = require('child_process')

  atomProcs = childProcess.execSync("launchctl list | grep #{process.pid} | grep Atom").toString().split()
  launchRegex = ///^#{process.pid}\t[0\-]\t.*\.Atom///
  matches = (line for line in atomProcs when launchRegex.exec(line))

  if matches.length > 0
    console.log("Fixing paths for Atom")
    process.env.PATH = childProcess.execFileSync('/bin/bash', ['-c', 'echo $PATH']).toString().trim()
    process.env.GOPATH = childProcess.execFileSync('/bin/bash', ['-c', 'echo $GOPATH']).toString().trim()

catch exception
  console.error("Exception during PATH fixes: #{exception}")
```

PR #86, opened 2015-07-20
Fix the type of session to be a string containing a UUID

### LINKED ISSUE #85: `session` is supposed to be a string (UUID)
It appears Hydrogen is sending `session` as a number (0) when the Jupyter messaging protocol expects it to be a string containing a UUID.

https://ipython.org/ipython-doc/dev/development/messaging.html#general-message-format

This prevented Hydrogen from working with the IHaskell kernel, which blew up because it was expecting a string instead of a number as the value of `session`. I originally filed a bug against IHaskell (gibiansky/IHaskell#532), if you're interested in following the discussion that led to this conclusion.

It also seems that `msg_id` is being sent as a string but not a UUID. This probably doesn't matter as much but it's another potential bug for kernels that might try to parse the string as a UUID.

radix:
(maybe) fixes #85.

I don't know if this is actually gonna work; I just guessed by searching for `session` in the codebase and put in a null UUID instead of a 0. I don't know how to build or install atom packages so I haven't tested this; hopefully it's just a good basis for someone to come along and test/merge.

### DIFF
--- a/lib/kernel.coffee
+++ b/lib/kernel.coffee
@@ -92,7 +92,7 @@ class Kernel
         header = JSON.stringify({
                 msg_id: requestId,
                 username: "",
-                session: 0,
+                session: "00000000-0000-0000-0000-000000000000",
                 msg_type: "execute_request",
                 version: "5.0"
             })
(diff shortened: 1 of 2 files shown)


willwhitney:
People who file PRs with their bugs are my favorite people. I'll get this tested and merged ASAP.

radix:
@willwhitney I actually just tested this change by duplicating it in my ~/.atom/packages/Hydrogen directory directly, and it works! :D

willwhitney:
Sold! Certainly looks like it would to me.

willwhitney:
Shipped as 0.4.3.

rgbkrk:
legit

PR #95, opened 2015-08-22
Update Windows build requirements

TKCen:
- Changed the required compiler to VS 2013 as node zmq requires the v120
build tools.
- Removed required zmq install.

Visual Studio 2015 Community doesn't seem to provide the required build tools for node-gyp required by zmq.node. 
Tested on windows 10 with fresh install of atom.


willwhitney:
Thanks for this!

PR #97, opened 2015-08-31
Sign Jupyter messages

n-riesco:
I've adapted some of the code in JMP so that Hydrogen can sign Jupyter messages.

I've tested the changes against the python, IJulia and IJavascript kernels.

### DIFF
--- a/lib/config-manager.coffee
+++ b/lib/config-manager.coffee
@@ -25,7 +25,7 @@ module.exports = ConfigManager =
     buildConfiguration: (ports) ->
         config =
             version: 5
-            key: ""
+            key: uuid.v4()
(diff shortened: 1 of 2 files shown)


lib/config-manager.coffee
```diff
@@ -25,7 +25,7 @@ module.exports = ConfigManager =
     buildConfiguration: (ports) ->
         config =
             version: 5
-            key: ""
+            key: uuid.v4()
```
rgbkrk: A+

rgbkrk:
This looks great. We could change `_send` to `signedSend`.

willwhitney:
This looks awesome! Thanks for the work!

I'll merge and ship this.

PR #98, opened 2015-08-31
kernel-manager: Fix issue #90

### LINKED ISSUE #90: "SyntaxError: Unexpected token u"
This happens when I run my first line of Julia code which triggers starting the Julia kernel. It's getting thrown in getLanguageMappings.

n-riesco:
* Do not parse languageMappings if null.

Fix #90


willwhitney:
:+1:

PR #100, opened 2015-09-03
Fix issue #76 (bubble removed after enter)

### LINKED ISSUE #76: Output disappears when hitting enter/return/newline
It seems that in general Hydrogen will keep output in place when you modify the line that you've just evaluated. For example,

```python
foo(3)  #output here
```
if you edit this to say `foo(2)` after evaluating it, the "#output here" will stay in place. You can even insert many more characters and the output will flow along to the right. But if you hit enter to go to the next line, it disappears. I try to remember to hit enter *before* evaluating, but it's difficult.

This would help screencasting go better :)

n-riesco:
Fixes #76

### DIFF
--- a/keymaps/hydrogen.cson
+++ b/keymaps/hydrogen.cson
@@ -8,13 +8,15 @@
 # For more detailed documentation see
 # https://atom.io/docs/latest/behind-atom-keymaps-in-depth
 '.platform-darwin atom-text-editor':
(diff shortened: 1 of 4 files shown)


n-riesco:
Hi Will, 

please, don't merge this PR yet. I've just found a problem with the commit fixing findCodeBlock.

Nico

PR #101, opened 2015-09-04
Fix issue #76 (bubble removed after enter)

### LINKED ISSUE #76: Output disappears when hitting enter/return/newline

n-riesco:
Fixes #76

### DIFF
--- a/keymaps/hydrogen.cson
+++ b/keymaps/hydrogen.cson
@@ -8,13 +8,15 @@
 # For more detailed documentation see
 # https://atom.io/docs/latest/behind-atom-keymaps-in-depth
 '.platform-darwin atom-text-editor':
-  'cmd-alt-enter': 'hydrogen:run'
+  'cmd-alt-enter': 'hydrogen:run-and-move-down'
+  'shift-alt-enter': 'hydrogen:run'
 
 '.platform-darwin atom-workspace':
   'cmd-alt-backspace': 'hydrogen:clear-results'
 
 '.platform-win32 atom-text-editor, .platform-linux atom-text-editor':
-  'ctrl-alt-enter': 'hydrogen:run'
+  'ctrl-alt-enter': 'hydrogen:run-and-move-down'
+  'shift-alt-enter': 'hydrogen:run'
 
 '.platform-win32 atom-workspace, .platform-linux atom-workspace':
   'ctrl-backspace': 'hydrogen:clear-results'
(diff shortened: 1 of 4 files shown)


willwhitney:
This looks pretty good to me. Since these are flow-breaking changes for the user, I'll wait to do a more thorough review and merge this when I release other major changes.

n-riesco:
Would you want me to split the PR?
4cef6d6 is actually a bug fix.

willwhitney:
Ah, that’d be great!

n-riesco:
#102

willwhitney:
@n-riesco is this ready to go? We split out the bug fix and I think that got merged, so is this mergeable?

(Sorry for the embarrassing delay — lost track of this PR)

n-riesco:
Is it alright if I look at this in January? I'm currently travelling.

willwhitney:
Yup, no worries. I'm obviously not one to talk :)

rgbkrk:
Since you did #102 @n-riesco, should we close this down or would you want to make more progress?

rgbkrk:
I'm going to post an issue about where I'd like to take a refactor here as well as start thinking about/experimenting with the new inline support in Atom.

n-riesco:
I'm really sorry. I completely forgot about this PR.

This PR is really about whether we want two bindings to execute code:
```
'cmd-alt-enter': 'hydrogen:run-and-move-down'
'shift-alt-enter': 'hydrogen:run'
```

I'd like to hear what others think? Is this worth having? Are choice of bindings OK?

rgbkrk:
I agree with us having two bindings for executing code. The notebook has this notion too. 
`ctrl-enter` tends to mean "just run" while `shift-enter` means "run and move down". I'm guessing these conflicted with other Atom bindings at the time of creation (or an unwillingness to cover a binding used elsewhere), so we may not match amongst bindings with the notebook environments.

willwhitney:
Yep, when I first wrote this `shift-enter` was bound to something else by default in Atom. Not sure if it still is.

n-riesco:
Shall we merge this PR as it is?

rgbkrk:
Yeah, let's go for it and iterate.

PR #102, opened 2015-09-04
findCodeBlock: Selection text ending in column 0

n-riesco:
* Accounted for the case when the selection text ends in column 0.


willwhitney:
Thanks!

PR #113, opened 2015-09-20
Coffeescript integration

terraqout:
works with https://github.com/terraqout/jupyter-nodejs until https://github.com/notablemind/jupyter-nodejs/pull/16 isn't merged

Code review and any help appreciated

### DIFF
--- a/lib/kernel.coffee
+++ b/lib/kernel.coffee
@@ -18,7 +18,7 @@ class Kernel
         @language = @kernelInfo.language.toLowerCase()
         @executionCallbacks = {}
         @watchCallbacks = []
-
+        @loadedExtensions = []
         grammar = @getGrammarForLanguage(@language)
         @watchSidebar = new WatchSidebar(this, grammar)
         @statusView = new StatusView(@language)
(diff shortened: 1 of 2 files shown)


lib/kernel.coffee
```diff
@@ -145,15 +145,22 @@ class Kernel
 
         @signedSend message, @shellSocket
 
-    execute: (code, onResults) ->
+    handleExtensions: (code, language) ->
+        if language == 'coffeescript'
+            if language not in @loadedExtensions
+                @_execute '%load_ext coffee', => @loadedeExtensions.push language
+            "%%coffee bare no-prompt\n"+code
```
rgbkrk: My main issue with this is that it flies in the face of the JavaScript kernel I'm currently using, [ijavascript](https://github.com/n-riesco/ijavascript). It would be nice to [unify the JS/Node kernels](https://github.com/notablemind/jupyter-nodejs/issues/4), otherwise we're going to have this special matrix of handlers for Hydrogen.

Then again, if there's a way we can tell which kernel is setup, we can defer to the right one.

terraqout:
@rgbkrk I just use nodejs kernel because there coffeescript is ready to integrate. 
Is iJavascript better or supported well? I use ipython for three days, so all kernels look alike.

This non-generic approach isn't good at all, there are many X->JS and I'll write something generic when have a time for that.

rgbkrk:
Thinking on this, I might make a PR to the notable mind js kernel to have coffeescript, babel, etc. be kernelspecs you can install instead of using magics.

terraqout:
would be great

willwhitney:
My guiding principle is to not include any language-specific code in Hydrogen. Kyle's suggestion sounds like the best way to deal with this — that's what kernelspecs are for, after all!

I also want to have a way to switch which kernel you're using with a file, but I'm swamped right now, so I'm not sure when that's going to happen.

@rgbkrk having a single Babel kernel would be awesome.

terraqout:
@rgbkrk Notify me, please, when you finish with kernels. I'll implement hydrogen part

terraqout:
Finally we can use this: https://github.com/n-riesco/jp-coffeescript

PR #115, opened 2015-09-27
Fixed autocompletion for javascript and coffeescript

terraqout:
According to this https://github.com/atom/autocomplete-plus/wiki/Provider-API jupyter kernel names aren't equal to atom selectors now.
If you think this is brute and we need to get this selectors from atom itself, maybe you know how to do this?

Besides I remove js regexp, default one works pretty well for both languages

### DIFF
--- a/lib/autocomplete-provider.coffee
+++ b/lib/autocomplete-provider.coffee
@@ -2,8 +2,11 @@ _ = require 'lodash'
 
 KernelManager = require './kernel-manager'
 
-module.exports = AutocompleteProvider = ( ->
-    selectors = _.map KernelManager.getAvailableKernels(), ({language}) -> '.source.' + language
+module.exports = AutocompleteProvider = do ->
+    selectors = _.map KernelManager.getAvailableKernels(), ({language}) ->
+        '.source.' + language
+            .replace 'coffeescript', 'coffee'
+            .replace 'javascript', 'js'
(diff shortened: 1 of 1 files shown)


n-riesco:
Just a bit more background: currently Hydrogen doesn't send auto-completion requests to IJavascript. I've tested this PR and now IJavascript receives the requests.

rgbkrk:
Thanks @n-riesco and @terraquot!

rgbkrk:
What happens when this is run with e.g. Python or Julia?

n-riesco:
@rgbkrk It shouldn't affect any other language other than js and coffee. Anyway, I've checked and I can confirm that auto-completion continues to work for Python and Julia.

n-riesco:
Just in case it's relevant. I had to install the package `language-julia` for auto-completion to work with Julia files.

terraqout:
@n-riesco This is by design. If atom can't classify file as some specific language it can't forward this to hydrogen. But maybe this should be covered in docs.

terraqout:
I'll fix regexp for julia in other PR. It's frustrating to have no autocompletion with 'using', etc.

terraqout:
besides somebody uses litcoffee? I think nobody will create separate kernel for that. But it can be interpreted as .source.coffee

rgbkrk:
Thanks for checking it out @n-riesco.

PR #116, opened 2015-09-29
Kernel errors experimental mode

terraqout:
Works well for me with iJulia, don't test well with other languages.
But maybe too verbose like with Julia0.5.

Give a try, please

### DIFF
--- a/lib/kernel.coffee
+++ b/lib/kernel.coffee
@@ -55,15 +55,24 @@ class Kernel
 
         @kernelProcess.stdout.on 'data', (data) ->
             console.log "kernel process received on stdout:", data.toString()
-        @kernelProcess.stderr.on 'data', (data) ->
-            console.error "kernel process received on stderr:", data.toString()
-
+        @kernelProcess.stderr.on 'data', @stderr
             # console.log "launching:", commandString
             # @kernelProcess = child_process.exec commandString, (error, stdout, stderr) ->
             #     console.log 'stdout: ', stdout
             #     console.log 'stderr: ', stderr
             #     if error != null
             #         console.log 'exec error: ', error
+     stderr: (data, caption = "kernel answer") ->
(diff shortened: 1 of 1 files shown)


lib/kernel.coffee
```diff
@@ -55,15 +55,24 @@ class Kernel
...
+     stderr: (data, caption = "kernel answer") ->
+        detail = data.toString()
+        method = 'addInfo'
+        if /warning/gi.test detail
+            method = 'addWarning'
+        if /error/gi.test detail
+            method = 'addError'
+
+        console.error "kernel process received on stderr:", detail
+        atom.notifications[method] "Hydrogen: #{caption}",
```
willwhitney: We should probably change this to `"<language> Kernel: #{caption}"`

willwhitney:
If this is too verbose, we could throttle it and make any errors beyond the first in a window of time just say "More error messages received. Check the Developer Console for details." At least for now; at some point it would be cool to have a little panel where you can view all the messages from the kernel.

terraqout:
@willwhitney fixed.

It isn't verbose if kernel is not. Well Julia is a chatterbox. In python or coffee I get only kernel errors.

willwhitney:
We should deal gracefully with noisy kernels like Julia as well; I think the Notebook does something pretty concise in that case, and it'll be really annoying if lots of messages pop down in front of our code.

terraqout:
Let's try. I as Julia user didn't see trouble in all this warning bloating. If someone find this annoying he can create a ticket and we'll discuss it again.

I thought error handling is kernel internal, so it should be managed there.

willwhitney:
OK, sounds good. Merge away!

terraqout:
Just leave this here: https://github.com/JuliaLang/julia/issues/4895

Issue #130, opened 2015-11-05
Clear visual indication which code has been executed


onyxfish:
Okay, first of all, I flipping love this software. It's made me tremendously more productive. Thank you for creating it.

I'm wondering if it would be possible to add a clear visual indication which lines of code have been executed--either a background highlight or marker in the margin. Ideally this visual indication would disappear when that line of code has subsequently been modified.

When I'm doing analysis I find myself jumping around executing lines of code and it's sometimes difficult to tell for certain if I've run all the code that the current block depends on.

willwhitney:
Glad you like it!

This sounds like a great feature. I'm super busy right now (applying to PhD programs!) but if you're interested in writing a PR for this that would be awesome.

onyxfish:
I may do that! I'm not at all familiar with the architecture of Atom, but I'll poke at it and see what I can figure out. Thanks!

willwhitney:
I was thinking about this again today while writing some Julia in Hydrogen. The highlighting version of this would be _awesome_.

onyxfish:
Sat down and took a first whack at this. The good news: it's definitely possible. I've got a proof of concept that adds a class to lines in the "selection" case only. Kudos to you for having readable code. The bad news: I can't get the damn class style to show up, nor is the marker invalidation working in the way the documentation describes. I'm certain both of these things are my error. I'm going to take a break and I might take another crack at it later this weekend.

onyxfish:
Couldn't let it go. I've got it about 75% working on my master: https://github.com/onyxfish/hydrogen

willwhitney:
What a boss! This is looking really nice.

A couple of things to think about before shooting me a PR:
- You may have to do some funky conditional Less logic to make it look right on light backgrounds too. That's what `contrast` is doing here: https://github.com/willwhitney/hydrogen/blob/master/styles/hydrogen.less#L17
- These should get cleared when the kernel is reset, just like the bubbles. This is probably the right place for that: https://github.com/willwhitney/hydrogen/blob/master/lib/main.coffee#L107
- Let's make this opt-in with a settings switch for now. You can add a field to the config object: https://github.com/willwhitney/hydrogen/blob/master/lib/main.coffee#L16, and the docs for that are here: https://atom.io/docs/api/v1.1.0/Config

Great work!

onyxfish:
Sounds great, thanks for the tips. I'll send you a PR as soon as I can get it finished up. (I also need to re-implement your special casing for Ruby.)

[...]

PR #131, opened 2015-11-08
Add visual indication of modified code (See issue #130)

onyxfish:
Here is a basic implementation of the visual highlights discussed in #130. A few notes:

* I added an opt-in config switch.
* The highlights are now cleared when the kernel is restarted.
* There is a block of Ruby-specific code [here](https://github.com/onyxfish/hydrogen/blob/master/lib/main.coffee#L298) that I couldn't suss out the purpose of. All my test cases seem to work fine with it commented out.
* I haven't been able to get the `contrast` selection for the highlight color to work to my satisfaction. Do you have a better idea how to select a good color?
* There may be a weird case where the selected code errors and you don't display a persistent result bubble, but the highlight persists. I had a hard time tracing the logic of showing/not showing the result "flag" so I wasn't sure where to check if the highlight should persist.

I substantially reimplemented the code to find the correct row range. I think it's quite bit simpler now. Hopefully you agree.

Let me know if you see other issues! I tested against Python and Ruby.

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -27,6 +27,14 @@ module.exports = Hydrogen =
                           {"Python Django": "python", "Ruby (Rails)": "ruby"}'
             type: 'string'
             default: '{}'
+          highlightExecuted:
+            title: 'Highlight Executed Code Until Modified'
+            description: 'Highlight lines of code that have been executed. The
+                          highlight will be removed automatically when the code
+                          is modified so you will know your Jupyter context is
+                          out of date.'
+            type: 'boolean'
+            default: false
 
     subscriptions: null
     statusBarElement: null
@@ -183,7 +191,9 @@ module.exports = Hydrogen =
                         console.log(range)
 
                     if code?
-                        @highlightBlock(range)
+                        if atom.config.get('Hydrogen.highlightExecuted')?
+                          @highlightBlock(range)
+
                         @clearBubblesOnRow(range.end.row)
                         view = @insertResultBubble(range.end.row)
 
@@ -351,7 +361,7 @@ module.exports = Hydrogen =
     getFoldContents: (row) ->
         buffer = @editor.getBuffer()
         range = @getFoldRange(@editor, row)
-        
+
         return [
                 buffer.getTextInRange(range),
                 range
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -1,4 +1,4 @@
-{CompositeDisposable} = require 'atom'
+{CompositeDisposable, Range} = require 'atom'
 
 fs = require 'fs'
 zmq = require 'zmq'
@@ -177,12 +177,15 @@ module.exports = Hydrogen =
                     # @showWatchSidebar()
 
                     codeBlock = @findCodeBlock()
+
                     if codeBlock?
-                        [code, row, range] = codeBlock
+                        [code, range] = codeBlock
+                        console.log(range)
+
                     if code?
                         @highlightBlock(range)
-                        @clearBubblesOnRow(row)
-                        view = @insertResultBubble(row)
+                        @clearBubblesOnRow(range.end.row)
+                        view = @insertResultBubble(range.end.row)
 
                         KernelManager.execute language, code, (result) =>
                             view.spin(false)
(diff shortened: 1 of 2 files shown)


willwhitney:
Oh! That isn't actually Ruby-specific code. It ensures that if you've selected a bunch of blank lines at the end of your selection, Hydrogen doesn't send the empty lines to the kernel, and puts the results in the appropriate place. This is the behavior we want:



If I do the same thing with your PR, I get this:

screenshot 2015-11-08 12 51 18

> I substantially reimplemented the code to find the correct row range.

Good man! That was some of the first code I wrote on this project, and as a result the messiest. Thanks for this!

> There may be a weird case where the selected code errors and you don't display a persistent result bubble, but the highlight persists. I had a hard time tracing the logic of showing/not showing the result "flag" so I wasn't sure where to check if the highlight should persist.

I'm guessing you mean whether to show the result inline or as a popped-out bubble? The logic for that is in [ResultView](https://github.com/willwhitney/hydrogen/blob/master/lib/result-view.coffee#L73) and the function used to change that is [`setMultiline`](https://github.com/willwhitney/hydrogen/blob/master/lib/result-view.coffee#L146). `setMultiline` just adds/removes a CSS class. ResultView is pretty due for a refactor.

I think the highlight should pretty much always persist, even if the result is an error. It's not really feasible for Hydrogen to know if some code ran to completion, or only the first few lines of it ran, or whatever else. Given that, keeping a consistent rule that "highlighted means I ran it" seems reasonable.

> I haven't been able to get the contrast selection for the highlight color to work to my satisfaction. Do you have a better idea how to select a good color?

I'll play with this a little to see if I can figure something out.

---------

I found a bug in the code that finds the block to run: https://gist.github.com/willwhitney/a537b9b8a12b01806b43

willwhitney:
The color switching seems to work great if we contrast with `@syntax-background-color` instead of `@syntax-text-color`:

screenshot 2015-11-08 13 24 42
screenshot 2015-11-08 13 24 30

Using `contrast` against the text color will pick the color that's further away from the text color; for light text on dark background, that will pick the darker background option. But `contrast` against the background color will lighten a dark background or darken a light one.

Here's the changed line:

```less
.line.hydrogen-run {
    background-color: contrast(@syntax-background-color, @background-color-dark, @background-color-light, 20%);
}
```

willwhitney:
I also just noticed that it's impossible to see whether a line is currently selected or not when this is used. Adding some transparency might work?

onyxfish:
Fixed the highlight/contrast issue and the blank line trimming. Trying to replicate the other traceback you linked to. Spinning up a lua dev environment now.

onyxfish:
So, I'm still setting up my build environment, but I'm noticing that lua folds differently than other languages in atom. I'm wondering if this could be an issue with the lua module:

screen shot 2015-11-09 at 8 27 22 pm

screen shot 2015-11-09 at 8 27 26 pm

onyxfish:
Okay, I see now. The difference in how lua folds means that even though this row triggers the "end" special case, that logic can't find the correct row range because "end" is not considered part of the fold. I believe this code from before the refactor was addressing this case:

```
if @getRow(range[1] + 1).trim() == 'end'
    range[1] = range[1] + 1
```

However, this code, for languages that fold "correctly" could cause one extra line to be executed, if end wasn't followed by whitespace.

Thoughts?

willwhitney:
Ah, yeah. If there are two consecutive `end`s in a language that folds properly, this will run one `end` too many. Initially piggybacking on Atom's folding to find blocks saved a bunch of time, but it definitely has its problems. 

The Right solution probably involves language-specific parsing code, and I'm planning on implementing plugins for Hydrogen that support this whenever I get a solid block of time to work on it.

Until then, do you think indentation level would be a more consistent indicator? Or is there something else we could use as a proxy for this?

onyxfish:
Hey, I can try using indentation--that would be my best guess for something that will work. Although that's only truly reliable for languages with significant whitespace. I didn't realize until I started digging into this that the language folding is implemented in such an approximate manner. The Atom grammars themselves don't define the fold points. (Which is surprising, because the Textmate spec the grammars are based on does.)

I've been wrapped up in starting a new job this week, but I'm starting two weeks of vacation next week so I can take a deeper dive on this then.

willwhitney:
Awesome! Congrats on the new job :)

terraqout:
@onyxfish What is status of this PR?

onyxfish:
It works except for the whitespace issue. Unfortunately I've been totally absorbed in my new job so I haven't had any time to figure out the solution to that problem. If somebody wants to jump in here they should feel free.

rgbkrk:
I'll go ahead and close this. If you pick it up or another person does @onyxfish, let's have at it.

PR #136, opened 2015-11-18
Result scrolling

terraqout:
1. No hardwrap, only scrolling if content view is bigger than 500 px at any dimension
2. 500 was hardcoded because i don't know how to get maxHeight from .bubble-output-container
3. Replacing of escaped \n was added because of:
*If I execute language print variable command (print in python, console.log,...) I get multiline answer with any kernel. But when I execute declaration of variable with multiline text returned answer contains escaped newlines which by design is ok, but looks ugly in hydrogen*
4. ```overflow: scroll``` adds additional padding in all result views with/without scrollbars, maybe change it dynamically? But I didn't see advantages

All of statements can be discussed

### DIFF
--- a/styles/hydrogen.less
+++ b/styles/hydrogen.less
@@ -79,6 +79,7 @@
 
         .bubble-output-container {
             padding: 3px;
+            overflow: scroll;
         }
 
         .rich-close-button::before {
--- a/lib/result-view.coffee
+++ b/lib/result-view.coffee
@@ -130,8 +130,9 @@ class ResultView
                 @resultType = 'text'
                 container.innerText = container.innerText + result.data
 
-                if /\r|\n/.exec(container.innerText.trim())
-                    @setMultiline(true)
+                if /(\\n|\r|\n)/.exec container.innerText.trim()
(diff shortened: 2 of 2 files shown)


styles/hydrogen.less
```diff
@@ -79,6 +79,7 @@
 
         .bubble-output-container {
             padding: 3px;
+            overflow: scroll;
```
willwhitney: If this is only going to have an effect on watches, we could put this in the `.watch-sidebar` section: https://github.com/willwhitney/hydrogen/blob/result-watch/styles/hydrogen.less#L185

willwhitney:
Otherwise this looks good to go.

terraqout:
I occasionally added "watch" to ticket name. It should work on all result views. Would this be merged on next release?

willwhitney:
You've tested that this does work in practice? Previously marker
decorations were simply not scrollable in Atom.

terraqout:
If answer can't be presented in a marker smaller than 500px (default bubble lenght) presenter view changed to multiline answer box:
labels

willwhitney:
Cool. If it works, merge it!

PR #143, opened 2015-12-05
History of watch results

slavaGanzin:
out

I don't like code, but it works

Besides what tool do you use to record gifs? My is ugly

### DIFF
--- a/lib/watch-view.coffee
+++ b/lib/watch-view.coffee
@@ -25,6 +25,33 @@ class WatchView
         @element.appendChild(@inputElement.element)
         @element.appendChild(@resultView.element)
 
+        @addHistoryScrollbar().clearHistory()
+
+    clearHistory: (@currentHistory=[]) -> this
+    addToHistory: (result) ->
+      return if result.data is 'ok'
+      @currentHistory.push(result)
+      @historyScrollbar.querySelector('.hidden').style.width =
+        (total = @currentHistory.length * @historyScrollbar.offsetWidth) + 'px'
+      @historyScrollbar.scrollLeft = total
+      this
+
+    addHistoryScrollbar: ->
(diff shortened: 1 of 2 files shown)


slavaGanzin:
@willwhitney Hello, Will. Can you spend some time to review my PRs?

willwhitney:
This is an interesting tool! However, it's not showing up for me. I'm on the history branch and up to date, but no scrollbar (using Python plaintext or matplotlib results).

A few notes based on the gif:

1. Is there a policy for only holding a certain amount of history? We shouldn't keep all of it around forever; that'll result in high memory usage and may not help much.
1. A label on the control that says "History" or something would be good.
1. We should use a UI element with discrete steps for paging through these. The scrollbar is a little confusing, and may make it hard to take just one step when there are many in the history. Maybe something like Fantastical's `< previous | current | next >` element (screenie below). Alternatively, if we only have a few, we could show dots for each of them and let you click directly to one (sort of like in an image gallery).

screenshot 2015-12-30 11 34 32

For gifs sometimes I use CloudApp or Recordit, but they charge money for high quality gifs. I actually get the best results from recording a video with Quicktime and then converting it into gifs. That's a pain in the ass, though.

slavaGanzin:
1. For now you can remove watch and add it again. I think about adding a limit near 10000 values or better some bytesize limit, which should be an error by design. I think one button to drop all history would be enough

2. I'll add some
3. I've tried to make something like that: http://debug.elm-lang.org/edit/Stamps.elm
In this example discreet switcher would be a pain. So maybe just add step indicator?

Besides, one history for all watches maybe great idea.

rgbkrk:
How are you feeling about this if we go ahead and merge it and iterate @slavaGanzin?

slavaGanzin:
Let's try

PR #144, opened 2015-12-08
Kernel manager impovement

slavaGanzin:
I was unable to test https://github.com/willwhitney/hydrogen/pull/138 because my python kernels was installed into /usr/local/share/kernels and I have only 'default' hardcoded python kernel.


slavaGanzin:
Reimplemented here: https://github.com/willwhitney/hydrogen/commit/530b02795864d4b6f19550a14fb4acdfee410d1b

PR #145, opened 2015-12-09
Reimplement paths and cut uneed tests

slavaGanzin:
https://github.com/willwhitney/hydrogen/issues/21#issuecomment-163370467

>The jupyter --paths are:
>config:
>`/Users/me/.jupyter
>/usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/etc/jupyter
>/usr/local/etc/jupyter
>/etc/jupyter`

Previous paths.coffee was a port of jupyter code which is a bit outdated and messy and can't find something like this: /usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/etc/jupyter

and besides didn't check if it was a real directory or just "should be"

```ipython kernelspec list```
Can be used for available kernels, so we can cut paths.coffee

### CODE REVIEW

File: lib/paths.coffee [hunk 1]
```diff
@@ -1,139 +1,31 @@
...
-#
-# > jupyterPath()
-# ['/Users/rgbkrk/.local/jupyter', '/usr/local/share/jupyter']
-#
-# >jupyterPath
-# ['/Users/rgbkrk/.local/jupyter/kernels', '/usr/local/share/jupyter/kernels']
-jupyterPath = (subdirs...) ->
-  paths = []
+jupyterResolve = _.memoize (category) ->
+    match = child_process.spawnSync('jupyter', ['--paths']).stdout.toString()
```
rgbkrk: There's now a `--json` flag which was created because of this use case in Atom, node, and Electron. I'd like to see this get addressed in https://github.com/nteract/jupyter-paths and just rely on it there (more than one package needs this functionality).

slavaGanzin:
@rgbkrk Hydrogen can handle what we need with just parsing ```ipython kernelspec list --json```. So paths.coffee can be removed. And code changed to this: https://github.com/nteract/jupyter-paths/issues/4 .  I'll cut this regexp madness and change it to JSON.parse as you do. (thanks for a flag).

Or you prefer to move this code to jupyter-paths too? And use it as dependency?

rgbkrk:
I prefer more of the jupyter centric code to move to separate libraries and used as dependencies.

slavaGanzin:
@rgbkrk Kyle, so will you add ipython kernelspec to jupyter-paths or should I create my own lib?

rgbkrk:
Done. Just made a new library called ipython-paths.

https://github.com/nteract/ipython-paths

Two functions exported: one for getting the expected set of kernel directories (existence not checked) and another which returns a `Promise` that resolves to an array of directories that do exist.

slavaGanzin:
great, I'll rewrite and merge this after

slavaGanzin:
@rgbkrk As I understand your ipython-paths wouldn't find this:
/usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/etc/jupyter

so what's the usefulness of that code?

rgbkrk:
The usefulness of the original or the current code in ipython-paths? `ipython-paths` implements exactly the paths listed in [kernel-specs](https://ipython.org/ipython-doc/3/development/kernels.html#kernel-specs).

    | Unix	| Windows
---|---------|--------------
System | /usr/share/jupyter/kernels /usr/local/share/jupyter/kernels | %PROGRAMDATA%\jupyter\kernels
User | ~/.ipython/kernels

As for that path you're asking about, that comes from the very Python specific `sys.prefix`. It was certainly in the original implementation that was here before. Since you questioned that old IPython code (which covers lots of cases not shown here), I didn't include it - just the bare ones that could be done without having to rely on a Python installation.

It's not hard to add back in, and I'll push back on wanting to introduce that. What would make me say so? `sys.prefix` comes from python, virtualenv and conda installations included. This also means that in order for Hydrogen/Atom to pick up this value properly, it needs to be run from the command line directly `atom .`. If we stop doing this, we have one less irritation paths as set after an Atom restart.

Here's `sys.prefix` on my machine followed by what it looks like if in a virtualenv:

```
$ python -c "import sys; print(sys.prefix)"
/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5
$ mkvirtualenv test
Using base prefix '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5'
New python executable in test/bin/python3.5
Not overwriting existing python script test/bin/python (you must use test/bin/python3.5)
Installing setuptools, pip, wheel...done.
Overwriting test/bin/activate with new content
$ python -c "import sys; print(sys.prefix)"
/Users/rgbkrk/.virtualenvs/test/bin/..
```

Ask yourself if Jupyter is truly language independent when relying on this.

If a user wants one of these, they should be setting up a kernel for their virtualenv instead of coding Hydrogen according to any given language (see the last bullet in #149).

slavaGanzin:
I had troubles install path of different kernels, https://github.com/alexarchambault/jupyter-scala for example. They tend to settle in unexpected folders and I don't think it's good idea say to user: "it's your trouble that kernel you use don't give a care to install folder we like".

I would prefer parsing ```ipython kernelspec list --json``` approach, it simplifies code. But it should be tested.

rgbkrk:
> I would prefer parsing ipython kernelspec list --json approach, it simplifies code. But it should be tested.

I certainly agree with that. Perhaps we had a miscommunication here. I thought you wanted the old ipython paths, which is why I put together `ipython-paths`. `ipython kernelspec list --json` is the same as `jupyter kernelspec list --json`, as that command doesn't exist in old ipython. I'd make it `jupyter` in your code there.

While you're at it if you're looking for good user experience, you should use the async spawn instead of the synchronous one.

> I had troubles install path of different kernels, https://github.com/alexarchambault/jupyter-scala for example. They tend to settle in unexpected folders and I don't think it's good idea say to user: "it's your trouble that kernel you use don't give a care to install folder we like".

We would want to open an issue or patch it upstream if possible, instead of maintaining a large collection of hacks to make things work for every language.

slavaGanzin:
> While you're at it if you're looking for good user experience, you should use the async spawn instead of the synchronous one.
Hydrogen written in sync manner, so it would be unexpected interface. Added memoization to speed up.

Besides do this spawnSync freezes all atom or hydrogen is working in separate process?

willwhitney:
spawnSync will freeze all of Atom. Unfortunately plugins run in the same process.

Everything so far has been sync, but only because all of the calls are basically instantaneous. Definitely use async calls for anything that will take time.

slavaGanzin:
I benchmark this call and it's "almost instant" and faster than bunch of code before.
Rewriting everything in async mode wouldn't speed up hydrogen perceptually, because zmq/kernel start delay is by an order more.

rgbkrk:
Cool, have at it.

willwhitney:
:clap:

PR #147, opened 2015-12-15
Testing dependencies hook

slavaGanzin:
Must fix (added to review):
https://github.com/willwhitney/hydrogen/issues/126
https://github.com/willwhitney/hydrogen/issues/106
https://github.com/willwhitney/hydrogen/issues/59
https://github.com/willwhitney/hydrogen/issues/142
https://github.com/willwhitney/hydrogen/issues/141
https://github.com/willwhitney/hydrogen/issues/157

Not finished. Any help approved


terraqout:
Will refactor to use:
```  installPackageDependencies: -> atom.notifications ```

**Comment by slavaGanzin** [comment]
Some work was done here: https://github.com/jovyan/jupyter-devinstall. Maybe reuse?

**Comment by rgbkrk** [comment]
This is definitely a really good idea for prompting users.

**Comment by slavaGanzin** [comment]
It would be great if you test it on Mac, because I write Mac-specific code copypasting it from stack overflow without even try. 

Besides it works on windows 7. But if someone can confirm that it would be great

**Comment by willwhitney** [comment]
This is a really good thing to do.

File: dependencies.coffee [hunk 1]
```diff
@@ -0,0 +1,37 @@
...
+  #TODO: test on windows
+  spawn("cl").stdout.replace /Version \d+/, (x) ->
+    if parseInt(x.split(' ')[1]) < MINIMAL_WIN_CL_VERSION
+      throw new Error 'You need Visual Studio 2013 Community Edition to compile ZMQ'
+
+
+# if os.platform() == 'linux'
+
+if os.platform() == 'darwin'
+  unless spawn("brew list | grep 'pkg-config'").stdout.length
```
willwhitney: We should allow for builds of `pkg-config` installed from other places as well. Replacing this with `which pkg-config` should work the same.

dependencies.coffee
```diff
@@ -0,0 +1,37 @@
...
+    throw new Error 'You need pkg-config to install zmq: brew install pkg-config'
+  unless spawn("brew list | grep 'zeromq'").stdout.length
```
willwhitney: We can rely on the Python `import zmq` test below to check whether zeromq is installed. As above, we don't want to just ask brew whether it has zeromq installed, because it may come from somewhere else too.

dependencies.coffee
```diff
@@ -0,0 +1,37 @@
...
+    throw new Error 'You need ZMQ to use Hydrogen: brew install zeromq'
+
+
+unless (/\ 2\./.test(spawn('python --version').stderr) ||
+  /\ 2\./.test(spawn('python2 --version').stderr))
+    throw new Error 'Python2 is required to build Hydrogen'
```
willwhitney: We should make it clear exactly how Python 2 must be available. Maybe add something like "Either `python` or `python2` must be in your path and refer to Python 2."

willwhitney:
Thanks for doing this — it'll really improve the install process for people (and cut down on the confused issues we get).

rgbkrk:
Dependencies.js should likely be gitignored (or so I assume). Other than that, how's about we roll this?

slavaGanzin:
@rgbkrk 
I use it here: https://github.com/nteract/hydrogen/blob/dependencies/package.json#L24
In preinstall I have no coffee script to launch it. Any ideas?

rgbkrk:
How do you feel about writing plain JS for that one?

We can also package it before publication like one does for transpiled scripts, using the `prepublish` key on scripts. Then we'd have to make coffeescript be a dev dependency.

slavaGanzin:
I'll drop dependencies.coffee

rgbkrk:
Works for me.

rgbkrk:
We can always make that idiomatic JS after.

PR #175, opened 2016-02-24
Add buttons on watches sidebar

### LINKED ISSUE #174: BUG: Check ticks above search panel
I'm running with Ubuntu 14.04 and atom 1.0.5 and see the tick marks above the search panel:

bug_hydrogen

lgeiger:
I added buttons to the watches sidebar to trigger the `toggle watches` and `remove watch` commands. What do you think about the styling?

old vs new:

old new

The possibility to copy one line outputs by clicking on them was added as well:

copy

It fixes #174 with this hack https://github.com/atom/atom/issues/6857

Apart from that this PR fixes some styling issues mostly related to scrollbars by changing `overflow` from `scroll` to `overlay`. Can you test if it works on Windows and Linux as well?

### DIFF
--- a/styles/hydrogen.less
+++ b/styles/hydrogen.less
@@ -192,26 +192,15 @@
     padding: 10px;
     border-left: 1px solid @base-border-color;
 
-    .watch-sidebar-title {
-        text-align: center;
-        margin-top: 5px;
-        margin-bottom: 5px;
-    }
-
-    .watch-sidebar-language {
-        margin-top: 5px;
-        margin-bottom: 5px;
-        text-align: center;
-        cursor: pointer;
-
-        &:hover {
-            text-decoration: underline;
+    .toolbar {
+        margin-bottom: 10px;
+        .btn-group{
+            float: right;
         }
     }
 
     .watch-view {
         width: 100%;
-        margin-top: 10px;
         .watch-input {
             padding: 3px;
             border-radius: 3px;
@@ -282,7 +271,7 @@
             display: inline-block;
             width: 100%;
             max-height: 400px;
-            overflow-y: scroll;
+            overflow-y: overlay;
             white-space: pre-wrap;
         }
 
--- a/styles/hydrogen.less
+++ b/styles/hydrogen.less
@@ -11,6 +11,10 @@
     z-index: 6;
 }
 
+.bottom{
+    z-index: 4;
+}
+
 .hydrogen.output-bubble {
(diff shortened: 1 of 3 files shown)


slavaGanzin:
Would be great if you add command for copying result too. 
Styles look great.

If your interested in watches functionality additionally I ask you too invent some interface for this https://github.com/nteract/hydrogen/pull/143

lgeiger:
> Would be great if you add command for copying result too.

I don't know what you mean?

> I ask you too invent some interface for this #143 

I'll take a look at this.
But I must say I quite like the scrollbar history ;)

slavaGanzin:
I'm talking about copying result with hydrogen command which can be mapped to some keybindings.
> But I must say I quite like the scrollbar history ;)

Does it looks intuitive for you? If so we shouldn't reinvent wheel

rgbkrk:
@slavaGanzin do you think we could merge this PR and you could address the review feedback in a follow on PR?

styles/hydrogen.less
```diff
@@ -79,7 +83,8 @@
 
         .bubble-output-container {
             padding: 3px;
-            overflow: scroll;
+            padding-right: 10px;
+            overflow: auto;
         }
```
slavaGanzin: Is scrolling available with overflow: auto?

slavaGanzin:
It's all about view. If this https://github.com/lgeiger/hydrogen/commit/e5b061b54551cac6c3758c50fb5a0baea49b110a were separated this PR could be merged even without review. Views don't broke logic.

I think we should merge this.

slavaGanzin:
@lgeiger Thanks for your work

rgbkrk:
Thank you @lgeiger!

styles/hydrogen.less
lgeiger: Yes scrolling works.
For me the only difference is that the space for the scrollbar isn't reserved with `auto` when no scrolling is needed. I'm talking about these dark bars:
bildschirmfoto 2016-02-26 um 11 58 28

lgeiger:
For me the scrollbar is fine. The only issue is that on Mac the default is that scrollbars don't show when you work with a trackpad. Do you know if there is an option in Atom to force the scrollbars?

In the future I'll split my PR in a more logical way and don't try to squash everything in one. Sorry :)

@slavaGanzin About C++: I work with [ROOT](https://root.cern.ch/) they have a [Jupyter Kernel](https://github.com/root-mirror/root/tree/master/bindings/pyroot/JupyROOT) and it works quite well.

Issue #173, opened 2016-02-22
Hydrogen v0.7.0 incompatible with Ubuntu 14.04 LTS (IPython v1.2.1)


n-riesco:
We shouln't use `spawnSync` if Node.js is version 0.10. See [this commit](https://github.com/nteract/hydrogen/commit/2d372b5b5b7bfc86bad8dd34de9371f661de7582), [this commit](https://github.com/nteract/hydrogen/commit/530b02795864d4b6f19550a14fb4acdfee410d1b) and [this issue](https://github.com/nteract/hydrogen/issues/172).

@slavaGanzin @rgbkrk what are your thoughts on this issue?

slavaGanzin:
I tested async methods and the only thing it fixes - small freeze of UI on hydrogen start.
 Jupyter/ipython works slow on building list of kernels (second on my not so fast ssd, two-tree on hdd), but cache them after first run.

What I can propose:
1. memoize kernels in hydrogen settings or in file
2. on getAvailableKernels call we return our cached version and trigger async update of kernels cache

this may lead to first-time undiscoverability of freshly installed kernel, but thats ok, we had this before this commit

n-riesco:
I'd like to find a solution that is kind to Ubuntu users.

I think having the option to store kernel specs in the Hydrogen settings would help a little. An Ubuntu user could edit the settings to workaround this issue. However, this won't do for Ubuntu users that lack the technical knowledge to create a kernel spec.

In the past, I thought of adding a flag to install the IJavascript kernel in Hydrogen. This would be possible, if the Hydrogen settings stored kernel specs.

slavaGanzin:
What the trouble with Ubuntu? They don't have ipython or jupyter in their paths?
@n-riesco Can you explain this in depth, please?


In iTourch docs i found this:
```
iTorch requires or works with

Mac OS X or Linux (tested in Ubuntu 14.04 and Arch Linux)
Torch-7
IPython version 2.2 or above (you can check your version of ipython using ipython --version)
ZeroMQ
```
So they should have iPython to work. And so it isn't in PATH as I think.

> We shouln't use spawnSync 
I don't wont user to edit anything, just memoize kernelspecs generated by ```jupyter kernelspec``` or ```ipython kernelspec``` in a file to speedup a little

n-riesco:
[...]

Issue #38, opened 2015-05-26
remote kernel support


jzthree:
This was in the  original to-do list ("support remote kernels"). I think this feature would be extremely helpful for people who need to run the code on remote servers.  Is this feature still part of the roadmap for the near future :)?

TKCen:
Maybe [remote kernel](https://github.com/korniichuk/rk) can be used/integrated to support this feature?

willwhitney:
I would still like to support this. 

@rgbkrk, I think you had some concerns about zmq on random ports making it through people's networks?

@TKCen it looks like `rk` is helping write kernelspecs for remote systems. If they work with Jupyter, I would expect them to work with Hydrogen. The only caveat is that if you have a local kernel for a particular language, Hydrogen may use that instead (it uses the first matching kernel it finds). 

Would someone be willing to give `rk` that a try and let us know if it works?

willwhitney:
Just opened #39 to address switching between multiple kernels for a language.

rgbkrk:
`rk` appears to be over SSH which works for a lot of folks, myself included. The "template" kernel would probably work just fine as is. Might as well try it out. I wouldn't want to manage dependencies this way, but it's at least simple for most users. It doesn't seem like it would require any special requirements on Hydrogen at all.

As for my concerns, these are my observations from teaching classes, running a few tmpnb deployments, JupyterHub deployments, and my own security perspectives. Feel free to take it all with a grain of salt.

## Remotes over ZeroMQ

It's all about the kernels communication layer. Not every ZeroMQ installation includes encryption by default (libsodium is an optional dependency for ØMQ). ØMQ4 does have curve, which is wire-compatible with ØMQ3, but won't necessarily have libsodium. One can build it that way for themselves, certainly. This tends to mean either using websockets (through a notebook server) or tunneling through SSH.

## Remotes over websockets

The most reliable way I've seen for getting websockets through is making sure they're served over HTTPS on the standard port (443). That boils down to corporate (MITM) proxies totally mucking up the sticky connection with websockets, interfering with the flow, and generally being nasty. As for port, that's about stringent controls (no high ports getting through, etc.)

That being said, I think the remote kernel issue should be addressed with upstream Jupyter to work for the notebook server, etc. As long as we're adhering to the same protocol, it should help multiple projects at once.

[...]

PR #177, opened 2016-02-24
caching available kernels into config

slavaGanzin:


lib/kernel-manager.coffee
```diff
@@ -4,15 +4,26 @@ child_process = require 'child_process'
...
-
+    
+    getAvailableKernels: ->
+        @availableKernels = @getConfigJson 'kernelspecs', @availableKernels
+        @updateKernels() unless @availableKernels.length
+        @availableKernels
+               
+    updateKernels: ->
+        save = (out) =>
+            @availableKernels = _.pluck JSON.parse(out).kernelspecs, 'spec'
```
n-riesco: This throws because `out` is not valid JSON when using IPython v1.2.1

n-riesco:
Also you need to update ` lib/main.coffee`:

```coffeescript
module.exports = Hydrogen =
    config:
        kernelspecs:
            title: "Kernel Specs"
            description: 'Kernel specs as reported by:

                            $ jupyter kernelspecs list --json

                          E.g.:

                            {
                              "ijavascript": {
                                "spec": {
                                  "env": {},
                                  "argv": [
                                    "node",
                                    "/home/user/node_modules/ijavascript/lib/kernel.js",
                                    "--protocol=5.0",
                                    "{connection_file}"
                                  ],
                                  "language": "javascript"
                                },
                                "resources_dir": "/home/user/node_modules/ijavascript/images"
                              }
                            }'
            type: 'string'
            default: '{}'
```

lib/kernel-manager.coffee
rgbkrk: Yeah we need tooling for upgrading from old specs to new specs, though I think it can be a follow on PR after.

n-riesco: How about guarding with a `try/catch` block?

```coffeescript
         save = (out) =>
            try
                @availableKernels = _.pluck JSON.parse(out).kernelspecs, 'spec'
                @setConfigJson 'kernelspecs', @availableKernels
                atom.notifications.addInfo 'Hydrogen Kernels updated:',
                  detail: (_.pluck @availableKernels, 'display_name').join('\n')
            catch
```

n-riesco:
As a user, I don't want to type my custom kernelspecs every time I launch Atom. I think if a user defines manually the kernel specs, Hydrogen shouldn't delete the user's settings.

slavaGanzin:
@n-riesco And it wouldn't. If you have non working ipython kernelspec and jupyter kernelspec, it would use your kernelspec option.
 What for you want to edit kernelspecs? If you want to optimize something or provide keys to kernel I though it should be done internally in jupyter config (if any)

n-riesco:
#38

n-riesco:
[This commit](https://github.com/nteract/hydrogen/commit/152e42e4c3b8ce9632db2166195090e9723ecf56) is not working for me.

I'm still debugging, but I've noticed typos [here](https://github.com/nteract/hydrogen/blob/152e42e4c3b8ce9632db2166195090e9723ecf56/lib/kernel-manager.coffee#L33) and [here](https://github.com/nteract/hydrogen/blob/152e42e4c3b8ce9632db2166195090e9723ecf56/lib/kernel-manager.coffee#L35)

n-riesco:
I don't understand the logic [here](https://github.com/nteract/hydrogen/blob/152e42e4c3b8ce9632db2166195090e9723ecf56/lib/kernel-manager.coffee#L21-L27). If we don't return, `kernelspec` is undefined.

n-riesco:
Why is the setting `kernelspec` and not `kernelspecs`?

slavaGanzin:
>I'm still debugging, but I've noticed typos here and here 

That typos were for my debug purposes. And I left them - my fault

> Why is the setting kernelspec and not kernelspecs?
Because it's an output of kernelspec command, which contain kernelspecs property.

slavaGanzin:
#38 
You definitely shouldn't use that param for remote kernels.
You can specify kernel.json with same functionality in your jupyter|ipython home directory and use it for this.

slavaGanzin:
PS: Does anyone know where atom stores the settings? /home/user/.atom/config.cson
Yes

slavaGanzin:
https://jupyter-client.readthedocs.org/en/latest/kernels.html#connection-files

n-riesco:
@slavaGanzin I disagree on #38 (we shouldn't force users to install IPython to use Hydrogen, specially users of languages other than Python).

> https://jupyter-client.readthedocs.org/en/latest/kernels.html#connection-files
Connection files represent actual connections (they don't define a kernel spec).

slavaGanzin:
https://jupyter-client.readthedocs.org/en/latest/kernels.html#kernel-specs
Sorry wrong link.
Nobody forces to use Ipython, you can use jupyter it works the same

n-riesco:
[This line](https://github.com/nteract/hydrogen/blob/152e42e4c3b8ce9632db2166195090e9723ecf56/lib/kernel-manager.coffee#L36) is causing the `kernelspec` spec settings to be reset each time Atom is launched. It should be:

```coffeescript
save stdout unless e
```

n-riesco:
I reckon [this line](https://github.com/nteract/hydrogen/blob/152e42e4c3b8ce9632db2166195090e9723ecf56/lib/kernel-manager.coffee#L11) should be:

```coffeescript
kernels = _.pluck @getConfigJson('kernelspec', {kernelspecs:{}}).kernelspecs, 'spec'
```

slavaGanzin:
>save stdout unless e
I rewrite a bit and now it's guarded with:
https://github.com/nteract/hydrogen/commit/a5791554740151ccc51b078e28d77af69e8cb236#diff-aa15d36609a9a564a24156e6778cd62fR28

it can't be ```unless e``` because we should show this message https://github.com/nteract/hydrogen/commit/a5791554740151ccc51b078e28d77af69e8cb236#diff-aa15d36609a9a564a24156e6778cd62fR23

n-riesco:
> Nobody forces to use Ipython, you can use jupyter it works the same

That's not my point. Ideally, there shouldn't be any need for Hydrogen users to install Jupyter or IPython.

The current state of Hydrogen is that either  a recent version of `IPython` or `Jupyter` need to be present to run Hydrogen.

Previous versions of Hydrogen used to work on a box with IPython 1.2.1 (as long as the kernel specs were installed in one of the locations known to Hydrogen).

When I reported #173 , I was hoping to find a friendly compromise. I'm not asking to stop using the newer versions. What I'm asking is that we write robust code (i.e. doesn't throw uncaught exceptions) and that allows Hydrogen to run in a standard Ubuntu 14.04 box.

n-riesco:
I've tested [the latest commits](f3a720cba25441a8fc3be8773f723871551a4c6f) and it's working in my machine.
Thanks!

slavaGanzin:
@n-riesco Thanks for review
@rgbkrk I think this is ready to merge. And release

rgbkrk:
What would you say our version bump is here?

n-riesco:
@rgbkrk I haven't followed some of the latest commits, so I can't answer your question. But I would suggest that:
- if new dependencies have been added, we would bump at least the minor number (I wouldn't bump to 1.0.0 just yet)
- if no new dependencies are required, then I would only bump the patch number.

slavaGanzin:
0.7.1

rgbkrk:
Great, done! :tada: 

```
$ apm publish patch
Preparing and tagging a new version ✓
Pushing v0.7.1 tag ✓
Publishing Hydrogen@v0.7.1 ✓
```

PR #184, opened 2016-03-05
Make result bubble appear consistent

lgeiger:
Put the result bubble in the right place with `run-all`  and `run-all-above` commands.

They appear at the last not empty line.

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -308,9 +308,7 @@ module.exports = Hydrogen =
 
     findCodeBlock: (runAll = false, runAllAbove = false) ->
         if runAll
-            endRow = @editor.getLastBufferRow()
-            while @blank(endRow) and endRow > 0
-                endRow -= 1
+            endRow = @editor.getBuffer().getText().replace(/\s+$/).split('\n').length - 1
             return [@editor.getText(), endRow]
 
         buffer = @editor.getBuffer()
@@ -331,9 +329,8 @@ module.exports = Hydrogen =
         console.log "row:", row
 
         if runAllAbove
-            while @blank(row) and row > 0
-                row -= 1
-            return [@getRows(0, row), row]
+            endRow = @getRows(0, row).replace(/\s+$/).split('\n').length - 1
+            return [@getRows(0, row), endRow]
 
         indentLevel = cursor.getIndentLevel()
         # indentLevel = @editor.suggestedIndentForBufferRow row
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -308,7 +308,10 @@ module.exports = Hydrogen =
 
     findCodeBlock: (runAll = false, runAllAbove = false) ->
         if runAll
-            return [@editor.getText(), @editor.getLastBufferRow()]
+            endRow = @editor.getLastBufferRow()
+            while @blank(endRow) and endRow > 0
+                endRow -= 1
+            return [@editor.getText(), endRow]
 
         buffer = @editor.getBuffer()
         selectedText = @editor.getSelectedText()
@@ -328,6 +331,8 @@ module.exports = Hydrogen =
         console.log "row:", row
 
         if runAllAbove
+            while @blank(row) and row > 0
+                row -= 1
             return [@getRows(0, row), row]
 
         indentLevel = cursor.getIndentLevel()


slavaGanzin:
Boo, don't use while - nobody is secured from endless loops. You can always omit it with

```for ... when``` in coffee style
or 
```_.filter``` if you prefer going functional

Another way to achieve last non empty line:
```
@editor.getBuffer().getText().replace(/\s+$/).split('\n').length
```

I know this code is full of "while" and you just mimic behavior. But I had fixed issue with while at this project recently, so I want to prevent them. This code is far from functional "pureness", but while is absolute evil.

**Comment by n-riesco** [comment]
Could we merge #185 before this PR?

**Comment by n-riesco** [comment]
@slavaGanzin I'm not sure I understand. What I would write in Coffeescript is:

```coffeescript
endRow = @editor.getLastBufferRow()
endRow-- while @blank(endRow) and endRow > 0
```

rgbkrk:
:+1: for readability

lgeiger:
I'll make an easy to merge PR.

slavaGanzin:
@n-riesco I think this https://github.com/lgeiger/hydrogen/commit/cb7db1206b4436a928e0e9941901271c01b549d8 is more readable and descriptive.

n-riesco:
@slavaGanzin The thing that I find most wrong about https://github.com/lgeiger/hydrogen/commit/cb7db1206b4436a928e0e9941901271c01b549d8 :

```js
endRow = @editor.getBuffer().getText().replace(/\s+$/).split('\n').length - 1
```

is that it creates unnecessary strings:
- `replace(/\s+$/)` creates a string that, in the worst-case scenario, is as big as the whole document;
- `split('\n')` again creates a number of strings that, again in the worst-case scenario, are as big as the whole document;
- all these strings are created and dumped straight way, because we are only interested in the `length`;

On the other hand, I find @lgeiger 's initial version more readable and descriptive:

```coffeescript
endRow = @editor.getLastBufferRow()
while @blank(endRow) and endRow > 0
    endRow -= 1
```

because it resembles more closely the algorithm that it's trying to implement; namely, that of skipping blank lines, i.e.:
- check whether a line is blank
- iterate until finding a non-blank line or reaching the the first line

Anyway, I don't feel strongly about while-loops. If Hydrogen, as a project, decided to ban the use of while-loops, I would comply. Personally, I believe that this is a bike-shedding argument, and that, at this time, Hydrogen is better off without this kind of constraints.

slavaGanzin:
If you analyze this code and look underneath you'll find pretty the same: https://github.com/lgeiger/hydrogen/blob/cb7db1206b4436a928e0e9941901271c01b549d8/lib/main.coffee#L367-L369 and even worse. How do you think you can take last line of buffer, except slicing it with '\n'? 

If you are so afraid of over consumption memory try to avoid calling functions (inline them!) in a cycle - they copy arguments. 
 As for me, we don't leave in 90s when "640kb is enough for everyone". Ram  is cheap and our time is not. While cycles often causes problems - they aren't pure.

 But generally I don't care. You can use any code you want. But, please, move it to separate function.

n-riesco:
On 07/03/16 13:34, Slava wrote:
> If you analyze this code and look underneath you'll find pretty the same: https://github.com/lgeiger/hydrogen/blob/cb7db1206b4436a928e0e9941901271c01b549d8/lib/main.coffee#L367-L369 and even worse. How do you think you can take last line of buffer, except slicing it with '\n'?


```coffeescript

    blank: (row) ->
        return @editor.getBuffer().isRowBlank(row) or
               @editor.languageMode.isLineCommentedAtBufferRow(row)

```

I trust Atom implements this efficiently (because that's the purpose of a text editor).

I would imagine that, to avoid the creation of unnecessary strings, Atom uses a scanner.

I also imagine that buffers in Atom have been implemented using a data structure that allows both the efficient edition and scanning of the text in the buffer.


> If you are so afraid of over consumption memory try to avoid calling functions (inline them!) in a cycle - they copy arguments.


The right word is "concerned".

Inlining makes no difference. Strings don't get copied (strings get created and references to the strings are passed around).


> As for me, we don't leave in 90s when "640kb is enough for everyone". Ram is cheap and our time is not. While cycles often causes problems - they aren't pure.


(bike-shedding)


> But generally I don't care. You can use any code you want. But, please, move it to separate function.


I don't understand this last point. Could you give an example?

slavaGanzin:
>I trust Atom implements this efficiently (because that's the purpose of a text editor).

https://github.com/atom/text-buffer/blob/master/src/history.coffee
As I can see they don't write something good looking or efficient.

> Inlining makes no difference. Strings don't get copied (strings get created and references to the strings are passed around).

Maybe we are talking about different things, but where did you read that?
```
a = '123'
b = (a) -> a = '1234'
b(a)
console.log a
```

I wouldn't try to promote functional programming, you can write imperative one's and achieve the same. But taking care about memory nowadays in non highloaded environments is bike-schedding for me.


> I don't understand this last point. Could you give an example?

That's easy: if you have something working based on state, and so impure - answer correlates with state; preferably to refactor it to a function - in such way you encapsulate it's side effects.

n-riesco:
On 07/03/16 16:34, Slava wrote:
>     I trust Atom implements this efficiently (because that's the purpose of a text editor).
> 
> https://github.com/atom/text-buffer/blob/master/src/history.coffee
> As I can see they don't write something good looking or efficient.

Could you be more specific? I don't know what I'm supposed to look at.

>     Inlining makes no difference. Strings don't get copied (strings get created and references to the strings are passed around).
> 
> Maybe we are talking about different things, but where did you read that?
> 
> |a = '123' b = (a) -> a = '1234' b(a) console.log a |

I don't understand what the above example is trying to illustrate.

The only thing I see is that the above piece of code allocates memory for two strings '123' and '1234'.

>     I don't understand this last point. Could you give an example?
> 
> That's easy: if you have something working based on state, and so impure - answer correlates with state; preferably to refactor it to a function - in such way you encapsulate it's side effects.

When I said that I din't understand your last point, by "your last point" I meant your last paragraph:

> But generally I don't care. You can use any code you want. But, please, move it to separate function.

slavaGanzin:
I don't want to continue this.

My last point - I'm not the one who dictates here. So do like you want.

n-riesco:
@slavaGanzin Yes, I agree that it's better that we stop discussion here.

Please, don't say things like _"you're not the one who dictates here"_. You're the person that contributed the most to Hydrogen other than Will, and your opinion is very important.

PR #187, opened 2016-03-09
Move code from Hydrogen to KernelManager

n-riesco:
* Moved startKernelIfNeeded to KernelManager.

* Moved language kernel check from createResultBubble to
  startKernelIfNeeded.

* Added property `language` to `Hydrogen`.


rgbkrk:
Awesome. Thanks for moving forward with this @n-riesco

PR #191, opened 2016-03-11
Strip Ansi color from output

### LINKED ISSUE #43: handle ANSI colors
We already do this if Python is started without a kernelspec, but we need to do this regardless. Otherwise we get ANSI color codes as in #42.

lgeiger:
Fixes #43.

Before:
old
Stripped:
stripped

### DIFF
--- a/lib/result-view.coffee
+++ b/lib/result-view.coffee
@@ -1,5 +1,6 @@
 {CompositeDisposable} = require 'atom'
 _ = require 'lodash'
+stripAnsi = require('strip-ansi')
 
 module.exports =
 class ResultView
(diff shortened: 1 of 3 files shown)


rgbkrk:
In the future we can use ansi to html here to show the colors instead.

lgeiger:
>use ansi to html

Yes. I tried to convert it to html but it didn't work at first try. So I went with stripping the colors for now ;)

rgbkrk:
There are a few packages you can try out with jupyter kernel output as well, depending on if you want to do React or not:

* https://github.com/nteract/react-jupyter-display-area
* https://github.com/nteract/jupyter-display-area

No one is maintaining the web component (2nd option), though the [underlying packages that it depends on are](https://github.com/nteract?utf8=%E2%9C%93&query=transformime), since they're used in https://github.com/jupyter/jupyter-js-notebook.

Issue #48, opened 2015-05-29
overlap problem


X-G:
Sorry about my poor English. I'm trying to report this by screenshot.
It happens when the codes are long enough and you scroll down.
![qq 20150529195641](https://cloud.githubusercontent.com/assets/6829403/7889027/0dd2f050-063d-11e5-9d93-9c9e7f649d21.png)

willwhitney:
This is a problem with the Atom API for markers. I've opened an issue at https://github.com/atom/atom/issues/6857 — I need them to fix it before I can address this, unfortunately.

willwhitney:
Or rather, I could fix it for us, but at risk of breaking other packages (as I mention on that thread). I think it's unlikely that any packages depend on the exact `z-index` of the tab bar, so if they don't get a fix done soon I'll just do that.

willwhitney:
Went ahead and did this. Shipped in `v0.4.0`!

X-G:
Thanks Will.

n-riesco:
This issue is still present in v0.7.1

n-riesco:
[This comment](https://github.com/atom/atom/issues/6857#issuecomment-128684193) shows the issue clearly.

rgbkrk:
Block decorations should be the way forward. http://blog.atom.io/2016/02/03/introducing-block-decorations.html

n-riesco:
@rgbkrk I've tried the following in 1.7.0-dev-7040035 and it doesn't seem to work:

```
$ git diff
diff --git a/lib/main.coffee b/lib/main.coffee
index 7afdb18..10b2ef0 100644
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -159,9 +159,9 @@ module.exports = Hydrogen =
         view.statusContainer.setAttribute('style', "height: #{lineHeight}px")
 
         @editor.decorateMarker marker,
-            type: 'overlay'
+            type: 'block'
             item: element
-            position: 'tail'
+            position: 'after'
 
         @markerBubbleMap[marker.id] = view
         marker.onDidChange (event) =>
```

![untitled](https://cloud.githubusercontent.com/assets/6199391/13718225/f5342d68-e7df-11e5-8f21-df6ae019e4ed.png)

lgeiger:
@n-riesco if you remove [this line](https://github.com/nteract/hydrogen/blob/master/styles/hydrogen.less#L29) the result should be displayed. But of course the styling will be totally messed up.

[...]

PR #192, opened 2016-03-11
Code Introspection

lgeiger:
Support introspection (https://jupyter-client.readthedocs.org/en/latest/messaging.html#introspection).

The kernel messaging works well.

Unfortunately the result display is very basic. I couldn't get a proper display panel to work so for now the inspection result just appears in an normal editor. So there are some major problems:
- the highlights from the atom spell checker are ugly
- the file is saved in order to allow the panel to be closed without an popup

bildschirmfoto 2016-03-11 um 15 07 57

Is this a functionality you would like to see in hydrogen?
What ist the best way to display the inspection result? And how can can this be achieved?

### DIFF
--- a/lib/kernel-manager.coffee
+++ b/lib/kernel-manager.coffee
@@ -7,12 +7,12 @@ Kernel = require './kernel'
 module.exports = KernelManager =
     runningKernels: {}
     kernelsUpdatedOnce: false
-      
+
     getAvailableKernels: ->
         kernels = _.pluck @getConfigJson('kernelspec', {kernelspecs:{}}).kernelspecs, 'spec'
         @updateKernels() unless @kernelsUpdatedOnce
         kernels
-               
+
     updateKernels: ->
       saveKernelsToConfig = (out) =>
         try
@@ -29,13 +29,13 @@ module.exports = KernelManager =
           @setConfigJson 'kernelspec', kernelspec
           atom.notifications.addInfo 'Hydrogen Kernels updated:',
             detail: (_.pluck @getAvailableKernels(), 'display_name').join('\n')
-      
+
       @kernelsUpdatedOnce = true
       child_process.exec 'jupyter kernelspec list --json --log-level=CRITICAL', (e, stdout, stderr) ->
           return saveKernelsToConfig stdout unless e
           child_process.exec 'ipython kernelspec list --json --log-level=CRITICAL', (e, stdout, stderr) ->
               saveKernelsToConfig stdout
-       
+
     getRunningKernels: ->
         return _.clone(@runningKernels)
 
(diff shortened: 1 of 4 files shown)


rgbkrk:
That's really cool.

You'll want to do an [interactive rebase against origin/master](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Changing-Multiple-Commit-Messages) so this can be merged cleanly.

n-riesco:
* I'm sure this will look even better, once [this proposal in Jupyter](https://github.com/jupyter/jupyter_client/issues/51#issuecomment-130411409) is completed.

* As Kyle suggested, I would do a `git rebase -i origin/master` and I would also squash some of the commits that add nothing to the repository history.

* I've also noticed a bad interaction between the autocomplete window and the new inspector pane. See below:

bad-css

lgeiger:
Sorry for the messy commits.

This causes the overlay issue: https://github.com/nteract/hydrogen/blob/master/styles/hydrogen.less#L11
See: https://github.com/nteract/hydrogen/issues/48

I'll give the `autocomplete-plus` overlays an higher `z-index`.

n-riesco:
@lgeiger 4d516fe fixes the overlay issue with autocomplete, but it would bring back #48

lgeiger:
@n-riesco I checked with that issue. `z-index: 5` works fine for me.

n-riesco:
@lgeiger See below

untitled

n-riesco:
@lgeiger I'm not familiar with the styling in Hydrogen. Does it make sense to set the z-indez of `.bubble-output`?

lgeiger:
Oh I think this is a general issue. On master there is the same behavior if a editor panel is split down.

@n-riesco I'm not familiar with the styling either, but I think this would be the only option.

n-riesco:
@lgeiger I've reopened #48.

This PR looks good to merge.

rgbkrk:
:rocket:

rgbkrk:
@n-riesco want to try your hand at a release today or tomorrow?

n-riesco:
@rgbkrk Service with a :smile:  !

lgeiger:
@n-riesco I solved the overlay issue by including
```
atom-text-editor {
    z-index: 0;
}
```
But this messes with other packages like `autocomlete-plus`. They now also slide behind the pane:
bildschirmfoto 2016-03-11 um 21 09 42

If that's no problem I can make a PR.

n-riesco:
@lgeiger I don't know what the implications of setting the `z-index` of `atom-text-editor` are. Perhaps @rgbkrk or @slavaGanzin know.

The partially-hidden completion list becomes a nuisance once you type code on the bottom lines.

I think we need to come up with alternative way to display Hydrogen results. Overlay markers are perfect for completion lists, but they're not meant to display Hydrogen results.

@rgbkrk @slavaGanzin What do you think?

n-riesco:
Let's continue the discussion on #48 .

rgbkrk:
We need to use the inline markers feature that is coming to Atom. They built it specifically for our use case here as well as merge conflicts (I talked with the Atom team at CodeConf about this and now it exists)

rgbkrk:
http://blog.atom.io/2016/02/03/introducing-block-decorations.html

PR #194, opened 2016-03-14
Colorful Output in the inspector and commands for closing and minimizing

lgeiger:
The ansi colors are hardcoded. In order to make them look good with every theme they all get displayed in `@text-color-info`.

@rgbkrk I tried to use `transformime` but somehow I wasn't able to get it to work all the way. However the inspect reply is just `text/plain`, at least for the python kernel. So I think this solution is fine for now.

color

I added two commands to toggle the size and to close the inspector.
Should we map a default key binding for the `inspect` command? `cmd-ctrl-i` would work great for me.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,8 +38,7 @@
     "jmp": "0.4.x",
     "strip-ansi": "3.0.1",
     "uuid": "^2.0.1",
-    "ansi-to-html": "0.4.1",
-    "atom-message-panel": "1.2.4"
+    "ansi-to-html": "0.4.1"
   },
   "consumedServices": {
     "status-bar": {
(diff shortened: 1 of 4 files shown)


lgeiger:
Oh something went wrong with my rebase. I try to resole this.

lgeiger:
Merge conflicts should be resolved now.

package.json
```diff
@@ -37,7 +37,9 @@
     "lodash": "^3.8.0",
     "jmp": "0.4.x",
     "strip-ansi": "3.0.1",
-    "uuid": "^2.0.1"
+    "uuid": "^2.0.1",
+    "ansi-to-html": "0.4.1",
+    "atom-message-panel": "1.2.4"
```
n-riesco: `atom-message-panel` is already listed (see first dependency)

lgeiger: Thanks a lot! I'm sometimes still at war with git rebase.

Issue #112, opened 2015-09-18
ipython plots not showing


schlichtanders:
Dear Hydrogen team,

I just started to use hydrogen and everything is working but the plots.
It shows ``[<matplotlib.lines.Line2D at 0x7f009fd1b950>`` instead of a nice plot

I tried plotting within qtconsole solely and get the same output. Here I have to enable ``%matplotlib inline`` to get the nice plots.
Maybe such is missing when starting the ipython kernel?

hoping for help,
best,
Stephan

willwhitney:
You'll definitely have to run `%matplotlib inline` to make them work. This is because the IPython kernel  doesn't do this itself.

I don't want to add any kernel-specific code to Hydrogen, but I do have plans to (when I get the time...) implement a language plugin system for Hydrogen that would allow code to be run when the kernel is started and/or before user code.

schlichtanders:
Thank you very much for this fast clarification - and of course for
creating such a lovely plugin.
I am looking forward to language specific configurations

a simple json configuration file keyworded by standard filenames, referring
to values which represent the lines to be executed in the beginning could
do the trick already, doesn't it? (or is this too insecure?)

willwhitney:
My rationale for doing it with extensions is that there are several pressing language-specific features* which are going to take a nontrivial amount of code. By building an API for Hydrogen extensions, any language-specific features can be independently maintained and distributed by the people who use those languages. I don't want to have any language-specific code in the core Hydrogen package, and I don't want every user to have to go add this themselves.

*most notably, a per-language function for finding the preceding block of code to execute from the buffer

schlichtanders:
thanks for explaining,
that would be a great feature,

looking forward to the future,
best,
Stephan

rgbkrk:
> My rationale for doing it with extensions is that there are several pressing language-specific features* which are going to take a nontrivial amount of code. By building an API for Hydrogen extensions, any language-specific features can be independently maintained and distributed by the people who use those languages. I don't want to have any language-specific code in the core Hydrogen package, and I don't want every user to have to go add this themselves.

> *most notably, a per-language function for finding the preceding block of code to execute from the buffer

Wow that's sounding great.

n-riesco:
Closed by #195

PR #195, opened 2016-03-15
[WIP] Execute custom code on Startup

lgeiger:
A (not very fancy) solution for https://github.com/nteract/hydrogen/issues/112.

Execute custom code when the kernel is started.

### DIFF
--- a/lib/kernel-manager.coffee
+++ b/lib/kernel-manager.coffee
@@ -132,10 +132,11 @@ module.exports = KernelManager =
         ConfigManager.writeConfigFile (filepath, config) =>
             kernel = new Kernel(kernelInfo, config, filepath)
             @runningKernels[kernelInfo.language] = kernel
-            startupCode = @getConfigJson('startupCode')[kernelInfo.language]
+            startupCode = @getConfigJson('startupCode')[kernelInfo.display_name]
             if startupCode?
+                console.log "executing startup code"
                 startupCode = startupCode + ' \n'
-                KernelManager.execute kernelInfo.language, startupCode
+                @execute kernelInfo.language, startupCode
             onStarted?(kernel)
 
     startKernelIfNeeded: (language, onStarted) ->
--- a/lib/kernel-manager.coffee
+++ b/lib/kernel-manager.coffee
@@ -132,6 +132,10 @@ module.exports = KernelManager =
         ConfigManager.writeConfigFile (filepath, config) =>
             kernel = new Kernel(kernelInfo, config, filepath)
             @runningKernels[kernelInfo.language] = kernel
+            startupCode = @getConfigJson('startupCode')[kernelInfo.language]
+            if startupCode?
+                startupCode = startupCode + ' \n'
+                KernelManager.execute kernelInfo.language, startupCode
             onStarted?(kernel)
 
     startKernelIfNeeded: (language, onStarted) ->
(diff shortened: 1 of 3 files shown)


n-riesco:
@rgbkrk I'm reviewing this PR. Please, don't merge it for the time being.

n-riesco:
@lgeiger Here's my review:

* `main.coffee` has grown too large, we should aim at moving functionality out of `main.coffee`; by this I mean:

  * If there's a choice between adding functionality to `main.coffee` or to any other file, it's preferable to add the functionality to other files.

  * For example, functionality recently added, like the Inspector, should be moved into its own file (similarly to the autocomplete provider or watch view).

* In this case, the PR touches `main.coffee`, and `kernel-manager.coffee`. There is no need to touch `main.coffee`, it is possible to limit all the changes to `kernel-manager.coffee`. I think, though, that since the startup code is only run on kernel construction, this PR should be limited to changing `kernel.coffee`.

* There is also the question whether the startup code should be specific to a language or to a kernel. For example, for Javascript, there is one kernel that won't support magics, and the other two don't support the same magics.

lib/kernel-manager.coffee
```diff
@@ -176,3 +177,6 @@ module.exports = KernelManager =
 
     destroy: ->
         _.forEach @runningKernels, (kernel) -> kernel.destroy()
+
+    getStartupCode: (language) ->
+        return @getConfigJson('startupCode')[language] + ' \n'
```
n-riesco: If no startup script is defined, this function returns the string 'undefined\n'.

lgeiger: Thanks for noticing! I added the line break at some point and didn't thought about this.

lgeiger:
1. addressed in #194 

2. I try to figure out a way to do this

3. I didn't thought about this. For me it would make sense to have a kernel specific startup code.

n-riesco:
On 15/03/16 11:20, Lukas Geiger wrote:
>  2. I try to figure out a way to do this

I think moving all the relevant code to the kernel constructor (or to a method called by the constructor) should be enough. Please, let me know if there are any difficulties.

lgeiger:
The easiest thing for me to do is moving it to `startKernel` in `kernel-manager.coffee`. It would reduce the code to just four lines. I'll just have to do some testing.

lgeiger:
Can we use use the `kernelInfo.display_name` instead of `kernelInfo.language` to achieve point 3?

n-riesco:
On 15/03/16 11:52, Lukas Geiger wrote:
> The easiest thing for me to do is moving it to |startKernel| in |kernel-manager.coffee|. It would reduce the code to just four lines. I'll just have to do some testing.

SGTM

n-riesco:
On 15/03/16 12:10, Lukas Geiger wrote:
> Can we use use the |kernelInfo.display_name| instead of |kernelInfo.language| to achieve point 3?

I think that's the easiest option I can see at the moment.

lgeiger:
@n-riesco can you test it? I don't have multiple kernels for the same language installed.

n-riesco:
@lgeiger It's working for me. I've tested it with `startupCode` configured as:

```json
{"IPython 2": "test = \"IPython 2\"", "IPython 3": "test = \"IPython 3\""}
```

and `kernelspec`:

```json
{
    "kernelspecs": {
        "ipython2": {
            "spec": {
                "display_name": "IPython 2",
                "env": {},
                "argv": [
                    "ipython",
                    "kernel",
                    "--no-secure",
                    "--colors=NoColor",
                    "-f",
                    "{connection_file}"
                ],
                "language": "python"
            }
        },
        "ipython3": {
            "spec": {
                "display_name": "IPython 3",
                "env": {},
                "argv": [
                    "ipython3",
                    "kernel",
                    "--no-secure",
                    "--colors=NoColor",
                    "-f",
                    "{connection_file}"
                ],
                "language": "python"
            }
        }
    }
}
```

n-riesco:
Shall we merge?

rgbkrk:
> I'm reviewing this PR. Please, don't merge it for the time being.

Thanks for the heads up as well as the thorough review.

PR #202, opened 2016-03-16
ResultView: Use transformime to output results

n-riesco:


n-riesco:
One thing that complicates the interaction between Kernel and all the other classes is that Kernel transforms Jupyter messages into an intermediate object of the form:

```coffeescript
                result =
                    data:   'ok'
                    type:   'text'
                    stream: 'status'
```

In future PRs, I'm planning to get rid of these intermediate objects where possible.

rgbkrk:
Ah, great.

rgbkrk:
This is great, thank you so much for taking this for a spin @n-riesco.

n-riesco:
@rgbkrk [Here](https://github.com/n-riesco/hydrogen/blob/2da9a5d23dcd31623a71cd112f008506cc02fc62/lib/result-view.coffee#L210-L224) and [here](https://github.com/n-riesco/hydrogen/blob/2da9a5d23dcd31623a71cd112f008506cc02fc62/lib/result-view.coffee#L105-L144) are the lines where tranformime is used.

After this initial experience with transformime, there are two things I would change:
- Instead of passing `document` in the call to `transform`, I could pass it to the constructor.
- After quickly going through the code in transformime, I think it would also be possible to pass `document.createElement` instead of `document`.

rgbkrk:
> I think it would also be possible to pass `document.createElement` instead of `document`

That makes sense.

> Instead of passing document in the call to transform, I could pass it to the constructor.

The reason this was broken up separately was likely for iframe usage, which no one is doing with transformime currently. I'm not opposed to that. If I changed the interface, I'd make `document` or `document.createElement` the first argument so that a user could declare `const t = transform.bind(null, document)` and use the bound function.

For our React components, the API for [transformime-react](https://github.com/nteract/transformime-react) (and by extension the display area) are quite different, very declarative, and opinionated in the usage of Immutable.js.

n-riesco:
On 16/03/16 20:42, Kyle Kelley wrote:
> The reason this was broken up separately was likely for iframe usage,

One could always create several instances of Transformime:

```js
documentTransformer = new transformime.Transformime(document.createElement, transformers);
iframeTransformer = new transformime.Transformime(iframe.contentDocument.createElement, transformers);
```

rgbkrk:
Feel free to make PRs or issues to `transformime`, I'm not opposed to that. Will be good to see what other users of `transformime` think.

lgeiger:
Great!

Unfortunately some hydrogen features aren't working for me with this PR:

- svg plots aren't showing up. Just the close button is displayed.
- ansi colors aren't rendered.
- error notifications aren't showing up.
- output font is much bigger than before.

new:
new
before:
old

n-riesco:
@lgeiger Since this PR is already merged, let's move the discussion to #207

Issue #204, opened 2016-03-17
Plotting does not work.  


dpavlic:
I have an odd problem where graphs will not show.  Sometimes they will show, but they pop up in a new window.  Most of the time, however, they just plain do not work.  I am reproducing a very simple matplotlib graph (indeed, this is on their example page) that works in qtconsole.  Code below:

`
%matplotlib inline
import pandas as pd
import matplotlib.pyplot as plt
import numpy as np

t = np.arange(0.0, 2.0, 0.01)
s = np.sin(2*np.pi*t)
plt.plot(t, s)

plt.xlabel('time (s)')
plt.ylabel('voltage (mV)')
plt.title('About as simple as it gets, folks')
plt.grid(True)
plt.savefig("test.png")
plt.show()
`

I get an empty box inline when I try to run this.  Attached is a picture.
![no_graph](https://cloud.githubusercontent.com/assets/1075079/13861801/b5de710c-ec65-11e5-892c-23a4eafb477a.png)

n-riesco:
What you see is the output of the last command `plt.show()`, which doesn't return anything.

Try adding `plt` at the end of your code:

```python
%matplotlib inline
import pandas as pd
import matplotlib.pyplot as plt
import numpy as np

t = np.arange(0.0, 2.0, 0.01)
s = np.sin(2*np.pi*t)
plt.plot(t, s)

plt.xlabel('time (s)')
plt.ylabel('voltage (mV)')
plt.title('About as simple as it gets, folks')
plt.grid(True)
plt.savefig("test.png")
plt.show()
plt
```

n-riesco:
This is working in master.

I will check if I can reproduce the issue in the package available on the Atom package repository.

n-riesco:
It's working for me on v0.7.2.

@dpavlic Could you open the dev console in Atom (press shift-ctrl-i) and run the code again.

You will see many log messages in the console. The one I would like to see is the last IO message that has the `prefix` set to `"display_data"`. This is what I see in my case:

![untitled](https://cloud.githubusercontent.com/assets/6199391/13864257/e83e39c0-ec97-11e5-89dd-1abf1893d9f8.png)

dpavlic:
Well, this definitely gets me somewhere!

`
[...]

PR #207, opened 2016-03-18
Use block decorations to show output results

### LINKED ISSUE #48: overlap problem

### LINKED ISSUE #205: Pandas dataframes cut off
Attempting to display a pandas dataframe returns a HTML representation (great!), however, scrolling around the dataframe does not seem to work: the display clips.
dataframe_issue

n-riesco:
* Requires Atom v1.6.0+

Fixes #205
Fixes #48

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -143,7 +143,7 @@ module.exports = Hydrogen =
 
         marker = @editor.markBufferPosition {
             row: row
-            column: lineLength
+            column: 0
         }, {
             invalidate: 'touch'
         }
@@ -160,9 +160,9 @@ module.exports = Hydrogen =
         view.statusContainer.setAttribute('style', "height: #{lineHeight}px")
 
         @editor.decorateMarker marker,
-            type: 'overlay'
+            type: 'block'
             item: element
-            position: 'tail'
+            position: 'after'
 
         @markerBubbleMap[marker.id] = view
         marker.onDidChange (event) =>
--- a/lib/result-view.coffee
+++ b/lib/result-view.coffee
@@ -106,7 +106,8 @@ class ResultView
             mimeType = mimetype
             htmlElement = el
 
-            console.log "ResultView: Rendering as " + mimeType
+            console.log "ResultView: Rendering as MIME", mimeType
+            console.log "ResultView: Rendering as ", htmlElement
 
             if mimeType is 'text/plain'
                 @element.classList.remove 'rich'
@@ -206,14 +207,21 @@ class ResultView
     getElement: ->
         @element
 
-
 transformime = require 'transformime'
 transformimeJupyter = require 'transformime-jupyter-transformers'
+
+SVGTransform = (mimetype, value, document) ->
+    el = document.createElement 'img'
+    el.setAttribute 'src',
+        "data:image/svg+xml;base64," + (new Buffer value).toString 'base64'
+    return el
+SVGTransform.mimetype = 'image/svg+xml'
+
 transformer = new transformime.Transformime [
     transformime.TextTransformer,
     transformimeJupyter.PDFTransform,
     transformime.ImageTransformer,
-    transformimeJupyter.SVGTransform,
+    SVGTransform,
     transformimeJupyter.consoleTextTransform,
     transformimeJupyter.LaTeXTransform,
     transformimeJupyter.markdownTransform,
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
   "repository": "https://github.com/nteract/hydrogen",
   "license": "MIT",
   "engines": {
-    "atom": ">=0.174.0 <2.0.0"
+    "atom": ">=1.6.0 <2.0.0"
   },
   "dependencies": {
     "ansi-to-html": "0.4.1",
--- a/styles/hydrogen.less
+++ b/styles/hydrogen.less
@@ -5,20 +5,6 @@
 @import "ui-variables";
 @import "syntax-variables";
 
-// this is a hack to deal with a failing of the markers API
-// https://github.com/atom/atom/issues/6857
-.tab-bar {
-    z-index: 5;
-}
-
-.bottom{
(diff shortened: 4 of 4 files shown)


n-riesco:
@lgeiger 


* Sorry about breaking the ANSI colours (I should've remembered you just implemented that). Could you open a separate issue for it?

* SVG works for me, Could you show me how to reproduce the failure you see?

* Could you open a separate issue for the error notifications?

* I will address the issue with the font size in this PR

lgeiger:
Awesome!

This isn't working for me in Atom 1.6.0
```
%matplotlib inline
%config InlineBackend.figure_format = 'svg'
import matplotlib.pyplot as plt
import numpy as np

t = np.arange(0.0, 2.0, 0.01)
s = np.sin(2 * np.pi*  t)

plt.plot(t, s)
plt.show()
```

n-riesco:
@lgeiger For the issue with SVG, could you try to check the logs as described in #204 ?

lgeiger:
Hmm I can't find an IO message with `prefix` `"display_data"`:
bildschirmfoto 2016-03-18 um 20 25 02

lgeiger:
Side note:
Do you think we can move the checkmarks inline despite being block decorations?

n-riesco:
@lgeiger Thanks for the code. Now I can reproduce it.

n-riesco:
@lgeiger, re checkmarks: I don't think so. The block decorations don't even live in the same `div` as the code line.

n-riesco:
@lgeiger The problem with SVGs is triggered with SVG that do not start with an <svg> tag. In your example, if I remove the first few line, Hydrogen shows the image:

```svg
<?xml version="1.0" encoding="utf-8" standalone="no"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN"
  "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<!-- Created with matplotlib (http://matplotlib.org/) -->
<svg height="255pt" version="1.1" viewBox="0 0 378 255" width="378pt" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
...
</svg>
```

n-riesco:
@lgeiger Is this font size ok?

untitled

lgeiger:
looks good to me :+1:

rgbkrk:
Woo, awesome!!!

n-riesco:
@rgbkrk The problem displaying SVGs was caused by `transformime-jupyter-transformers`. The SVG transformer converts the SVG data into a DOM element [here](https://github.com/nteract/transformime-jupyter-transformers/blob/140291e2440bf33a125a6345e3e73e857420ae43/src/svg.transform.js#L7-L8) but it only returns [the first child](https://github.com/nteract/transformime-jupyter-transformers/blob/140291e2440bf33a125a6345e3e73e857420ae43/src/svg.transform.js#L9), which often is just a comment.

I see two ways to fix this:
* as I have done [in Hydrogen](https://github.com/n-riesco/hydrogen/blob/block-decorations/lib/result-view.coffee#L213-L218)
* or alternatively, return the `<div>` element with all the child nodes (but this isn't friendly with styles).

rgbkrk:
Yuck, thanks for finding the problem @n-riesco. Seems like with your solution, transformime for jupyter should behave just like the rest of the image transforms.

rgbkrk:
Related project to dig through (not necessarily rely on), for Jupyter output in general: https://github.com/nteract/jupyter-display-area

rgbkrk:
Seems like we're definitely gearing up for a major release here.

slavaGanzin:
Hello everyone, I accidentally make git pull because wanted to fix #205. And now I have no result bubbles at all. Should I recompile anything or ...?

And if we decided to move to block decorations, I think we shouldn't use different font(which was great in visual separating hydrogen popups from code) and use user defined.

n-riesco:
On 19/03/16 12:52, Slava wrote:
> Hello everyone, I accidentally make git pull because wanted to fix #205 <https://github.com/nteract/hydrogen/issues/205>. And now I have no result bubbles at all. Should I recompile anything or ...?

There are new dependencies. Try `apm install` on the Hydrogen folder.

> And if we decided to move to block decorations,

I'm OK with trying other solutions, but we do need to move away from overlays because they're meant for auto-completion lists.

> I think we shouldn't use different font(which was great in visual separating hydrogen popups from code) and use user defined.

Transformime wraps 'text/plain' with a `<pre>` tag. If you want to play with the style, have a look at these lines:

https://github.com/nteract/hydrogen/blob/104d8cee62b53c78e01d4b37b80b208be936abc6/styles/hydrogen.less#L109-L117

and

https://github.com/nteract/hydrogen/blob/104d8cee62b53c78e01d4b37b80b208be936abc6/styles/hydrogen.less#L133-L141

slavaGanzin:
I cloned hydrogen, make apm install and nope it doesn't work for me even with default atom styling. Result blocks aren't visible.

> Transformime wraps 'text/plain' with a `<pre>` tag. If you want to play with the style, have a look at these lines:

I think it would be better do not use <pre> in transformime and just use span with:
```
.preformatted {
    white-space: pre;
}
```
Should I describe why?

rgbkrk:
Make the PR and explain there. I did the same with spans inside a code block in nteract/nteract, when using ansi-to-react.

n-riesco:
What Atom version are you using?
Block decorations were introduced in Atom v1.6.0.

slavaGanzin:
@n-riesco Accidentally 1.4, but my package system was quite sure that 1.6. Thanks for that point

n-riesco:
@slavaGanzin Note also that Atom v1.6.0 broke embedding images (see #204 ).

slavaGanzin:
@n-riesco Already noticed :+1:

PR #210, opened 2016-03-19
Render text with Jupyter.consoleTextTransform

lgeiger:


n-riesco:
Previously, error results were coloured in red. I really liked this feature.

The changes in `styles/hydrogen.less` remove the colour styling in error messages.

Other than that, this PR is a neat way to use `ansi-to-html` via `transformime`.

rgbkrk:
I'm a fan of this because I like my tracebacks in the native colors.

lgeiger:
for me they are now displayed like this:
bildschirmfoto 2016-03-19 um 21 17 56

Actually I override the ansi colors with red. Because they don't really fit in with the syntax color:
bildschirmfoto 2016-03-19 um 21 20 28

lgeiger:
I could add an red border to make it clear.

n-riesco:
This is how it looks look with the IJavascript kernel (which doesn't use ANSI colours):

untitled

rgbkrk:
If someone uses a library that outputs ANSI colors, what happens with IJavaScript?

n-riesco:
@rgbkrk I expect it to work.

I've just tested it. It's working in the notebook:

notebook

but it doesn't in Hydrogen:

hydrogen

lgeiger:
Thats because `transformime.TextTransform` doesn't handle ansi color codes.

I would propose replacing the text transformer with a combination of `TextTransform`and `consoleTextTransform`. This works for me in Hydrogen:
```
Convert = require 'ansi-to-html'
hasAnsi = require 'has-ansi'

TextTransform = (mimetype, data, document) ->
    if hasAnsi(data) is true
        convert = new Convert
            escapeXML: true
            newLine: true
        el = document.createElement('pre')
        el.innerHTML = convert.toHtml(data)
        return el
    else
        el = document.createElement('pre')
        el.textContent = data
        return el

TextTransform.mimetype = ['text/plain']
```
I think this should be handled directly via `transformime`.

rgbkrk:
> Or we remove the hasAnsi check and simply render all text/plain messages as jupyter/console-text.

I'm certainly a fan. That's what the notebook does too.

lgeiger:
- All text is rendered with Jupyter.consoleTextTransform
- Errors are displayed like before.

bildschirmfoto 2016-03-20 um 14 09 27

n-riesco:
@lgeiger @rgbkrk I need to your help. In my machine, the background colours are ignored. I've boiled down the problem to `ansi-to-html`. For example:

```js
> var Convert = require('ansi-to-html')
undefined
> convert = new Convert({excapeXML: true, newLine:true})
Filter {
  opts: 
   { fg: '#FFF',
     bg: '#000',
     newline: false,
     escapeXML: false,
     stream: false,
     excapeXML: true,
     newLine: true },
  input: [],
  stack: [],
  stickyStack: [] }
> convert.toHtml("\u001b[48;5;6m\u001b[38;5;12mtest\u001b[39m\u001b[49m")
'<blink><blink><span style="color:#55F">test<span style="color:#FFF"><span style="background-color:#000"></span></span></span></blink></blink>'
```

What output do you get when the above code is run?

lgeiger:
I don't have a Javascript kernel on my system so I can't test your code.

With python and hydrogen I get with this code the above seen output:
```
from __future__ import print_function

ESC = '\x1b['
RESET = ESC + "00m"

for bg in range(40,48):
    for fg in range(30,38):
        print ("{ESC}01;{bg};{fg}mtext {RESET}".format(**locals()), end='')
    print ()
```

rgbkrk:
Interesting. I'll give that a shot here shortly.

n-riesco:
@lgeiger Thanks for the code. That gave me a hint to understand the problem.

It looks like a bug in `ansi-to-html`, for example, it works with plain ANSI colours:

```js
> convert.toHtml("\u001b[30;40mtest\u001b[00m")
'<span style="color:#000"><span style="background-color:#000">test</span></span>'
```

but it gets confused with xterm colours:

```js
> convert.toHtml("\u001b[48;5;0m\u001b[38;5;0mtest\u001b[39m\u001b[49m")
'<blink></blink><span style="color:#000">test<span style="color:#FFF"><span style="background-color:#000"></span></span></span>'
```

rgbkrk:
Interesting!

Any ideas @IonicaBizau?

We're using `ansi-up` in another code base (for the React based stuff). Seems like we would need to extend those. I'm noticing that my terminal type on my local box doesn't produce ANSI coloring in the jupyter notebook yet the same code does on tmpnb.org.

IonicaBizau:
@rgbkrk I don't really know the context, but why do you want to convert ANSI into HTML and not using `term.js`?

It has the `write` method where you can *write* terminal data (that means **any** ANSI snippets). That would make it like a real terminal (including cli animations and so on). :grin:

rgbkrk:
The backends (kernels) these are built on send messages that have text with ANSI code. It's small enough most of the time and very static that it's fairly cheap to convert to a straight markup. Otherwise we have to manage _many_ term.js instances on a page. It would be cool if someone tried it out regardless of whether I think that's an expensive operation or not.

rgbkrk:
Now that I'm thinking about it, I see how the flow would be with term.js. Each time we send execution to a kernel, we would create a DOM node for term.js and everytime we got [stdout or stderr stream](http://jupyter-client.readthedocs.org/en/latest/messaging.html#streams-stdout-stderr-etc) we would `term.write`.

IonicaBizau:
I don't think that using term.js would be a bad idea in terms of performance. It does pretty much the same thing: displaying ANSI in HTML, but the difference is that it does act like a real terminal.

slavaGanzin:
https://github.com/chalk/strip-ansi ?

PR #213, opened 2016-03-20
Style Block decorations 

lgeiger:
I started playing around with the block decorations. The next few days I will work a lot with hydrogen so I can test this PR and probably I will find some more styling problems. What do you think about the styling so far?
New:
pr
Master:
master

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -158,6 +158,9 @@ module.exports = Hydrogen =
         view.spinner.setAttribute('style',
                 "width: #{lineHeight + 2}px; height: #{lineHeight - 4}px;")
         view.statusContainer.setAttribute('style', "height: #{lineHeight}px")
+        element.setAttribute('style',
+                "margin-left: #{lineLength + 1}ch;
+                margin-top: -#{lineHeight}px")
 
         @editor.decorateMarker marker,
             type: 'block'
--- a/styles/hydrogen.less
+++ b/styles/hydrogen.less
@@ -52,7 +52,7 @@
         background-color: @bubble-background;
         float: left;
         margin-left: 20px;
-        box-shadow: 0 0 2px 1px rgba(0,0,0,0.2);
+        box-shadow: 0 0 3px 0 rgba(0,0,0,0.3);
         .bubble-output-container {
             display: inline;
             pre {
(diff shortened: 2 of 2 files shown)


n-riesco:
I like the shadow in the multiline bubbles. Could we have them as well in the in the single line bubbles too?

lgeiger:
bildschirmfoto 2016-03-21 um 00 20 31

rgbkrk:
Think this is fit to ship now?

n-riesco:
How does it work? Is it the `float:left` what makes the block decoration to be on the same line?

lgeiger:
@rgbkrk There are just some minor inconsistencies but nothing major. If you like you can merge and I fix these in a future PR.

@n-riesco It's the negative [`margin-top`](https://github.com/nteract/hydrogen/pull/213/files#diff-2e1bb36917ff76ab7308da52bc968501R163) in combination with `float: left`. I just played around with the values until it worked ;).

n-riesco:
Neat!

rgbkrk:
We fit to take this in now @n-riesco?

lgeiger:
I just noticed a little problem with the watches panel. It'll be ready in a few minutes ;)

rgbkrk:
Ah, sorry. I noticed the `[WIP]` tag coming off and assumed that was you stating intent. :smile:

lgeiger:
I'm  good with merging it!

You are right. At the time I removed the `[WIP]` I thought it was ready. :laughing:

n-riesco:
@lgeiger Check the SVG example you gave me.

I gotta go offline for a while.

lgeiger:
@n-riesco The SVG output looks good to me.

n-riesco:
@lgeiger I've tested this branch and works well in my machine.

PS: The problem I saw before was actually caused by the beta version of the IJavascript kernel.

lgeiger:
@n-riesco no problem. Thanks for the testing.

rgbkrk:
Anything else we want to do before we make a release?

Were people willing to try out `zmq-prebuilt`? It has prebuilds for Linux, OS X, and 32-bit Windows. 64-bit Windows users still have to build (as they would in the past too).

n-riesco:
@rgbkrk This is now just a question of replacing the `jmp` dependency with `jmp-prebuilt`.

PR #214, opened 2016-03-21
Use Transformime v2.2.0

n-riesco:


n-riesco:
@rgbkrk Please, have a look at how I've implemented [SVGTransform](https://github.com/n-riesco/hydrogen/blob/transformime-v2.2.0/lib/result-view.coffee#L212-L222).

The previous implementation didn't work with:

```js
$$svg$$ = "<svg><rect width=80 height=80 style='fill: orange;'></rect></svg>";
```

If you're happy with the new implementation, I will push PR to the `transformime-jupyter-transformers
` repository. Also, please, have a look at [the error message](https://github.com/n-riesco/hydrogen/blob/transformime-v2.2.0/lib/result-view.coffee#L218), in case you'd like something else.

rgbkrk:
Looks good! It won't be able to use embedded scripts in SVG when using innerHTML, though it should display nicely.

n-riesco:
@rgbkrk I've tested the new and the previous `SVGTransform` and none of them run the embedded JS. Am I doing something wrong? Did the previous `SVGTransform` run embedded JS?

The previous `SVGTransform` doesn't try to run the embedded JS, and the new `SVGTransform` tries to run it but fails the Content Security Policy (CSP).

After further testing, I've noticed that, unlike the Jupyter notebook, I can't inject JS at all, not even in `text/html` responses. I think this is all due to the use CSP.

rgbkrk:
> I've tested the new and the previous SVGTransform and none of them run the embedded JS. Am I doing something wrong? Did the previous SVGTransform run embedded JS?

No, I was just stating what I saw. This is good.

Honestly, I'm glad about the default CSP in Atom. We likely need to figure out a secure way to sandbox.

/cc @jdfreder

n-riesco:
On 21/03/16 17:00, Kyle Kelley wrote:
>     I've tested the new and the previous SVGTransform and none of them run the embedded JS. Am I doing something wrong? Did the previous SVGTransform run embedded JS?
> 
> No, I was just stating what I saw. This is good.

:) Just making sure I didn't introduce any regressions.

PR #215, opened 2016-03-22
ResultView: Yet another SVGTransform

n-riesco:
* Try harder to find the SVG element. Use `getElementsByTagName` instead
  of `firstElementChild`.


n-riesco:
@rgbkrk If it's OK by you, this is the implementation of SVGTransform that I want to push to `transformime-jupyter-transformers`.

rgbkrk:
Sounds good to me.

PR #219, opened 2016-03-23
Kernel: Revert `jupyter/console-text` hack

n-riesco:
* This hack isn't necessary after adding `text/plain` to
  `consoleTextTransform`.


lgeiger:
Thanks I forgot about that :+1:

rgbkrk:
Great!

PR #221, opened 2016-03-23
Handle ansi colors in error messages

lgeiger:
Another ansi color related PR:
Tracebacks will be displayed in the native colors except for plain text messages. They will be display in red.

### DIFF
--- a/lib/result-view.coffee
+++ b/lib/result-view.coffee
@@ -138,6 +138,11 @@ class ResultView
 
             container.appendChild htmlElement
 
+            if @errorContainer.getElementsByTagName('span').length is 0
+                @errorContainer.classList.add('plain-error')
+            else
+                @errorContainer.classList.remove('plain-error')
+
         onError = (error) ->
             console.error "ResultView: Rendering error:", error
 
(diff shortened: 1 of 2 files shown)


n-riesco:
Works for me.

PS: It never struck me before how "colourful" IPython errors are! :stuck_out_tongue_winking_eye:

lgeiger:
@n-riesco I thought about remapping the ansi colors to more subtle ones:
```
span[style="color:#A00"] {
    color: @error-base-color !important;
}
```
 Or is there a better way of doing this?

n-riesco:
I think having code specific to kernel is going to makes Hydrogen more difficult to maintain.

In theory, the kernel has an option for dark backgrounds `--colors=Linux` that add a subtle `<b>`:
linux-colour-theme

and this is without:
lightbg-colour-theme

There are some (fairly old) issues open about the colour scheme in the IPython repository:
https://github.com/ipython/ipython/issues/2281
https://github.com/ipython/ipython/issues/7931

lgeiger:
Thanks for the links. Color schemes handled by the kernel would be great.

This is actually no kernel specific code I added. I remapped the colors from the `transformime` output respectively [`ansi-to-html`](https://github.com/rburns/ansi-to-html/blob/master/src/ansi_to_html.coffee#L9).
But I agree it would be hard to maintain this code if `transfomime` changes ever so slightly.

I'll remove my last commit.

PR #225, opened 2016-03-26
Refactor KernelManager

n-riesco:
* Simplified API by reducing the number of methods.

### DIFF
--- a/lib/autocomplete-provider.coffee
+++ b/lib/autocomplete-provider.coffee
@@ -34,25 +34,26 @@ module.exports = AutocompleteProvider = do ->
 
         # Required: Return a promise, an array of suggestions, or null.
         getSuggestions: ({editor, bufferPosition, scopeDescriptor, prefix}) ->
-            console.log "suggestions requested for prefix:", prefix
-            prefix = @getPrefix(editor, bufferPosition)
-            console.log "new prefix:", prefix
-            language = editor.getGrammar().name.toLowerCase()
-
-            hasKernel = KernelManager.languageHasRunningKernel(language)
-            if prefix.trim().length < 3 or not hasKernel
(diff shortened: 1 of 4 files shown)


lib/autocomplete-provider.coffee
```diff
@@ -34,25 +34,26 @@ module.exports = AutocompleteProvider = do ->
...
+
+            return new Promise (resolve) ->
+                kernel.complete prefix, (matches) ->
+                    matches = _.map matches, (match) ->
+                        text: match
+                        replacementPrefix: prefix
+                        iconHTML: "<img
+                            src='#{__dirname}/../static/logo.svg'
+                            style='width: 100%;'>"
+                    resolve(matches)
```
rgbkrk: Much simpler here.

lib/main.coffee
```diff
@@ -99,17 +99,18 @@ module.exports = Hydrogen =
...
 
         else if request is 'restart-kernel'
-            KernelManager.destroyKernelForLanguage language
+            KernelManager.destroyRunningKernelForLanguage language
             @clearResultBubbles()
             KernelManager.startKernelIfNeeded language
 
         else if request is 'switch-kernel'
-            KernelManager.destroyKernelForLanguage language
+            KernelManager.destroyRunningKernelForLanguage language
```
rgbkrk: I like the verbosity of this function name, makes more sense.

n-riesco: @rgbkrk The naming in this class is something we need to revisit. I added `Running` to hint that this method must be invoked. If a kernel is destroyed by calling `kernel.destroy()` directly, 'KernelManager` won't be aware of it and will keep it listed amongst its running kernels.

PR #226, opened 2016-03-26
Move get/setConfigJson to Config

n-riesco:
* Moved methods, getConfigJson and setConfigJson, from KernelManager to
  Config.


lgeiger:
Looks good to me :+1:

PR #227, opened 2016-03-27
Add the ability to define code cells

lgeiger:
A cell is defined by adding a breakpoint.
Should we add a shortcut for the commands?

### DIFF
--- a/lib/cell-manager.coffee
+++ b/lib/cell-manager.coffee
@@ -1,0 +1,50 @@
+_ = require 'lodash'
+
+module.exports = CellManager =
+    editor: atom.workspace.getActiveTextEditor()
+
+    removeAllBreakPoints: ->
+        decorations = @editor.getLineNumberDecorations({class: 'break-point'})
+        for decoration in decorations
+            decoration.marker.destroy()
+
+    removeLatestBreakPoint: ->
+        decorations = @editor.getLineNumberDecorations({class: 'break-point'})
+        if decorations.length isnt 0
+            _.last(decorations).marker.destroy()
+
+    addBreakPoint: ->
+        row = @editor.getLastCursor().getBufferRow()
+
+        marker = @editor.markBufferPosition {
+            row: row
+            column: 0
+        }, {
+            invalidate: 'never'
+        }
+        @editor.decorateMarker marker,
+            type: 'line-number'
+            class: 'break-point'
+
+    getCurrentCell: ->
+        row = @editor.getLastCursor().getBufferRow()
+        decorations = @editor.getLineNumberDecorations({class: 'break-point'})
+        breakPoints = []
+        for decoration in decorations
+            breakPoints.push(decoration.marker.getStartBufferPosition()['row'])
+
+        endArr = breakPoints.filter (x) -> x >= row
+        startArr = breakPoints.filter (x) -> x < row
+
+        if startArr.length is 0
+            start = 0
+        else
+            start = Math.max(startArr...) + 1
+
+        if endArr.length is 0
+            end = @editor.getLastBufferRow()
+        else
+            end = Math.min(endArr...)
+
+        console.log "CellManager: Cell [start, end]:", [start, end]
+        return [start, end]
--- a/lib/cell-manager.coffee
+++ b/lib/cell-manager.coffee
@@ -1,19 +1,19 @@
 _ = require 'lodash'
 
 module.exports = CellManager =
-    removeAllBreakPoints: ->
+    removeAllBreakpoints: ->
         editor = atom.workspace.getActiveTextEditor()
-        decorations = editor.getLineNumberDecorations({class: 'break-point'})
+        decorations = editor.getLineNumberDecorations({class: 'breakpoint'})
         for decoration in decorations
             decoration.marker.destroy()
 
-    removeLatestBreakPoint: ->
+    removeLatestBreakpoint: ->
         editor = atom.workspace.getActiveTextEditor()
-        decorations = editor.getLineNumberDecorations({class: 'break-point'})
+        decorations = editor.getLineNumberDecorations({class: 'breakpoint'})
         if decorations.length isnt 0
             _.last(decorations).marker.destroy()
 
-    addBreakPoint: ->
+    addBreakpoint: ->
         editor = atom.workspace.getActiveTextEditor()
         row = editor.getLastCursor().getBufferRow()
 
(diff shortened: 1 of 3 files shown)


rgbkrk:
That's interesting. Is this demarcating chunks of code to be a code cell like in the notebook?

lgeiger:
Exactly!

n-riesco:
I find this functionality very useful to edit and re-run a chuck of code. But without shortcuts, it becomes a bit of pain very quickly.

Perhaps, this is the time to re-think the key-bindings. Currently, we set:
- S-A-enter (Run)
- C-A-enter (Run-and-move-below)
- S-C-A-enter (Run-all)

I remember Will said that he chose S-A-enter to avoid collisions with other key bindings.

I've checked and in my machine, in the editor the following key bindings are free:
- A-enter, S-enter, S-A-enter, S-C-A-enter
- A-space, S-space, S-A-space, C-A-space, S-C-space, S-C-A-space
- A-tab, C-tab, S-A-tab, C-A-tab, S-C-A-tab

And key bindings already assigned:
- C-enter (insert-line-below), S-C-enter (insert-line-above)
- C-space (autocomplete)
- tab (indent), S-tab (outdent)

Note that we're currently shadowing S-C-enter (that's my bad!).

Unless there are other collisions I'm not aware of, I would support a move to use the same key bindings the notebook and nteract are using, i.e.:
- C-enter (Run)
- S-enter (Run-and-move-down)
We could use A for the corresponding cell versions:
- A-C-enter (Run-cell)
- A-S-enter (Run-cell-and-move-down)

We could maintain:
- S-C-A-enter (Run-all)

And to toggle breakpoints, perhaps, we could do S-enter or S-space.

lib/cell-manager.coffee
```diff
@@ -0,0 +1,50 @@
+_ = require 'lodash'
+
+module.exports = CellManager =
+    editor: atom.workspace.getActiveTextEditor()
```
n-riesco: Does this break when multiple files are open?

lib/cell-manager.coffee
```diff
@@ -0,0 +1,50 @@
...
+
+    addBreakPoint: ->
+        row = @editor.getLastCursor().getBufferRow()
+
+        marker = @editor.markBufferPosition {
+            row: row
+            column: 0
+        }, {
+            invalidate: 'never'
+        }
```
n-riesco: this works well without brackets. Try http://coffeescript.org/

lib/cell-manager.coffee
lgeiger: Oh no. This breaks when you want to add break points in multiple files.

lib/cell-manager.coffee
lgeiger: Thanks! I copied it [here](https://github.com/nteract/hydrogen/blob/master/lib/main.coffee#L147) but that's unnecessary.

lgeiger:
> C-enter (Run)
S-enter (Run-and-move-down) We could use A for the corresponding cell versions:
A-C-enter (Run-cell)
A-S-enter (Run-cell-and-move-down)
S-C-A-enter (Run-all)

This is a great Idea!
I would suggest using S-backspace for removing the latest break point and S-A-backspace for removing all break points if that's not to much.

rgbkrk:
Was there more you wanted to think on before we introduce this @n-riesco?

n-riesco:
@rgbkrk If no one opposes the change of key-bindings, let's go for it!

I guess this also means bumping the minor version and updating the README.md.

rgbkrk:
Ah, that's going to break user space then... would we want to shelve this until a major release?

lgeiger:
What do you think about including this in the next minor release and change the key-bindings not until `1.0.0`?

n-riesco:
Or how about having a setting to enable changes like the new key bindings that may be introduced in the next major release?

It'd give users a chance to complain before we finally introduce the changes.

rgbkrk:
That seems like a good way to start trying this out.

nils-werner:
Is there a way to define cells in the code itself? Similar to what Matlab does with

    Matlab Cell 1
    %%
    Matlab Cell 2

e.g.

    Hydrogen Cell 1
    # ==
    Hydrogen Cell 2

where `#` is the language specific comment symbol.

n-riesco:
@nils-werner This is something that would be possible once Hydrogen puts in place an API to develop plugins.

lgeiger:
This PR is sitting here for a while now. How do we proceed with this?
The only point left are new keybindings for 1.0.0 but this has nothing to do with this functionality of this PR. 
I would suggest we merge and add the keybindings later. What do you think?

lib/cell-manager.coffee
```diff
@@ -0,0 +1,51 @@
+_ = require 'lodash'
+
+module.exports = CellManager =
+    removeAllBreakPoints: ->
```
n-riesco: I feel code `Cell` might not be the best word.
@rgbkrk @willwhitney any suggestions from the native speaker?

lib/cell-manager.coffee
```diff
@@ -0,0 +1,51 @@
...
+
+    addBreakPoint: ->
+        editor = atom.workspace.getActiveTextEditor()
+        row = editor.getLastCursor().getBufferRow()
+
+        marker = editor.markBufferPosition
+            row: row
+            column: 0
+            invalidate: 'never'
+
```
n-riesco: Isn't the signature `::markBufferPosition(position, [options])`?

https://atom.io/docs/api/v1.8.0/TextEditor#instance-markBufferPosition

lib/main.coffee
```diff
@@ -153,12 +159,10 @@ module.exports = Hydrogen =
         buffer = @editor.getBuffer()
         lineLength = buffer.lineLengthForRow(row)
 
-        marker = @editor.markBufferPosition {
+        marker = @editor.markBufferPosition
             row: row
             column: lineLength
-        }, {
             invalidate: 'touch'
-        }
```
n-riesco: Isn't the signature `::markBufferPosition(position, [options])`?

lib/cell-manager.coffee
lgeiger: Hmm it doesn't throw an error.

I removed the brackets because of your [comment](https://github.com/nteract/hydrogen/pull/227#discussion_r57532735).
What would be the right way for this in coffeescript?

n-riesco: I think this should work:
```coffeescript
        marker = @editor.markBufferPosition
            row: row
            column: 0
        ,
            invalidate: 'never'
```

rgbkrk:
> I would suggest we merge and add the keybindings later. What do you think?

Works for me.

lib/cell-manager.coffee
rgbkrk: I'm fine with it because it has a close parallel to the notebook cells. No real suggestions from me here.

n-riesco: The concern I have is that `cell` is notebook-speak (some users won't be familiar with jupyter notebooks). How about renaming the commands to:
- `hydrogen:run-breakpoint-cell`
- `hydrogen:run-breakpoint-cell-and-move-down`
- `hydrogen:add-breakpoint`
- `hydrogen:remove-all-breakpoints`
- `hydrogen:remove-latest-breakpoint`

I think it's OK to write [breakpoint](http://www.yourdictionary.com/breakpoint) together. I guess the corresponding methods would need renaming.

lgeiger: `hydrogen:run-breakpoint-cell` sounds a bit odd to me.

An alternative to `cell` would be `code-cell` (used by spyder) or `section`/`code-section` (I think matlab uses this).

PR #231, opened 2016-03-28
ResultView: Show status if `transformime` failed

n-riesco:
* If `transformime` fails to transform the result, then ensure that at
  least the status container is shown. Otherwise, the block decorations
  break.

### DIFF
--- a/lib/result-view.coffee
+++ b/lib/result-view.coffee
@@ -88,10 +88,6 @@ class ResultView
                 @statusContainer.style.display = 'inline-block'
             return
 
-        console.log "ResultView: Hide status container"
-        @_hasResult = true
-        @statusContainer.style.display = 'none'
-
         if result.stream is 'stderr'
             container = @errorContainer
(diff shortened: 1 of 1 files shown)


n-riesco:
Oops! Forgot to move the log message... hold on a sec

n-riesco:
OK, this PR is now ready to merge.

lgeiger:
LGTM. But I couldn't really test it because I can't get `transformime` to fail :smiley:

n-riesco:
@lgeiger Well, you still could test whether the PR doesn't break Hydrogen when `transformime` works! :p 

It's easy with IJavascript. IPython makes it harder by parsing the input to the display functions.

@rgbkrk +1 to patch release.

lgeiger:
>  Well, you still could test whether the PR doesn't break Hydrogen when transformime works! :p

Works like a charm! :smile:

rgbkrk:
Shipped `v0.8.1`!

PR #232, opened 2016-03-28
ResultView: Add action panel to text/html output

### LINKED ISSUE #26: make layout determined by actual size of the bubble and wrap lines
This is to better accommodate kernels like IRkernel which output rich text (e.g. HTML) for even small results.

Bubbles which only contain text (regardless of formatting) should have a multiline-result style bubble (pictured below) if they're larger than a single line, and should fit inline with no close button if they're small.

HTML results should also have the same close/new pane buttons as text ones if they only contain text.

lgeiger:
- Closes #26 
- ~~Just make the `pre` element clickable if it's a single line output.~~

### DIFF
--- a/lib/result-view.coffee
+++ b/lib/result-view.coffee
@@ -130,13 +130,14 @@ class ResultView
                 @element.classList.add 'rich'
                 @setMultiline true
 
+            container.appendChild htmlElement
+
             if mimeType is 'text/html'
-                container.classList.add('html')
+                if @getAllText() isnt ''
+                    @element.classList.remove 'rich'
 
             if mimeType is 'image/svg+xml'
                 container.classList.add('svg')
-
-            container.appendChild htmlElement
 
             if @errorContainer.getElementsByTagName('span').length is 0
                 @errorContainer.classList.add('plain-error')
--- a/lib/result-view.coffee
+++ b/lib/result-view.coffee
@@ -130,6 +130,8 @@ class ResultView
                 @element.classList.add 'rich'
                 @setMultiline true
 
+            # @getAllText must be called after appending the htmlElement
+            # in order to obtain innerText
             container.appendChild htmlElement
 
             if mimeType is 'text/html'


n-riesco:
I see a bug but I can't reproduce it reliably.

When it fails:
untitled-bug

When it works:
untitled-no-bug


The logs in the dev console look the same.

I will look at this more carefully. But I'm not sure if I'll be able to look at this tomorrow.

lgeiger:
Thanks for taking a look.
I can't reproduce this. I always get the second output no matter how many times I try.

n-riesco:
It happens more often when I use C-A-enter.

lgeiger:
Strangely enough this doesn't happen on my machine.

lgeiger:
Does it happen when you add back the `previousText` check at line 111 like this?

```
previousText = @getAllText()
container.appendChild htmlElement

if mimeType is 'text/plain'
    text = @getAllText()
    if previousText is '' and text.length < 50 and text.indexOf('\n')
```

n-riesco:
@lgeiger I that was it. After adding the check, I don't it to fail.

lib/result-view.coffee
```diff
@@ -108,36 +108,37 @@ class ResultView
...
+                text = @getAllText()
+                if text.length < 50 and text.indexOf('\n') is -1
+                    if @element.classList.contains('rich') is false
+                        @setMultiline false
+
+                    try
+                        pre = container.getElementsByTagName('pre')[0]
+                        @tooltips.add atom.tooltips.add pre,
+                            title: "Copy to clipboard"
+                        pre.onclick = =>
```
n-riesco: My linter says this fat arrow is unnecessary.

lgeiger:
It should work now.

lib/result-view.coffee
```diff
@@ -108,36 +108,39 @@ class ResultView
...
-                        title: "Copy to clipboard"
-
-                    container.onclick = =>
-                        atom.clipboard.write @getAllText()
-                        atom.notifications.addSuccess "Copied to clipboard"
+                    if @element.classList.contains('rich') is false
+                        @setMultiline false
+
+                    try
+                        pre = container.getElementsByTagName('pre')[0]
```
n-riesco: Why `pre`? Doesn't it work with `container`?

lib/result-view.coffee
```diff
@@ -108,36 +108,39 @@ class ResultView
...
+                        @tooltips.add atom.tooltips.add pre,
+                            title: "Copy to clipboard"
+                        pre.onclick = ->
+                            atom.clipboard.write text
+                            atom.notifications.addSuccess "Copied to clipboard"
+                    catch
```
n-riesco: I would report this to the console (to help us debug)
```coffeescript
catch err
    console.log "ResultView:", err
```

lib/result-view.coffee
lgeiger: For example `matplotlib` shows your plot like this if `plt.show()`isn't used:
bildschirmfoto 2016-04-01 um 18 23 46

With `container` the image would be clickable as well despite just copying the text.

lib/result-view.coffee
lgeiger: You're right!
I don't know if this `try` is really necessary but I added it because of this line `container.getElementsByTagName('pre')[0]`.

n-riesco:
I think it's easier for the user to be able to click anywhere container.

Another problem with the `pre` hack is that it breaks when a block with multiple stdouts is executed, e.g.:
```python
print('A plot')
plt.plot(t, s)
```

lgeiger:
OK I'll revert my changes except those directly related to #26.

lgeiger:
All right, back to the basics :wink:

lib/result-view.coffee
```diff
@@ -130,14 +130,15 @@ class ResultView
                 @element.classList.add 'rich'
                 @setMultiline true
 
+            container.appendChild htmlElement
```
n-riesco: Isn't better to append the element after it has been fully styled?

lib/result-view.coffee
```diff
@@ -130,14 +130,15 @@ class ResultView
...
+
             if mimeType is 'text/html'
-                container.classList.add('html')
+                if @getAllText() isnt ''
+                    @element.classList.remove 'rich'
```
lgeiger: @n-riesco but then this wouldn't work.

n-riesco: I don't understand. Why wouldn't?

lgeiger: Because `@getAllText()` would always return `''`.

It gets the text from the `container` see [here](https://github.com/nteract/hydrogen/blob/master/lib/result-view.coffee#L152)

n-riesco: Thanks, I've just got it!

n-riesco: Could you add a comment [here](https://github.com/lgeiger/hydrogen/blob/issue-26/lib/result-view.coffee#L136) to remind our future selves that `@getAllText()` must be called after appending the htmlElement?

n-riesco: Wait, better put the comment above the line with  `container.appendChild htmlElement`

lgeiger: :+1:

n-riesco:
LGTM

PR #241, opened 2016-03-30
Styles: reposition rich close button

lgeiger:
PR:
bildschirmfoto 2016-03-30 um 11 59 24
master:
bildschirmfoto 2016-03-30 um 11 59 45

### DIFF
--- a/styles/hydrogen.less
+++ b/styles/hydrogen.less
@@ -30,12 +30,11 @@
         }
 
         .rich-close-button {
+            z-index: 0;
             &::before {
                 color: @button-text-color;
                 background-color: @bubble-background;
                 display: block;
-                top: -12px;
-                right: -12px;
                 cursor: pointer;
                 border-radius: 50%;
                 width: 1.5em;
@@ -43,7 +42,7 @@
                 line-height: 1.5em;
                 text-align: center;
                 box-shadow: 0 0 2px 0px rgba(0,0,0,0.5);
-                margin: 0;
+                margin: -50% 0 0 -50%;
             }
         }
     }
--- a/styles/hydrogen.less
+++ b/styles/hydrogen.less
@@ -23,6 +23,7 @@
 
         .bubble-output-container{
             max-height: 500px;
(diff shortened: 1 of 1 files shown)


styles/hydrogen.less
```diff
@@ -30,12 +30,11 @@
         }
 
         .rich-close-button {
+            z-index: 0;
```
n-riesco: Is this what makes it stay above the plot?

styles/hydrogen.less
```diff
@@ -43,7 +42,7 @@
                 line-height: 1.5em;
                 text-align: center;
                 box-shadow: 0 0 2px 0px rgba(0,0,0,0.5);
-                margin: 0;
+                margin: -50% 0 0 -50%;
```
n-riesco: How about `0 0 0 -50%` so that it doesn't hide the line above?

styles/hydrogen.less
lgeiger: Yes

styles/hydrogen.less
lgeiger: Hmm but then it will look odd. I've got another idea, I'll try that later if it works.

lgeiger: bildschirmfoto 2016-03-30 um 23 49 42
Do you like this one better?

n-riesco:
It looks better, but it may not work for small images?

lgeiger:
Yes you're right design should follow function. Two more suggestions:
1. similar to old close button:
bildschirmfoto 2016-03-31 um 01 16 37
2. similar to action-panel:
bildschirmfoto 2016-03-31 um 01 16 15

n-riesco:
Both look good.

lgeiger:
OK lets go with number 1

n-riesco:
Is this PR ready to merge?

lgeiger:
@n-riesco yes

n-riesco:
Sorry. I should've merged this PR after merging #232 .

PR #242, opened 2016-03-30
Fix: Downscale images to fit.

### LINKED ISSUE #240: Scalling on High Resolution Display
First of all, thanks for the absolutely great package. Very useful.

I recently got a 4K screen and notices a small glitch. While the plots in Toggle Watch are nicely scaled, when I plot in the main editor body, the plot is to big to fit in the plotting window and I have to scroll horizontally and vertically. 

Any chance for an option to scale? It would be great if a full plot is shown in the box so no scrolling is necessary.

Also it would be useful if the plot is shown above the line from the 'wrap-guide' package. When using Atom on half of the screen the line always go through the plot.

Thanks for your work!

scale

lgeiger:
Fixes #240

### DIFF
--- a/styles/hydrogen.less
+++ b/styles/hydrogen.less
@@ -102,7 +102,7 @@
         display: inline-block;
         max-height: 250px;
         max-width: 95%;
-        overflow: scroll;
+        overflow: auto;
     }
 
     .bubble-result-container > * {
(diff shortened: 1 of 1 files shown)


n-riesco:
Doesn't break anything here! :smile_cat:

PR #247, opened 2016-03-31
Support markdown output

lgeiger:
transformime :tada:

### DIFF
--- a/lib/kernel.coffee
+++ b/lib/kernel.coffee
@@ -363,6 +363,9 @@ class Kernel
             else if not (imageMimes.length is 0)
                 mime = imageMimes[0]
 
+            else if data.hasOwnProperty 'text/markdown'
+                mime = 'text/markdown'
+
             else if data.hasOwnProperty 'text/plain'
                 mime = 'text/plain'
 
--- a/lib/result-view.coffee
+++ b/lib/result-view.coffee
@@ -220,6 +220,7 @@ class ResultView
 
 transformime = require 'transformime'
 transformimeJupyter = require 'transformime-jupyter-transformers'
+marked = require 'marked'
 
 SVGTransform = (mimetype, value, document) ->
     container = document.createElement 'div'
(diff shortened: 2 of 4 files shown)


n-riesco:
Neat!

I mentioned this [here](https://github.com/nteract/hydrogen/wiki/Development-Notes): I'm planning to get rid of the intermediate `result` object, so that most `_parse*` methods will become unnecessary and we will be able to implement this kind of hacks without touching `kernel.coffee`.

lgeiger:
> I mentioned this here: I'm planning to get rid of the intermediate result object, so that most _parse* methods will become unnecessary and we will be able to implement this kind of hacks without touching kernel.coffee.

Great! :tada:

lgeiger:
I added a new Markdown Transformer using [`marked`](https://github.com/chjj/marked).
This is the same parser which `jupyter-notebook` is using. It supports github flavored markdown and tables.

@rgbkrk Is this something you would like to have added to`transformime`?

rgbkrk:
Ack! Sorry I closed this on accident while reading on mobile.

rgbkrk:
We have a commonmark transform that we used in other places which I tended towards because commonmark is moving forward (even as a spec) while marked is fairly stagnant.

That being said, I think we should include a markdown transform by default in transformime.

rgbkrk:
D'oh! Now I see the markdown transformer included from before yet not included.

rgbkrk:
> Is this something you would like to see in `transformime`?

No harm in making the PR. Now that jupyterlab efforts are running their own phosphor based version of transformime, I don't think you'll conflict with anyone (and they can always override it). The only other project that I'm aware of that uses transformime is sidecar. An simpler thing to do is to create yet another repo just for the marked transform, much like how the [commonmark one stands alone](https://github.com/nteract/transformime-commonmark).

lgeiger:
> An simpler thing to do is to create yet another repo just for the marked transform, much like how the commonmark one stands alone.

I'll do that! I hope I'll find time for that next week.

lgeiger:
@rgbkrk I have a `transformime-commonmark` ripoff using `marked` working locally. I can push it to nteract now.

How do you handle the npm publishing?

rgbkrk:
Do you have an npm account and everything set up already @lgeiger? Feel free to push the code out to a public repository and we can discuss the release there.

lgeiger:
@rgbkrk I've pushed it to [tranformime-marked](https://github.com/nteract/transformime-marked).

lgeiger:
Now [`tranformime-marked`](https://github.com/nteract/transformime-marked) is used instead of `transformime-commonmark`.

PR #255, opened 2016-04-08
Cleanups

lgeiger:
- Remove unused transformers
- Remove unused imports
- Make the linter happy
- Fix small bug caused by missing `lodash`

### DIFF
--- a/lib/result-view.coffee
+++ b/lib/result-view.coffee
@@ -1,5 +1,4 @@
 {CompositeDisposable} = require 'atom'
-_ = require 'lodash'
 
 module.exports =
 class ResultView
@@ -227,12 +226,9 @@ transformimeJupyter.consoleTextTransform.mimetype = [
 ]
 
 transform = transformime.createTransform [
-    transformimeJupyter.PDFTransform,
     transformime.ImageTransformer,
     transformimeJupyter.SVGTransform,
     transformimeJupyter.consoleTextTransform,
-    transformimeJupyter.LaTeXTransform,
     MarkdownTransform,
     transformime.HTMLTransformer,
-    transformimeJupyter.ScriptTransform
 ]
--- a/lib/config.coffee
+++ b/lib/config.coffee
@@ -1,3 +1,5 @@
+_ = require 'lodash'
+
 module.exports = Config =
     getJson: (key, _default = {}) ->
         return _default unless value = atom.config.get "Hydrogen.#{key}"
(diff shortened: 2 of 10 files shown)


lib/result-view.coffee
```diff
@@ -227,12 +226,9 @@ transformimeJupyter.consoleTextTransform.mimetype = [
...
 
 transform = transformime.createTransform [
-    transformimeJupyter.PDFTransform,
     transformime.ImageTransformer,
     transformimeJupyter.SVGTransform,
     transformimeJupyter.consoleTextTransform,
-    transformimeJupyter.LaTeXTransform,
     MarkdownTransform,
     transformime.HTMLTransformer,
-    transformimeJupyter.ScriptTransform
```
rgbkrk: Do these get removed because of how we end up using transformime within Hydrogen, rather than relying on a transform based on the full bundle?

lgeiger: Yes. Currently if they are not included [here](https://github.com/nteract/hydrogen/blob/master/lib/kernel.coffee#L352) hydrogen will note use them.

n-riesco: It's in my TODO list to remove the intermediate objects that Hydrogen creates for each kernel response. Once I do that `_parseDataMime` will go and we will rely fully on transformime.

Will we remember then to restore this transformer?

n-riesco:
Tested on my machine :ok_hand: 

@lgeiger Thanks a lot for all the latest PRs! Hydrogen looks so much neater now (transformime-marked, handling the ANSI codes, the CSS of the bubble results, ...).

lib/result-view.coffee
lgeiger: I know. This is great!

I tried to include the transformers instead of removing, but they didn't work out of the box. So either way we will have to do some work on them in the future.
For example `PDFTransform` displays a link to the pdf, but atom doesn't allow it to be clicked.

n-riesco: OK, I now see the reason to remove them.

rgbkrk:
I'm unable to ship a release for the next day though I think we should soon.

lgeiger:
Done 🎉 
```
apm publish patch
Preparing and tagging a new version ✓
Pushing v0.8.4 tag ✓
Publishing Hydrogen@v0.8.4 ✓
```

rgbkrk:
Excellent!

PR #259, opened 2016-04-16
Support pdf output

lgeiger:
Wrap pdf data URLs in `webview` tags in order to be able to download the pdf output.

This is not a great solution but I think it's better than nothing.


rgbkrk:
Ah that makes sense. The Jupyter notebook does roughly the same, with a link.

PR #260, opened 2016-04-17
Clean up kernel and language usage

### LINKED ISSUE #258: Strange message when restarting kernel
```
Uncaught Error: No grammar found for language scala211
/Users/me/.atom/packages/hydrogen/lib/kernel.coffee:483
Show Stack Trace
The error was thrown from the hydrogen package. hydrogen is out of date: undefined installed; 0.8.4 latest. Upgrading to the latest version may fix this issue.
Check for package updates
```

I'm running 0.8.4. The `scala211` kernel is there (I had been using it), but I have a language mapping in my settings: `{"scala" : "scala211"}`.

n-riesco:
This PR is another go at partially cleaning up `KernelManager`; in this case, the handling of language names:
- I've replaced the setting `grammarToKernel` with `kernelMappings`.
- I've fixed the handling of the setting `languageMappings`.
- I've changed the behaviour of the command `hydrogen:update-kernels`. Now, this command only updates, i.e. it doesn't delete the old entries.

I've only tested this PR in Ubuntu 16.04. This means `ipython kernelspec` not working.

I'm yet to test whether this PR works on systems with `ipython kernelspec` or `jupyter kernelspec`. Please, do not merge before testing that.

I reckon that since this PR changes the settings offered by Hydrogen, the version should be bumped up to 0.9.0.

Fixes #258


rgbkrk:
version bump sounds good

rgbkrk:
That's quite a bit of cleanup! I'll have to check this out later. Happy to see patches flowing at least.

n-riesco:
@rgbkrk I'm sorry about that.

`KernelManager` is used by almost every other class in Hydrogen, and each class seemed to have their own way to call `KernelManager` (some used kernel specs, some used some kind of language name, some used the editor grammar directly). So in the end, what I thought it would only touch the handling of language names ended up having to with deal kernel specs and grammar too.

When you review the code, bear in my mind, that what Hydrogen used to call `kernelInfo` is actually a kernel spec and I've renamed it to `kernelSpec`.

`KernelManager` would benefit from a further cleanup:
- now that Hydrogen only uses one kind of language name (that derived from Atom's grammar), it'd be possible to rename everwhere `grammarLanguage` to simply `language` without ambiguity.
- to open the possibility of having multiple kernels running for same language, running kernels shouldn't be indexed by the their `grammarLanguage`. We need a way to link editors and kernels (this change olone would remove a lot of code everywhere).
- Although in this PR I removed code that isn't in use, I wasn't exhaustive and further work is needed.

n-riesco:
This PR isn't ready yet. I'm testing on Jupyter installation and I've find a bug. I will update the PR later today.

n-riesco:
@rgbkrk I've fixed the PR after testing on a VM with Jupyter. The PR is now ready for review.

lgeiger:
Thanks a lot for cleaning this up!

The `Select Watch Kernel` command isn't working for me. The kernel names don't appear in the list.

n-riesco:
@lgeiger Oops! Forgot to check that one!

n-riesco:
@lgeiger It should be fixed now!

PS: I really miss having the safety net of a good test suite.

lgeiger:
It works 👍 

@n-riesco I completely agree with you, a test suite would be great!

rgbkrk:
I'd like to assume only good here since @lgeiger says it's working well. How about we merge and keep going?

lgeiger:
I didn't test it with multiple kernels for the same language but that should be fine too.

So 👍 for merging.

rgbkrk:
It's ok for you to merge others' PRs as well @lgeiger, so feel free when you're happy to.

Want to cut the release?

sbromberger:
Happy to test against javascript, python, scala, and julia if you tag a release.

lgeiger:
I can cut a release by tomorrow morning. (I'm on the go)

n-riesco:
@sbromberger , @updiversity The new version will offer two settings:
- `languageMappings` now does what it was meant to do. It maps kernels' languages to Atom's languages. For example in @sbromberger 's case, for the Scala kernel, `{"scala211" : "scala"}` should work.
- `kernelMappings`, on the other hand, maps Atom's languages to kernels. For example, in @updiversity 's case, to tell Hydrogen to use a specific kernel for python files, this should work `{"python": "Python Venv 2"}`.

lgeiger:
https://github.com/nteract/hydrogen/releases/tag/v0.9.0 🎉

sbromberger:
Python, Julia and Javascript -> no problems. Will try scala when I get into work. Thank you! The new settings are much clearer.

lgeiger:
Great! Thanks for testing 👍

sbromberger:
Confirmed scala is working. Note that the instructions are accurate but reflect a change to the configuration that may trip up some folks. Previously, the mapping was {"scala" : "scala211"}. Now it's reversed.

n-riesco:
For the record, the reason why `languageMappings` uses the kernel language as index is to allow different kernel languages to map to the same Atom language, e.g. `{"scala211" : "scala", "scala210" : "scala"}`

sbromberger:
Understood, and please don't interpret my comment as a complaint - I noted it in case someone who has recently upgraded is puzzled as to why his/her mapping broke.

This is really wonderful work and has made my coding environment immensely better. Thank you :)

PR #262, opened 2016-04-17
New Demo GIF

lgeiger:


n-riesco:
I couldn't help trying to read the code while watching the animation and that makes the animation feel a wee too fast. How about adding a comment before each section of code?

And how about also showing:
- a "Hello, World!" one-line bubble
- and an example with ANSI colours
?

lgeiger:
Good points!
I made a new one.

rgbkrk:
Very awesome.

PR #273, opened 2016-05-21
Upgrade to lodash v4

lgeiger:


rgbkrk:
Would you want to enable Travis @lgeiger?

lgeiger:
@rgbkrk Actually I've set up Travis for hydrogen a while ago but I didn't had time to write specs.

I can take a look at it again and write some tests for a few functionsso we can iterate from that.

slavaGanzin:
All of that can be done with coffeescript and lodash it just a dumb wrapper around underscore(which do not change function names "because fuck you").

I'm not arguing to move to another library, just annoyed with lodash.
p.s. @lgeiger Try [ramda|(http://ramdajs.com/0.21.0/docs/)] one day

PR #278, opened 2016-05-25
Add tests for kernelspec handling

lgeiger:


rgbkrk:
Rekicking your build just in case.

lgeiger:
It looks it timed out: https://github.com/nteract/hydrogen/blob/master/spec/hydrogen-spec.coffee#L69

Issue #283, opened 2016-06-06
Your Hydrogen config is broken: kernelspec


Kiminaka:
Hi, I'm trying to use the extension but I have trouble setting up..
I got this error:
<img width="462" alt="screen shot 2016-06-05 at 10 27 30 pm" src="https://cloud.githubusercontent.com/assets/2072406/15812685/bf57f53c-2b6c-11e6-90a7-9d65f212484b.png">

My setting:
<img width="705" alt="screen shot 2016-06-05 at 10 34 39 pm" src="https://cloud.githubusercontent.com/assets/2072406/15812780/be74c0ae-2b6d-11e6-8829-f494f2de50c1.png">

Could someone tell me how to fix it?

n-riesco:
To fix this problem, please delete `{jupyter kernelspec list --json}` in your Kernel Specs setting.

---

I'm leaving this issue open as a reminder to improve the error message.

Kiminaka:
Thanks n-risesco! 
I deleted `jupyter kernelspec list --json` but I still get this error.. 
<img width="468" alt="screen shot 2016-06-06 at 8 35 51 am" src="https://cloud.githubusercontent.com/assets/2072406/15827858/ca0fd79c-2bc1-11e6-950d-c561e6676d16.png">

n-riesco:
What OS are you using? If you are using the IPython distributed with Ubuntu 16.04, you'd need to edit the Kernel Spec setting as:

```json
{
  "kernelspecs": {
    "python2": {
      "spec": {
        "language": "python",
        "display_name": "Python 2",
        "env": {},
        "argv": [
          "/usr/bin/python",
          "-m",
          "ipykernel",
          "-f",
          "{connection_file}"
        ]
      }
    }
  }
}
```

Kiminaka:
I'm using OSX El Captain

lgeiger:
Have you saved the file you want to execute with a correct file extension (`.py` for python etc.)?

Otherwise Atom can't recognize the correct grammar and Hydrogen can't start the kernel.

Kiminaka:
Thanks lgeiger! I see. I was testing it with an unsaved file. But if I try it with a .py file, I still get the same error.. 
`No kernel for language python found`

lgeiger:
Can you run `jupyter kernelspec list --json` in the terminal and post the output.

Kiminaka:
Here is the output!
```
{
  "kernelspecs": {
    "python2": {
[...]

PR #284, opened 2016-06-07
Save kernelMapping on change

slavaGanzin:
fixes #271

Issue #272, opened 2016-05-21
Handling of kernelspecs


n-riesco:
I'm moving the discussion about kernelspecs from #271 to this issue.

@slavaGanzin @rgbkrk Please, could you clarify your position? Are you supporting making the installation of Jupyter a requirement to use Hydrogen?

The latest Ubuntu 16.04 LTS comes with `IPython v2.4.1`. This version of IPython doesn't support the command `ipython kernelspec`. This is very unfortunate and it's raising the hurdle users have to hop in order to use Hydrogen.

@rgbkrk Would any of the `nteract` libraries be ready to help us here? If not, I would consider contributing to them so that we can solve this unfortunate situation sooner than later.

slavaGanzin:
I don't care about Ubuntu problems, but as I said before:
 if ```jupyter/ipython kernelspec``` fails hydrogen should use kernelspec mappings that can be set in hydrogen settings. That means jupyter wouldn't be a requirement for hydrogen.

if we can get kernelspecs from ipython/jupyter, than that mappings should be merged internally (without showing them in settings) with kernelspecs from settings. And so if kernel was uninstalled from system it wouldn't be listed in kernel list (as now).
In this way you can manage only remote kernels (or specific ones) in settings.

rgbkrk:
[`kernelspecs`](https://github.com/nteract/kernelspecs) is the library for querying kernel specs, though it only relies on `jupyter-paths` at the moment. We could introduce a way to check for `ipython-paths` as well as old IPython. Based on the fact that we "require" a kernel to be installed, even with `jupyter`, we likely need a hokey default kernel to be provided back from kernelspecs.

slavaGanzin:
> I don't care about Ubuntu problems
@n-riesco Life is ironic, even sarcastic - now I work on ubuntu 16.04 LTS (laughing allowed)

I read docs for ijavascript, jp-babel and installed env as described:
```sudo apt-get install nodejs-legacy npm ipython ipython-notebook libzmq3-dev```

and was discouraged with:

```
updateKernelSpecs: `jupyter kernelspec` failed Error: Command failed: /bin/sh -c jupyter kernelspec list --json --log-level=CRITICAL(…)
kernel-manager.coffee:71 updateKernelSpecs: `ipython kernelspec` failed Error: Command failed: /bin/sh -c ipython kernelspec list --json --log-level=CRITICAL(…)
```

after that I try to intall ```jupyter``` from pip and everything works like a magic. 

[...]

PR #286, opened 2016-06-07
Select kernel from available and not only running

slavaGanzin:
> Because I fucked up with all javascript grammars I have in my setup, so I just want to match kernel to grammar. Editing JSON is not user "unfrendly", it's just a mess. 

> And for example I work with jade(pug). It has inline jstransformers http://jade-lang.com/reference/filters/ and I want to switch between languages for one grammar.

Anyone could just switch to proper kernel. If he gets something like this https://github.com/nteract/hydrogen/issues/283#issuecomment-223997205

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -306,10 +306,7 @@ module.exports = Hydrogen =
 
     showKernelPicker: ->
         unless @kernelPicker?
-            @kernelPicker = new KernelPicker( =>
-              KernelManager.getAllKernelSpecsFor(
-                KernelManager.getGrammarLanguageFor(@editor.getGrammar())
-            ))
+            @kernelPicker = new KernelPicker KernelManager.getAllKernelSpecs.bind(KernelManager)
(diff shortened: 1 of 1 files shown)


n-riesco:
No they wouldn't be able, because they use anaconda.

What did you think I meant in this [comment](https://github.com/nteract/hydrogen/pull/284#issuecomment-224372947)?

slavaGanzin:
@n-riesco @rgbkrk @lgeiger 

Why this hasn't been merged?

n-riesco:
@slavaGanzin Waiting to agree a solution in #272

rgbkrk:
> Why this hasn't been merged?

* @n-riesco had an unresolved question
* this branch has conflicts with `master`

lgeiger:
A nice way of handling this would be to add an item (`Show all available kernels`) to the kernel picker which expands the list and shows all available kernels.
This would mean one click more for selecting an non matching kernel, would keep the list compact if a lot of kernels are installed and removes the need for two separate commands.

Issue #287, opened 2016-06-07
Connect to local "remote" existing kernel (for Docker)


tiangolo:
It would be great to have a way to make Hydrogen connect to an existing local Kernel using an existing connection file, like in:

`jupyter console --existing /home/user/test-docker-python/tmp/kernel/kernel-000.json`

This would allow developing in a Docker container with all its dependencies but having autocompletion, code run, etc.

With that, it would be possible to start a Docker container with all of its dependencies installed, mounting a volume to a local directory and starting the IPython kernel (from within the container) with the file in that directory. And then running the code from Atom using that Kernel.

Of course, this would leave to the user starting the Kernel the creation of the connection file, etc. And also every security concerns would be up to the user. But having a local Docker environment there wouldn't be any security concerns.

slavaGanzin:
@tiangolo try 
``` jupyter kernelspec list ``` or ``` iptyhon kernelspec list```
You'll see something like this:
```
Available kernels:
  coffeescript       /home/vganzin/.local/share/jupyter/kernels/coffeescript
  javascript         /home/vganzin/.local/share/jupyter/kernels/javascript
  python2-optirun    /home/vganzin/.local/share/jupyter/kernels/python2-optirun
  python2            /usr/local/share/jupyter/kernels/python2
  python3            /usr/local/share/jupyter/kernels/python3

```

just make a symlink to one of this folders like:

```
ln -s /home/user/test-docker-python/tmp/kernel/kernel-000.json /home/user/.local/share/jupyter/kernels/kernel.json

```

And trigger 'hydrogen update kernels' it should work

**Comment by tiangolo** [comment]
Thanks @slavaGanzin for the tip.

Here's what happens.

When I run:

```
$ jupyter kernelspec list
```

I get:

```
Available kernels:
  python2    /home/user/miniconda2/lib/python2.7/site-packages/ipykernel/resources
```

(I'm using miniconda / Anaconda Python).

But that folder has no contents (just two images):

```
$ ls /home/user/miniconda2/lib/python2.7/site-packages/ipykernel/resources 
logo-32x32.png  logo-64x64.png
```

Nevertheless, it seems that [recently the location has moved](http://stackoverflow.com/a/36433438), and to get it is with:

```
$ jupyter --runtime-dir
/run/user/1000/jupyter
```

And inside it is the kernel connection file:

```
ls /run/user/1000/jupyter/       
kernel-16604.json
```

[...]

PR #295, opened 2016-06-13
Custom kernel connection file

tiangolo:
Allow using a custom kernel connection file from inside the project directory, it looks for `./hydrogen/kernel/hydrogen-kernel.json` inside the project and uses it as the kernel connection.

This would presumably solve #287, allowing the development using a kernel inside a Docker container.

The `README.md` is also updated including a thorough but simple recipe to test it with Docker.

I think @slavaGanzin and @lgeiger may want to test it out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ Hydrogen was inspired by Bret Victor's ideas about the power of instantaneous fe
 - code can be inspected to show useful information provided by the running kernel
 - one kernel per language (so you can run snippets from several files, all in the same namespace)
 - interrupt or restart the kernel if anything goes wrong
+- use a custom kernel connection (for example to run code inside Docker), read more in the "Custom kernel connection (inside Docker)" section
 
 <!-- <img src="http://i.imgur.com/KiHQFO4.png?1" width=300> -->
 
@@ -160,6 +161,94 @@ Hydrogen implements the [messaging protocol](http://ipython.org/ipython-doc/stab
 ## Jank
 
 - In order to have access to your `$PATH` to find where IPython and other binaries are, Atom has to be launched from the command line with `atom <location>`. If you launch Atom as an app, this package won't work.
+
+
+## Custom kernel connection (inside Docker)
+
+You can use a custom kernel connection file to connect to a previously created kernel.
+
+For example, you can run a kernel inside a Docker container and make Hydrogen connect to it automatically. If you are using Docker this would allow you to develop from Atom but with all the dependencies, autocompletion, environment, etc of a Docker container.
+
+Hydrogen will look for a kernel JSON connection file under `./hydrogen/kernel/hydrogen-kernel.json` inside your project. If that file exists, Hydrogen will try to connect to the kernel specified by that connection file.
+
+Here's a simple recipe for doing and testing that with Python:
+
+* In your project directory, create a `Dockerfile` with:
+
+```
+FROM python:2.7
+
+RUN pip install markdown
+
+RUN pip install ipykernel
+RUN echo "alias hydrokernel='python -m ipykernel "'--ip=$(hostname -I)'" -f /tmp/kernel/hydrogen-kernel.json'" >> /etc/bash.bashrc
+```
+
+You will test using the Python package `markdown` from inside the Docker container in your local Atom editor, with autocompletion, etc.
+
+The last two lines are the only (temporal) addition to your `Dockerfile` that will allow you to develop locally using the remote Python kernel. If you already have a Python project with a `Dockerfile` you only need to copy those 2 lines and add them to it:
+
+```
+RUN pip install ipykernel
+RUN echo "alias hydrokernel='python -m ipykernel "'--ip=$(hostname -I)'" -f /tmp/kernel/hydrogen-kernel.json'" >> /etc/bash.bashrc
+```
+
+The first of those two lines will install the Python package `ipykernel`, which is the only requisite to run the remote Python kernel.
+
+The second line creates a handy shortcut named `hydrokernel` to run a Python kernel that listens on the container's IP address and writes the connection file to `/tmp/kernel/hydrogen-kernel.json`.
+
+* Build your container with:
+
+```
+docker build -t python-docker .
+```
+
+* Run your container mounting a volume that maps `./hydrogen/kernel/` in your local project directory to `/tmp/kernel/` in your container. That's the trick that will allow Hydrogen to connect to the kernel running inside your container automatically. It's probably better to run it with the command `bash` and start the kernel manually, so that you can restart it if you need to (or if it dies).
+
+```
+docker run -it --name python-docker -v $(pwd)/hydrogen/kernel:/tmp/kernel python-docker bash
+```
+
+* Next, you just have to call the alias command we created in the `Dockerfile`, that will start the kernel with all the parameters needed:
+
+```
+hydrokernel
+```
+
+* You will see an output similar to:
+
+```
+root@24ae5d04ef3c:/# hydrokernel
+NOTE: When using the `ipython kernel` entry point, Ctrl-C will not work.
+
+To exit, you will have to explicitly quit this process, by either sending
+"quit" from a client, or using Ctrl-\ in UNIX-like environments.
+
+To read more about this, see https://github.com/ipython/ipython/issues/2049
+
+
+To connect another client to this kernel, use:
+    --existing /tmp/kernel/hydrogen-kernel.json
+```
+
+* And you will see that a file was created in `./hydrogen/kernel/hydrogen-kernel.json` inside your project directory.
+
+* Now you can create a file `test.py` with:
+
+```
+import markdown
+markdown.version
+```
+
+* Select the contents and run them with Hydrogen ("`cmd-shift-P`" and "`Hydrogen: run`").
+
+* You will see the inline execution and output that just ran from your kernel, even if you don't have the Python package `mardown` installed locally, because it's running inside your container.
+
+```
+import markdown [✓]
+markdown.version ['2.6.6']
+```
+
 
 
 ## Why "Hydrogen"?
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ You can use a custom kernel connection file to connect to a previously created k
 
 For example, you can run a kernel inside a Docker container and make Hydrogen connect to it automatically. If you are using Docker this would allow you to develop from Atom but with all the dependencies, autocompletion, environment, etc of a Docker container.
 
-Hydrogen will look for a kernel JSON connection file under `./hydrogen/kernel/hydrogen-kernel.json` inside your project. If that file exists, Hydrogen will try to connect to the kernel specified by that connection file.
+Hydrogen will look for a kernel JSON connection file under `./hydrogen/connection.json` inside your project. If that file exists, Hydrogen will try to connect to the kernel specified by that connection file.
 
 Here's a simple recipe for doing and testing that with Python:
 
@@ -181,7 +181,7 @@ FROM python:2.7
 RUN pip install markdown
 
 RUN pip install ipykernel
-RUN echo "alias hydrokernel='python -m ipykernel "'--ip=$(hostname -I)'" -f /tmp/kernel/hydrogen-kernel.json'" >> /etc/bash.bashrc
+RUN echo "alias hydrokernel='python -m ipykernel "'--ip=$(hostname -I)'" -f /tmp/hydrogen/connection.json'" >> /etc/bash.bashrc
 ```
 
 You will test using the Python package `markdown` from inside the Docker container in your local Atom editor, with autocompletion, etc.
@@ -190,12 +190,12 @@ The last two lines are the only (temporal) addition to your `Dockerfile` that wi
 
 ```
 RUN pip install ipykernel
-RUN echo "alias hydrokernel='python -m ipykernel "'--ip=$(hostname -I)'" -f /tmp/kernel/hydrogen-kernel.json'" >> /etc/bash.bashrc
+RUN echo "alias hydrokernel='python -m ipykernel "'--ip=$(hostname -I)'" -f /tmp/hydrogen/connection.json'" >> /etc/bash.bashrc
 ```
 
 The first of those two lines will install the Python package `ipykernel`, which is the only requisite to run the remote Python kernel.
 
-The second line creates a handy shortcut named `hydrokernel` to run a Python kernel that listens on the container's IP address and writes the connection file to `/tmp/kernel/hydrogen-kernel.json`.
+The second line creates a handy shortcut named `hydrokernel` to run a Python kernel that listens on the container's IP address and writes the connection file to `/tmp/hydrogen/connection.json`.
 
 * Build your container with:
 
@@ -203,10 +203,10 @@ The second line creates a handy shortcut named `hydrokernel` to run a Python ker
 docker build -t python-docker .
 ```
 
-* Run your container mounting a volume that maps `./hydrogen/kernel/` in your local project directory to `/tmp/kernel/` in your container. That's the trick that will allow Hydrogen to connect to the kernel running inside your container automatically. It's probably better to run it with the command `bash` and start the kernel manually, so that you can restart it if you need to (or if it dies).
+* Run your container mounting a volume that maps `./hydrogen/` in your local project directory to `/tmp/hydrogen/` in your container. That's the trick that will allow Hydrogen to connect to the kernel running inside your container automatically. It's probably better to run it with the command `bash` and start the kernel manually, so that you can restart it if you need to (or if it dies).
 
 ```
-docker run -it --name python-docker -v $(pwd)/hydrogen/kernel:/tmp/kernel python-docker bash
+docker run -it --name python-docker -v $(pwd)/hydrogen:/tmp/hydrogen python-docker bash
 ```
 
 * Next, you just have to call the alias command we created in the `Dockerfile`, that will start the kernel with all the parameters needed:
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -162,6 +163,94 @@ Hydrogen implements the [messaging protocol](http://ipython.org/ipython-doc/stab
...
+RUN echo "alias hydrokernel='python -m ipykernel "'--ip=$(hostname -I)'" -f /tmp/kernel/hydrogen-kernel.json'" >> /etc/bash.bashrc
+```
+
+You will test using the Python package `markdown` from inside the Docker container in your local Atom editor, with autocompletion, etc.
+
+The last two lines are the only (temporal) addition to your `Dockerfile` that will allow you to develop locally using the remote Python kernel. If you already have a Python project with a `Dockerfile` you only need to copy those 2 lines and add them to it:
+
+```
+RUN pip install ipykernel
+RUN echo "alias hydrokernel='python -m ipykernel "'--ip=$(hostname -I)'" -f /tmp/kernel/hydrogen-kernel.json'" >> /etc/bash.bashrc
```
n-riesco: Clever use of single and double quotes!

Wouldn't it make more sense to run the command `python -m ipykernel --ip=$(hostname -I) -f /tmp/kernel/hydrogen-kernel.json` from .bashrc rather than define an alias.

lib/config-manager.coffee
```diff
@@ -9,18 +9,31 @@ module.exports = ConfigManager =
...
-                throw e
-        filename = 'kernel-' + uuid.v4() + '.json'
-        portfinder.findMany 5, (ports) =>
-            config = @buildConfiguration ports
-            configString = JSON.stringify config
-            filepath = path.join(@fileStoragePath, filename)
-            fs.writeFileSync filepath, configString
-            onCompleted filepath, config
+        customKernelConnectionPath = path.join atom.project.rootDirectories[0].path, 'hydrogen', 'kernel', 'hydrogen-kernel.json'
+        fs.access customKernelConnectionPath, (err) =>
```
n-riesco: Why not `fs.readFile customKernelConnectionPath, 'utf8', (err, data) ->`?

lib/kernel-manager.coffee
```diff
@@ -149,8 +149,8 @@ module.exports = KernelManager =
 
         kernelSpec.grammarLanguage = grammarLanguage
 
-        ConfigManager.writeConfigFile (filepath, config) =>
-            kernel = new Kernel kernelSpec, grammar, config, filepath
+        ConfigManager.writeConfigFile (filepath, config, onlyConnect = false) =>
```
n-riesco: I think the functionality checking for `customKernelConnectionPath` belongs to `KernelManager`.

`ConfigManager.writeConfigFile` should only be called when a config file is written.

n-riesco:
I've tested this PR and works for me.

I like the idea of introducing a hydrogen folder per project.

I'm not sure about `./hydrogen/kernel/hydrogen-kernel.json`:
- unnecessary stutter ('./**hydrogen**/kernel/**hydrogen**-kernel.json' and './hydrogen/**kernel**/hydrogen-**kernel**.json')
- `hydrogen-kernel.json` is a bit of a misnomer (which jupyter also uses)

If we were to use `./hydrogen/connection.json` in this PR, we could have another PR with `./hydrogen/kernel.json` for custom kernels.

The main changes I'd like to see in this PR are:
- [ ] Move the functionality in this PR from `ConfigManager` to `KernelManager`
- [ ] Change the path from `./hydrogen/kernel/hydrogen-kernel.json` to `./hydrogen/connection.json`

lib/config-manager.coffee
```diff
@@ -9,18 +9,31 @@ module.exports = ConfigManager =
...
+                    fs.writeFileSync filepath, configString
+                    console.log "Creating kernel connection: ", filepath
+                    onCompleted filepath, config
+            else
+                fs.readFile customKernelConnectionPath, 'utf8', (err, data) ->
+                    unless err?
+                        config = JSON.parse data
+                        console.log "Using custom kernel connection: ", customKernelConnectionPath
+                        atom.notifications.addInfo 'Custom kernel connection:',
+                            detail: "Make sure to manually start the custom kernel that serves the connection in #{customKernelConnectionPath}", dismissable: true
```
n-riesco: I would remove this notification.
After a few times, it becomes tiring to have to dismiss this notification.

slavaGanzin: remove ```dissmissable: true``` and it will dissmiss automatically

tiangolo:
Thanks for the feedback! I'll work on it.

lib/config-manager.coffee
tiangolo: Would you like me to completely remove the notification or make it auto-dismiss (removing the `dismissabe: true`)?

n-riesco: On 14/06/16 14:10, Sebastián Ramírez wrote:
> Would you like me to completely remove the notification or make it auto-dismiss (removing the |dismissabe: true|)?

I don't think the notification is necessary, because the connection file is created when the custom kernel is launched.

n-riesco:
@tiangolo Before you work on it, let's hear what @lgeiger @rgbkrk @willwhitney think about introducing a hydrogen folder per project.

willwhitney:
Seems reasonable to me as long as it's opt-in (e.g., running some code with Hydrogen should not create this folder). But giving people a simple per-project way to specify kernels (and maybe other things in the future) seems like a win.

willwhitney:
Good work @tiangolo! In particular thanks for taking the time to update the readme.

And thanks @n-riesco for the quality reviewing as always!

lgeiger:
👍 👍 👍  from me! Thanks for your great work!

I like the idea of using `./hydrogen/connection.json` for this.

rgbkrk:
Since it's opt-in, I'm fine with that. It does seem odd to me that we're not using the jupyter configuration or runtime folders in general though.

n-riesco:
@rgbkrk Is there a Jupyter runtime folder for connection files?

rgbkrk:
Yup, it's at `~/Library/Jupyter/runtime/` on OS X. If using `jupyter-paths`, you can get it via `require('jupyter-paths').runtimeDir()`.

tiangolo:
Great! Thanks everyone for your feedback and :+1:s.

@willwhitney yes, it's opt in and doesn't create directories. Actually, Hydrogen wouldn't even create a directory in any case, the user would have to create it and put his connection file there (for example with Docker) before launching Hydrogen. That's how the user "opts in", creating a connection file under `./hydrogen/connection.json`. 

I'm glad you liked the README update! Documentation is important to me too :)

@lgeiger :D :+1: 

Yes @n-riesco, there's a directory for connection files, you can get it with: `jupyter --runtime-dir`. In my Ubuntu it is in: `/run/user/1000/jupyter`. That's where Jupyter puts the connection files of the kernels it creates directly.

@rgbkrk I agree that it's a bit strange to do not use the normal directories, but at least in the case of using Docker containers, adding the connection files in the standard location could end up "polluting" the current Jupyter space, as there might be many containers that could leave stray connection files (if the container dies, gets killed or in any case that stops before removing the connection file). I think that keeping the custom kernel inside the specific project will help isolate the environments.

And if the container stopped and left the connection file, just re-starting it and starting the kernel again will override the old connection file, keeping everything more "clean" (at least from my point of view).

tiangolo:
I think now I have all the :+1: required... 

So I'll work on it and update it with the changes asked by @n-riesco .

lib/config-manager.coffee
tiangolo: I removed it in the latest version, nevertheless, I forgot to mention why I added it.

If a user starts with a custom kernel, with the custom connection, and then triggers "Hydrogen: Restart Kernel", the shutdown signal will be sent to the external (Docker) kernel, which will stop the kernel and remove the file. And then, Hydrogen will "restart" the kernel, but as there's not a custom kernel connection anymore it will start a new, local, fresh kernel.

Using my example from the README, this will make the package `markdown` import while using the kernel container, but throw an error after restarting the kernel (using the local kernel).

I just wanted to make sure to put it here, in case you want to get the notification (auto-dismissed) back or do something else for that "corner case".

tiangolo:
@n-riesco I updated the code with the requested changes (including the README). Can you check if everything is OK?

n-riesco:
I'll check tomorrow. It's getting late here.

lib/kernel-manager.coffee
```diff
@@ -162,6 +164,23 @@ module.exports = KernelManager =
...
+            data = fs.readFileSync customKernelConnectionPath, 'utf8'
+            config = JSON.parse data
+            console.log "Using custom kernel connection: ", customKernelConnectionPath
+            kernel = new Kernel kernelSpec, grammar, config, customKernelConnectionPath, true
+            finishKernelStartup kernel
+        catch e
+            if e.code != 'ENOENT'
+                trow e
+            console.log(e)
+            ConfigManager.writeConfigFile (filepath, config, onlyConnect = false) =>
```
n-riesco: This bit is confusing because `ConfigManager.writeConfigFile` expects a callback with two arguments only (`filepath` and `config`). I think as below would be clearer:

```coffee
            ConfigManager.writeConfigFile (filepath, config) =>
                kernel = new Kernel kernelSpec, grammar, config, filepath, onlyConnect=false
                finishKernelStartup kernel
```

lib/kernel.coffee
```diff
@@ -27,61 +27,65 @@ class Kernel
...
-
-            regexp = getKernelNotificationsRegExp()
-            if regexp?.test data
-                kernelName = @kernelSpec.display_name ? @language
-                atom.notifications.addError kernelName + ' kernel:',
-                    detail: data, dismissable: true
+        if @onlyConnect
+          atom.notifications.addInfo 'Using custom kernel connection:',
+              detail: @configPath
+        unless @onlyConnect
```
n-riesco: `else` ?

lib/kernel-manager.coffee
```diff
@@ -149,9 +151,9 @@ module.exports = KernelManager =
 
         kernelSpec.grammarLanguage = grammarLanguage
 
-        ConfigManager.writeConfigFile (filepath, config, onlyConnect = false) =>
-            kernel = new Kernel kernelSpec, grammar, config, filepath, onlyConnect
+        customKernelConnectionPath = path.join atom.project.rootDirectories[0].path, 'hydrogen', 'connection.json'
 
+        finishKernelStartup = (kernel) =>
```
n-riesco: I don't think this is necessary.

CoffeeScript is smart enough to handle the scope of `kernel` (even if assigned within a try-catch block).

The CoffeeScript code below:
```coffee
try
    foo = false
catch err
    foo = true

console.log foo
```
compiles to
```javascript
var err, error, foo;

try {
  foo = false;
} catch (error) {
  err = error;
  foo = true;
}

console.log(foo);
```

n-riesco:
@tiangolo The PR is almost there! I'm looking forward to merging. It really is a very convenient way to use connection files directly.

lib/kernel-manager.coffee
tiangolo: You're right. Done.

lib/kernel.coffee
tiangolo: Right. Good catch.

lib/kernel-manager.coffee
tiangolo: The thing is that this code:

```CoffeeScript
            @_runningKernels[grammarLanguage] = kernel

            startupCode = Config.getJson('startupCode')[kernelSpec.display_name]
            if startupCode?
                console.log 'executing startup code'
                startupCode = startupCode + ' \n'
                kernel.execute startupCode

            onStarted?(kernel)
```

used to run inside this callback:

```CoffeeScript
        ConfigManager.writeConfigFile (filepath, config) =>
            kernel = new Kernel kernelSpec, grammar, config, filepath

            @_runningKernels[grammarLanguage] = kernel

            startupCode = Config.getJson('startupCode')[kernelSpec.display_name]
            if startupCode?
                console.log 'executing startup code'
                startupCode = startupCode + ' \n'
                kernel.execute startupCode

            onStarted?(kernel)
```

but now it should run either with a normal kernel (inside the callback) or with a custom kernel (called directly).

I tried first running it outside of both scopes after the `try-catch` with something like:

```CoffeeScript
            kernel = ConfigManager.writeConfigFile (filepath, config) =>
                kernel = new Kernel kernelSpec, grammar, config, filepath, onlyConnect=false
            

            @_runningKernels[grammarLanguage] = kernel
            startupCode = Config.getJson('startupCode')[kernelSpec.display_name]
            ...
```

but the value returned by `ConfigManager` is not the kernel but a `Server`, taken from the console:

```javascript
Server {domain: null, _events: Object, _eventsCount: 0, _maxListeners: undefined, _connections: 0…}
```

Digging in the final value (returns) of the callbacks I went through several callbacks in `config-manager.coffee`, `find-port.coffee` and thought that "debugging" that could lead to modifying several deep files unnecessarily while I could encapsulate that code in a function and just pass it.

Of course, there is the option of duplicating the code in both blocks, inside and outside the function, but I thought that would be a complete no-no.

Maybe I'm missing something and you can guide me...

tiangolo:
@n-riesco I made the changes you asked, and asked questions about the last one, as I can't make it work and maybe I'm missing something.

lib/kernel-manager.coffee
n-riesco: @tiangolo Oh! I missed that `finishKernelStartup` was called in the callback to `ConfigManager.writeConfigFile`. Now I understand the need for `finishKernelStartup`.

n-riesco:
@tiangolo Later today, I will test the PR once more and merge it.

Thanks a lot for this PR.

tiangolo:
Thanks @n-riesco! 

I'm glad to help... and actually, I need and will use this a lot. Once merged, Atom-Hydrogen-Docker will be my default dev setup. :smile:

n-riesco:
@tiangolo It works nicely!

@rgbkrk @willwhitney @slavaGanzin @lgeiger
Shall we make a release? v0.10.0 ?

tiangolo:
Great @n-riesco! Thanks for merging :tada: 

I hope you make a release so I can update my Atom with the new official version and more importantly, share the new functionality with some people in my team.

willwhitney:
Seems like time! @n-riesco if everything works for you, feel free to cut a release.

rgbkrk:
Welcome to nteract, @tiangolo!

tiangolo:
Oh! Thanks @rgbkrk! That's cool :D :tada:

jdavidheiser:
This looks pretty great - will it work with remote kernels that are behind jupyterhub?

n-riesco:
@jdavidheiser Do you know where jupyterhub stores the connection files?

tiangolo:
@jdavidheiser I haven't used Jupyterhub, but giving a quick look at its docs, here's what it probably does:

Once a user connects to Jupyterhub, it creates a Jupyter server for that user, and that server creates kernels and the connection files for those kernels. And then it starts (web) notebooks connected to each of those kernels for that user. As I understand, it creates a separate server for each user.

---

Here's what would be needed to do for it to work (somehow, I don't know if Jupyterhub already has configurations for this):

Once the user's Jupyter server inside Jupyterhub creates a kernel, before starting a web notebook (which it probably does automatically) the user should get the connection file for that kernel and put it locally inside his project under `./hydrogen/connection.json`. As there's a server per user, I think it's probable that the kernels for that user are isolated too.

---

But there's the authentication / authorization thing, which is a big thing. Following the docs it seems Jupyterhub solves this in different ways, but the main solution is that it needs each user to have an acount (user) in that *nix server and then somehow authenticate to the *nix server (I'm not sure how each user would authenticate with his *nix credentials in the web notebook interface).

And I don't know if a direct connection to a kernel would have any authentication mechanism, I don't think so. So it would probably be needed to set it up apart from the rest.

---

What I would do is:

First, make sure it's absolutely needed to have Jupyterhub, if not, use vanilla Jupyter. Create a user account in the *nix server for each user that needs to connect (Jupyterhub already needs to do that too). Then create a Jupyter kernel with its kernel connection file in a predefined location, and listening only to local connections (localhost IP), as with:

```
python -m ipykernel --ip=127.0.0.1 -f /tmp/hydrogen/connection.json
```

Then, copy (`rsync` or `scp`) that connection file to my local project under `./hydrogen/connection.json`.

Then open the file and check the ports used, and create a SSH tunnel for each port used. Something like:

```
ssh -f user@personal-server.com -L 2000:personal-server.com:25 -N
```

Here's an [article on how this SSH tunnel works](http://www.revsys.com/writings/quicktips/ssh-tunnel.html).

For a quick and formal explanation of each parameter, [explainshell.com](http://explainshell.com/explain?cmd=ssh+-f+user%40personal-server.com+-L+2000%3Apersonal-server.com%3A25+-N).

After creating all the needed tunnels and having the connection file in place, I would try it out with Hydrogen. It should work.

If that works, then I would probably try to automate the whole thing. I don't know if it should be part or Hydrogen or more like a separate complementary script. 

If done with NodeJS I would try [tunnel-ssh](https://www.npmjs.com/package/tunnel-ssh#remap-remote-port-to-localhost) or [ssh2](https://www.npmjs.com/package/ssh2) directly and [rsync](https://www.npmjs.com/package/rsync). 

If done with Python I would try [paramiko](https://github.com/paramiko/paramiko) and or a combination of [sshtunnel](https://sshtunnel.readthedocs.io/en/latest/), [pyrsync](https://github.com/isislovecruft/pyrsync), [scp.py](https://github.com/jbardin/scp.py) and [scpclient](https://pypi.python.org/pypi/scpclient).

rgbkrk:
@minrk, @parente, and I have talked about making kernels available remotely by API (REST API first for request, websockets for the actual connection).

PR #297, opened 2016-06-15
Use kernel display name for status bar

lgeiger:
This will change the text from `python kernel: idle` to `Python 3: idle` (useful for using 2 and 3 side by side).


n-riesco:
Nice!

PR #308, opened 2016-06-25
CellManager: Simplify algorithm of getCurrentCell

n-riesco:


lgeiger:
This works only if breakpoints are added sequentially, because the `decorations` aren't ordered by line number. So we have to loop over all `decorations` to find the correct start row.

PS: Thanks a lot for improving my ugly code 😉

n-riesco:
Oh! I now understand why the initial algorithm was so complicated. I still think it can be simplified. I'll give it another go.

lgeiger:
This is a lot cleaner 👍

PR #309, opened 2016-06-25
Add command to toggle breakpoints

n-riesco:


n-riesco:
I think that now that we have 'hydrogen:toggle-breakpoint', there isn't much need for 'hydrogen:remove-latest-breakpoint'. Shall we remove it?

lgeiger:
This is  great!

I think we can remove both 'hydrogen:remove-latest-breakpoint' and 'hydrogen:add-breakpoint'.

PR #320, opened 2016-07-03
Fix #319 (Fix interrupt kernel)

### LINKED ISSUE #319: Interrupt Kernel command kills the kernel instead
`handleKernelCommand` should call `kernel.interrupt` instead of killing the kernel.

n-riesco:
* Call kernel.interrupt instead of KernelManager.destroyRunningKernel


lgeiger:
Great! 👍

PR #322, opened 2016-07-04
Fix #321 (Update statusbar)

### LINKED ISSUE #321: Update status bar after kernel restart
It would be great to update the status bar directly after the `restart-kernel` and `switch-kernel` commands to reflect the changes. 

Currently the status bar is only updated on execute.

n-riesco:
First commit fixes #321 .
Second commit is mostly a linter clean-up.
I think it's better to review each commit separately.

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -22,6 +22,7 @@ module.exports = Hydrogen =
     statusBarTile: null
 
     editor: null
+
     markerBubbleMap: {}
 
     activate: (state) ->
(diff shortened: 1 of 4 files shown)


lgeiger:
Great!

On kernel startup the statusbar shows just the `kernelname`. I think it should show `kernelname: starting` instead.

In principle we have an io message  for this, but it gets marked as invalid because it has no `parent_header.msg_id`. However with the current solution a kernel has to be running anyway in order to set the satusbar element.
I think the easiest solution for this would be to simply change [this line](https://github.com/nteract/hydrogen/blob/master/lib/status-view.coffee#L9) to `@element.innerText = @language + ": starting"`.

lib/main.coffee
```diff
@@ -296,36 +288,49 @@ module.exports = Hydrogen =
...
+        unless @statusBarElement?
+            console.error 'removeStatusBarElement: there is no status bar'
+            return
+
+        while @statusBarElement.hasChildNodes()
+            @statusBarElement.removeChild @statusBarElement.lastChild
+
+    setStatusBarElement: ->
+        unless @statusBarElement?
+            console.error 'setStatusBarElement: there is no status bar'
```
lgeiger: This line throws an error every time Hydrogen is activated. This didn't happen before.

n-riesco: I don't see this error. Could you post the traceback?

lgeiger: If I restart atom and then start Hydrogen, this is the output:
bildschirmfoto 2016-07-05 um 09 04 46

n-riesco: Oh! I see the problem. This is actually a console error. It's happening because I dropped [this check](https://github.com/nteract/hydrogen/blob/20be0e3a22d296a2f282a17fa9d1ad777d222126/lib/main.coffee#L95). I'll fix it.

lgeiger:
Great! I'll take a look at this in the evening.

lgeiger:
Very nice! I'll go ahead and merge.

By the way wich linter are you using for coffeescript?

n-riesco:
On 05/07/16 18:41, Lukas Geiger wrote:
> By the way wich linter are you using for coffeescript?

I use coffeelint. But I haven't spent a lot of time to set it up.

lgeiger:
I use coffeelint.

Should we add a coffeelint config file to this repository to make things consistent?

PR #323, opened 2016-07-05
Statusbar: Fix error when editor is undefined

lgeiger:
This happens for example if a pane is split down and closed again.

The check was removed in #322.

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -64,8 +64,9 @@ module.exports = Hydrogen =
         @editor = atom.workspace.getActiveTextEditor()
 
         @subscriptions.add atom.workspace.observeActivePaneItem (item) =>
-            @editor = item
-            @setStatusBarElement()
+            if item
+                @editor = item
+                @setStatusBarElement()
 
         KernelManager.updateKernelSpecs()
 


n-riesco:
I don't understand how `@editor` could be `undefined`, since it's set in the constructor.

I neither understand why  the check is in `setStatusBarElement`, since this function doesn't use `@editor` explicitly.

lgeiger:
> I don't understand how @editor could be undefined, since it's set in the constructor.

The problem occurs [here](https://github.com/nteract/hydrogen/blob/master/lib/main.coffee#L66).
[This check](https://github.com/nteract/hydrogen/pull/322/files#diff-2e1bb36917ff76ab7308da52bc968501L81) was used before.

To reproduce this issue:
1. start hydrogen
2. split down the pane from the context menu of the tab
3. try closing the pane and you will see the error in the dev console

> I neither understand why the check is in setStatusBarElement, since this function doesn't use @editor explicitly.

I can move the check to [getCurrentKernel](https://github.com/nteract/hydrogen/blob/master/lib/main.coffee#L125) since it  fails here to get the grammar of the undefined editor.

n-riesco:
:disappointed: Again, I can't reproduce it. Do you do anything after closing the pane?

Wouldn't it make more sense to change [`@editor = item`](https://github.com/nteract/hydrogen/blob/master/lib/main.coffee#L67) to `@editor = atom.workspace.getActiveTextEditor()`?
 So that we ensure @editor is defined at all times and prevent this bug from happening somewhere else.

lgeiger:
> Wouldn't it make more sense to change @editor = item to @editor = atom.workspace.getActiveTextEditor()?

This doesn't help prevent this issue.

This is what I'm doing:
hydrogen

n-riesco:
Oh I see! I was using the context menu to close the panel.

n-riesco:
The problem is that when a panel is closed `atom.workspace.observeActivePaneItem` passes an `undefined` item to the callback. I think we should do this to ensure @editor is always defined:

```coffeescript
        @subscriptions.add atom.workspace.observeActivePaneItem (item) =>
            console.log 'observeActivePaneItem:', item
            if item
                @editor = item
                @setStatusBarElement()
```

PR #327, opened 2016-07-07
Fix run-and-move-down on last line

### LINKED ISSUE #306: Run-and-move-down doesn't display result on last line of file
Activating run-and-move-down while on the last line in the file (no trailing newline) causes the result bubble to not displayed. In addition to not showing the output, this creates the false impression that nothing was executed.

I suspect the cause is that the move-down operation has to create a new line where one previously did not exist, which triggers the output bubble to be hidden.

lgeiger:
Fixes #306
@nikitakit @n-riesco Please test if this PR is working for you.

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -266,8 +266,8 @@ module.exports = Hydrogen =
 
         [code, row] = codeBlock
         if code? and row?
-            @createResultBubble code, row
             @moveDown row
+            @createResultBubble code, row
 
     runCell: () ->
         [startRow, endRow] = CellManager.getCurrentCell()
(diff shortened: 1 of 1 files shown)


nikitakit:
Oh, this does work.

Didn't realize that just keeping the old row reference wouldn't cause any issues for the output bubble display.

nikitakit:
Oh and before the issue gets closed, I think "run cell and move down" needs the same treatment.

lgeiger:
Oh thanks I forgot.

n-riesco:
I don't remember there were any reasons to do the `moveDown` after the `createResultBubble`.
I've tested the PR and it also works for me.

PR #328, opened 2016-07-07
Code cells

### LINKED ISSUE #317: [DISCUSS] Save breakpoints
It would be awesome to save the breakpoints even if atom is restarted.

I see two ways of doing this:
1. Save the position of the breakpoints to a separate file in the `./hydrogen/` folder. A downside would be, that it will break if the file is changed outside of atom.
2. Use text to identify breakpoints like `# %%` and parse the file before execution or add markers to those positions. We could make this language independent by using something like [this](https://github.com/atom/atom/blob/master/src/language-mode.coffee#L351) to get the right comment symbol.

I tend towards the second option. What about you?

lgeiger:
Closes #317

Define code cells with
```
# %%
# <codecell>
# In[ ]
# In[9999]
```
or with whatever comment symbol your language is using.

### DIFF
--- a/grammars/breakpoints.cson
+++ b/grammars/breakpoints.cson
@@ -1,6 +1,6 @@
 'scopeName': 'hydrogen.breakpoints'
 'name': 'Hydrogen breakpoints'
-'injectionSelector': 'comment'
+'injectionSelector': 'comment.line'
 'patterns': [
   {
     'match': '%%| %%| <codecell>| In\\[[0-9 ]+\\]:'
--- a/grammars/breakpoints.cson
+++ b/grammars/breakpoints.cson
@@ -1,0 +1,9 @@
+'scopeName': 'hydrogen.breakpoints'
+'name': 'Hydrogen breakpoints'
+'injectionSelector': 'comment'
+'patterns': [
+  {
+    'match': ' %%| <codecell>| In\\[[0-9 ]+\\]:'
+    'name': 'entity.name.section.breakpoint'
+  }
+]
--- a/grammars/breakpoints.cson
+++ b/grammars/breakpoints.cson
@@ -3,7 +3,7 @@
 'injectionSelector': 'comment'
 'patterns': [
   {
-    'match': ' %%| <codecell>| In\\[[0-9 ]+\\]:'
+    'match': '%%| %%| <codecell>| In\\[[0-9 ]+\\]:'
     'name': 'entity.name.section.breakpoint'
   }
 ]
--- a/lib/cell-manager.coffee
+++ b/lib/cell-manager.coffee
@@ -1,57 +1,26 @@
-_ = require 'lodash'
-
 module.exports = CellManager =
-    _setBreakpointDecorationAt: (editor, row) ->
-        marker = editor.markBufferPosition
-            row: row
-            column: 0
-        ,
-            invalidate: 'never'
-
-        editor.decorateMarker marker,
-            type: 'line-number'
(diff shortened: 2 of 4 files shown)


lgeiger:
I added support for `#%%`.

lib/cell-manager.coffee
```diff
@@ -1,57 +1,26 @@
...
-                if (decorationRow < end)
-                    end = decorationRow
-            else
-                if (decorationRow >= start)
-                    start = decorationRow + 1
+        {commentStartString, commentEndString} =
+            editor.languageMode.commentStartAndEndStringsForScope(scope)
+
+        if commentStartString
+            regex = new RegExp(commentStartString + '%%|<codecell>|In\[[0-9 ]+\]:')
```
n-riesco: Should we enforce that the comment starts a line?

I would find the below example confusing:

```js
a = 1; # %% In[1]:
console.log(a);

a += 1; # %% In[2]:
console.log(a);
```

lgeiger: Yes good idea!

I tried this but somehow I couldn't get the syntax highlighting to respect this as well. Any ideas on how to achieve this?

n-riesco:
Oh! I really misunderstood this PR. I didn't realise it'd mean the previous functionality to toggle breakpoints would be removed.

The way I imagined this would be implemented was a parser that would convert the `#%%` lines to `toggleBreakpoint` commands.

I think this is a good call. Otherwise, the interactions between `#%%` lines and `toggleBreakpoint` commands would feel ackward.

lib/cell-manager.coffee
```diff
@@ -1,57 +1,26 @@
...
-            return
-
-        CellManager._setBreakpointDecorationAt editor, row
-
     getCurrentCell: ->
         editor = atom.workspace.getActiveTextEditor()
+        buffer = editor.getBuffer()
         row = editor.getLastCursor().getBufferRow()
-        decorations = editor.getLineNumberDecorations({class: 'breakpoint'})
+        scope = editor.scopeDescriptorForBufferPosition([row, 0])
```
n-riesco: I think [editor.getRootScopeDescriptor](https://atom.io/docs/api/v1.8.0/TextEditor#instance-getRootScopeDescriptor) is better here. Otherwise things like the below would be possible:

```coffeescript
`
//%%
(function() {console.log("This is Javascript");})();
//%%
`
console.log 'This is coffeescript'
```

lib/cell-manager.coffee
```diff
@@ -1,57 +1,27 @@
...
```
n-riesco: I think [editor.getRootScopeDescriptor](https://atom.io/docs/api/v1.8.0/TextEditor#instance-getRootScopeDescriptor) is better here. Otherwise things like the below would be possible:

```coffeescript
`
//%%
(function() {console.log("This is Javascript");})();
//%%
`
console.log 'This is coffeescript'
```

lib/cell-manager.coffee
n-riesco: I don't know (especially if we have to use `injectionSelector`)

grammars/breakpoints.cson
```diff
@@ -0,0 +1,9 @@
+'scopeName': 'hydrogen.breakpoints'
+'name': 'Hydrogen breakpoints'
+'injectionSelector': 'comment'
```
n-riesco: I think this should be 'comment.line'. Otherwise `/* %% */` would get highlighted in JS.

PR #329, opened 2016-07-07
Use coffeelint 

### LINKED ISSUE #325: Use coffeelint
Open issue as a [reminder](https://github.com/nteract/hydrogen/pull/322#issuecomment-230601773).

lgeiger:
Closes #325 

Are there any rules you want to add or remove?

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -253,11 +253,13 @@ module.exports = Hydrogen =
 
 
     runAllAbove: ->
-        codeBlock = @findCodeBlock(true)
-        unless codeBlock?
-            return
+        cursor = @editor.getLastCursor()
+        row = cursor.getBufferRow()
+        if row > 0
+            for i in [0 .. row - 1] when @blank(row)
+                row -= 1
+        code = @getRows(0, row)
 
-        [code, row] = codeBlock
         if code? and row?
             @createResultBubble code, row
 
@@ -372,7 +374,7 @@ module.exports = Hydrogen =
                     @watchSidebar.show()
         @watchKernelPicker.toggle()
 
-    findCodeBlock: (runAllAbove = false) ->
+    findCodeBlock: ->
         buffer = @editor.getBuffer()
         selectedText = @editor.getSelectedText()
 
@@ -389,12 +391,6 @@ module.exports = Hydrogen =
 
         row = cursor.getBufferRow()
         console.log 'findCodeBlock:', row
-
-        if runAllAbove
-            if row > 0
-                for i in [0 .. row - 1] when @blank(row)
-                    row -= 1
-            return [@getRows(0, row), row]
 
         indentLevel = cursor.getIndentLevel()
 
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -35,7 +35,7 @@ module.exports = Hydrogen =
         @statusBarElement = document.createElement('div')
         @statusBarElement.classList.add('hydrogen')
         @statusBarElement.classList.add('status-container')
-        @statusBarElement.onclick = @showKernelCommands.bind @
+        @statusBarElement.onclick = @showKernelCommands.bind this
 
         @subscriptions = new CompositeDisposable
 
@@ -233,7 +233,7 @@ module.exports = Hydrogen =
             column: 0
 
 
-    run: () ->
+    run: ->
         codeBlock = @findCodeBlock()
         unless codeBlock?
             return
@@ -243,7 +243,7 @@ module.exports = Hydrogen =
             @createResultBubble code, row
 
 
-    runAll: () ->
+    runAll: ->
         code = @editor.getText()
         row = @editor.getLastBufferRow()
         if row > 0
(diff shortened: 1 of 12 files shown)


n-riesco:
I'd like to set `line_endings`, `missing_fat_arrows`, `max_line_length`, `no_tabs`, `no_trailing_whitespace`, `no_unnecessary_fat_arrows`, and `cyclomatic_complexity`.

I would set all of the checks to error level, except maybe `missing_fat_arrows` and `cyclomatic_complexity`.

I've been using `no_this` for a while and I'm starting to get annoyed with it, because it forces me to use standalone @.

n-riesco:
I would split 2d6c229 into one commit per file, otherwise it's going to be painful to merge.

Or at least, it'd be nice to have the creation of `coffeelint.json` on it's own commit.

lgeiger:
The `missing_fat_arrows` rule isn't working for me, if I change the arrows according to `coffeelint` I get a lot of errors like this one:
```
Uncaught TypeError: _this.runAndMoveDown is not a function
```

We can use `no_this` along with [`no_stand_alone_at`](http://www.coffeelint.org/)  to prevent standalone `@`

n-riesco:
On 09/07/16 13:33, Lukas Geiger wrote:
> The |missing_fat_arrows| rule isn't working for me, if I change the arrows according to |coffeelint| I get a lot of errors like this one:
> 
> |Uncaught TypeError: _this.runAndMoveDown is not a function |
> 

I'm using:

```
  "missing_fat_arrows": {
    "level": "error",
    "is_strict": false
  },
```

and

```
$ coffeelint --version
1.15.0
```

> We can use |no_this| along with |no_stand_alone_at| <http://www.coffeelint.org/> to prevent standalone |@|

+1

lgeiger:
@n-riesco > I'm using:
```
  "missing_fat_arrows": {
    "level": "error",
    "is_strict": false
  },
```

This rule breaks functionality in `kernel-manager.coffee`.

I'm using the [`linter-coffeelint`](https://github.com/AtomLinter/linter-coffeelint) atom package which uses `coffeelint 1.9.7`.

lgeiger:
The coffeelint website says this about `missing_fat_arrow`:
> Warns when you use `this` inside a function that wasn't defined with a fat arrow. This rule does not apply to methods defined in a class, since they have `this` bound to the class instance (or the class itself, for class methods). The option `is_strict` is available for checking bindings of class methods. It is impossible to statically determine whether a function using `this` will be bound with the correct `this` value due to language features like `Function.prototype.call` and `Function.prototype.bind`, so this rule may produce false positives.
default level: ignore

I think we should remove this rule.

n-riesco:
On 09/07/16 14:20, Lukas Geiger wrote:
> The coffeelint website says this about |missing_fat_arrow|:
> 
> I think we should remove this rule.

I find it very convenient with callbacks. It's very easy to forget a fat arrow within a callback created inside a class method.

The README.md in https://github.com/AtomLinter/linter-coffeelint says that it is possible to use alternative coffeelint versions by specifying the version in our project's package.json.

lgeiger:
Coffeelint 1.15.7 highlights the same errors as version 1.9.7

lgeiger:
I don't have a problem with the rule. If I change the arrows in `kernel-manager` according to `coffeelint` Hydrogen fails.
I'm still new to coffescript. @n-riesco Can you take a look why after fixing the `missing_fat_arrow` errors hydrogen fails?

n-riesco:
On 09/07/16 14:49, Lukas Geiger wrote:
> Coffeelint 1.15.7 highlights the same errors as version 1.9.7

Could you run `node_modules/.bin/coffeelint --rules coffeelint.json lib/main.coffee` in the terminal?

I've tried both 1.9.7 and 1.15.7 without problems.

I'm gonna try linter-coffeescript next.

n-riesco:
On 09/07/16 14:55, Lukas Geiger wrote:
> I don't have a problem with the rule. If I change the arrows in |kernel-manager| according to |coffeelint| Hydrogen fails.
> I'm still new to coffescript. @n-riesco <https://github.com/n-riesco> Can you take a look why after fixing the |missing_fat_arrow| errors hydrogen fails?

The rule `missing_fat_arrows` gets confused with `KernelManager` (with `Hydrogen` too), because `KernelManager` is not defined as a class.

I've been meaning to re-write `KernelManager` a class for a while, so I guess this could be a good time.

I'm not sure about `Hydrogen`, though.

The examples in the Atom documentation use plain objects to define a package. This works just fine as long methods are always invoked using the dot notation; i.e. this works:

```
Hydrogen.activate(state)
```

but this wouldn't:

```
activate = Hydrogen.activate
activate(state)
```

I'm not going be able to look at this until the evening.
I will convert `KernelManager` to a class and see what the rule `missing_fat_arrows` reports.
Then, we can continue the discussion.

lgeiger:
I rebased this PR and did some additional refactoring of the run commands

lib/main.coffee
```diff
@@ -232,64 +232,46 @@ module.exports = Hydrogen =
...
         code = @getRows(startRow, endRow)
+
         if code?
+            if moveDown is true
+                @moveDown endRow
             @createResultBubble code, endRow
 
-    runCellAndMoveDown: () ->
-        [startRow, endRow] = CellManager.getCurrentCell()
+    escapeBlankRows: (startRow, endRow) ->
```
n-riesco: nice abstraction

PR #330, opened 2016-07-07
Support for remote kernels via kernel gateway

nikitakit:
Thanks to everyone for developing hydrogen!

I'm really happy with the features it provides, but unfortunately there are a few things that make it really hard to use with my everyday workflow. I'd like to share a rather substantial set of changes I made to get remote kernel support into a state where I can use it. (The code is still pretty rough; I haven't cleaned it up yet)

The work was driven by two guiding principles:
1. Make remote kernels first-class citizens
2. Never accidentally kill a kernel

The second point might seem obvious, but it is the main reason for using kernel gateways. ZMQ-forwarding approaches (like `rk`) tend to fail pretty badly on this regard. When the SSH connection to the server drops, they tend to activate cleanup routines that basically assume that the kernel is dead. This usually makes something as trivial as disconnecting from Wifi effectively kill a remote kernel. The desired behavior is tricky enough to get right that I know of only two implementations that can do it. They are the jupyter notebook and the jupyter kernel gateway (an official Jupyter project that mimics the programmatic API of the notebook)

Also remote kernels shouldn't be covered by hydrogen's resource management, which tends to destroy kernels on multiple codepaths (e.g. when hydrogen is disabled)

Changes in this patchset:
* Connect to kernels via kernel gateways instead of over ZMQ
* Specify remote endpoints and authentication credentials via settings
* Replaces the kernel switcher UI with a session switcher that can select among
local and remote sessions
* Kernel connections can survive survive both network failures and Atom restarts

For simplicity I implemented this as a replacement to ZMQ-based connections. This adds a hard dependency on jupyter and either jupyter_kernel_gateway or jupyter_notebook for local connections, but I don't see that as a problem because you're going to have to install some local kernels anyway.

At the moment it requires manually starting a notebook or gateway on localhost:8888, but if there is interest in the approach I think hydrogen could spawn its own gateway automatically.

## Quick setup for testing

* Start a Jupyter notebook (or jupyter kernel gateway) on localhost:8888
* Run the modified version of Hydrogen

Many of the more advanced features haven't been ported (so they don't work), but core execution functionality is there.

I think the current state of the patch covers enough functionality to support a discussion about the merits of the approach.


lgeiger:
Thanks a lot for this PR.

I really like the idea of moving towards kernel gateways. It would give the user total control of kernels. I also quite like the workflow of choosing a kernel before execution.

I think this could completely replace ZMQ and therefor the number of ZMQ related installation issues. For this we have to come up with a simple way of starting local (and remote) kernels and more important decide when a kernel should be destroyed.

@rgbkrk @n-riesco @slavaGanzin @willwhitney What do you think about a move toward kernel gateways?

rgbkrk:
@lgeiger I'm certainly a fan of the kernel gateway approach, though you're not going to escape all the "ZMQ related problems". If you'd like to take a stab at an Atom plugin that folks are using that uses the kernel gateway underneath, check out https://github.com/jupyter/atom-notebook.

My main fears with trying to exclusively rely on a separate process and CLI that the user installs is that we have zero control over what version is installed (the interface) as well as helping them with migrations (including the first one). The zeromq story can be fixed and I'm back in a situation to carve out more of that in the coming weeks. @hintjens made good progress with libzmq and czmq for us, even setting up some initial bindings - we need to finish up the prebuild work.

All that being said, I'm a fan of handling remote kernels in this way especially if using a token based auth approach (as you can do with kernel gateway). How are you escaping the cross origin websocket security protections @nikitakit?

nikitakit:
I think if we take a gateway-only approach, it makes sense for hydrogen to ship with its own gateway config. It also probably wouldn't be too hard to figure out how to autospawn the gateway using this config. This config can ensure that the correct gateway features are enabled, and maybe even perform some version checking to cleanly notify the user of incompatible versions.

This does exclude systems that only have an ancient version of `ipykernel` installed, but I don't think it's that hard to ask people to install jupyter. Regarding incompatibilities, the gateway and `jupyter-js-services` already live in separate repos and each have their own set of dependencies. I don't think things are unstable to the point where the developers of these libraries would suddenly break things.

The initial upgrade is its own story. It's certainly going to be distracting for people to suddenly need to install gateways, but I'm not sure how many people have locked-in dependencies that would prevent an upgrade.

The reasons I'm suggesting shipping our own gateway config is also because I think it's useful to change two defaults:
* Enabling listing kernels (the notebook does this by default, but the gateway disables it)
* Reconfigure cross-origin protection to allow hydrogen through (the gateway only runs on local ports, so it should be fine)

@rgbkrk I'm not really a web developer, so I didn't realize that there wasn't a standard solution to the cross-origin issues. I think a gateway config change can deal with them. The other way to avoid them, which I'm doing right now, is:
```coffeescript
ws = require('ws');
xhr = require('xmlhttprequest');
global.XMLHttpRequest = xhr.XMLHttpRequest;
global.WebSocket = ws;
```

n-riesco:
This PR is quite a substantial change. Personally, I would advise against this kind of substantial changes without prior discussion.

If this discussion had happened, I would have a chance to say this: *Please, don't remove functionality before we all have agreed*.

I think gateways are a great addition to Hydrogen, but I'd like to see them live along ZMQ kernels, so that we have a chance to evaluate the feasibility of a replacement.

From what I've read in this issue, I don't believe they can be a replacement if they require the installation of additional software (Hydrogen should work out of the box).

Our experience up to now is that if we require the installation of additional software (as we do now with ZMQ and Jupyter) we end up expending a lot of time supporting not only Hydrogen but also this additional software.

---

@rgbkrk @lgeiger @slavaGanzin @willwhitney Should we update the contribution guidelines to advise that substantial changes need discussing in issue.

nikitakit:
@n-riesco I apologize, my last message was too strongly worded. I should have made it more clear that I was merely making suggestions as a hydrogen user.

I would like to clarify that I never suggested merging this pull request in its present state. Between the rough state of the code (it doesn't even pass tests!) and the disclaimers in the text, I thought this would be readily apparent.

The code here was written for my own day-to-day use. It has already given me positive return-on-investment, and would have been worth writing even if I had not published it at all.

But since I already had the code, I decided I'd follow the spirit of open source and share it with the hydrogen team. Opening a PR seemed nice because GitHub's diff views look really good. I was hoping that this would be seen as part of an ongoing discussion about this feature.

n-riesco:
@nikitakit I understand.

The problem with a PR larger than necessary is that reviewers have to do more work and give less feedback than necessary.

Here's my somewhat scant feedback:
- as said above, please keep previous functionality and minimise the size of this PR.
- `config.coffee`: having a description would've helped with the review
- `kernel-manager.coffee`: I agree with [this comment](https://github.com/nteract/hydrogen/pull/330/files#diff-2e1bb36917ff76ab7308da52bc968501R111)
- `remote-kernel.coffee`: Rename `class Kernel` so that we can keep using ZMQ kernels
- I noticed some `global` variables are defined, is it really necessary?

rgbkrk:
> Should we update the contribution guidelines to advise that substantial changes need discussing in issue.

If someone feels impassioned to write code, they take a risk in submitting a big PR. Sometimes it's easier to communicate with code instead of just issues. We're not going to merge this, but I certainly want to have the discussion.

The work can always be split up - the remote kernel approach is why enchannel abstracts across using any of our available transports: zmq, websockets via notebooks, and socket.io websockets. We haven't yet done it in Hydrogen and we need people to explore approaches.

nikitakit:
I've just started one more re-do of the patchset. (For anyone reading the diffs, I recommend reading one commit at a time)

Last time I started working by creating a copy of the Kernel class, but it turns out that all of the hard parts of adding kernel gateways are elsewhere.

Currently the kernel management code couples language-to-grammar mappings, kernel-spawning functionality, and ZMQ all in one. The only way to get remote kernels in there without abandoning ZMQ is to split that apart and add extra abstraction. `jupyter-js-services` has a well-made API, so I've decided to abstract the ZMQ code into a few classes that are compatible with that API, to the extent needed by hydrogen.

I looked into using the enchannel API instead of the `jupyter-js-services` API, but unfortunately it doesn't provide the required functionality. It's missing key functions like listing kernel specs, enumerating running kernels, etc. Also the messaging implemention seems to be lower-level than `jupyter-js-services` (it doesn't appear to automatically associate messages with their parent and call the appropriate callback).

Summary of changes
* Commit 1 targets KernelManager. Anything relating to kernelspecs and spawning kernels gets moved into a ZMQGateway class, while KernelManager's sole purpose is to switch between kernels based on the active language/grammar.
* Commit 2 deals with kernel specs. Other than the newly-created ZMQGateway class, the rest of the code shouldn't need to access to the full kernel specs, it only really needs the kernel names and display names. It also makes any kernel spec accesses async.
* Commit 3 splits the Kernel class into a KernelInterface class (hydrogen-specific) and a ZMQSession class (designed to parallel the session class in jupyter-js-services). The session abstraction is used by jupyter-js-services to allow for restarting or switching kernels in-place, though I'm not yet taking advantage of that.

These commits shouldn't break any functionality (other than the test cases, which I'll get to later). There's also nothing new added yet, but I'll get to that after a brief pause for rest (and discussion).

lib/zmq-services.coffee
```diff
@@ -0,0 +1,216 @@
...
+
+# We follow jupyter-js-services in exporting global methods, so we can't just
+# export the ZMQGateway class and rely on it being constructed externally.
+class ZMQServices
+    constructor: ->
+        @zmqGateway = null
+
+    getZMQGateway: ->
+        unless @zmqGateway?
+            startZMQGateway()
```
nikitakit: Should be `@start()`

rgbkrk:
Cool, thanks for giving this a go @nikitakit. Checking it out now.

Enchannel is meant to be lower level on purpose, with a higher level setup tailored to how you spawn, list kernels, etc. which can vary. That part of the API hasn't been extracted to a separate piece yet.

lib/kernel-manager.coffee
```diff
@@ -6,12 +6,12 @@ path = require 'path'
 Config = require './config'
 ConfigManager = require './config-manager'
 Kernel = require './kernel'
+zmq_services = require './zmq-services'
```
n-riesco: Use Javascript convention, i.e. ZMQServices

lib/kernel-manager.coffee
```diff
@@ -98,7 +74,13 @@ class KernelManager
 
 
     getAllKernelSpecs: ->
-        return _.map @_kernelSpecs, 'spec'
+        # TODO(nikita): switch to async and jupyter-js-services API
+        # Note that using sync here is a temporary regression, because it forces
+        # a manual call to update-kernels by the user.
+        # This will be fixed in a later patch, that adds proper async back
+        kernelSpecs = zmq_services.getZMQGateway().getKernelSpecsSync()
+        _.forEach kernelSpecs, (obj, name) -> obj.spec.name = name
```
n-riesco: do we need another kind of name?

lib/kernel.coffee
```diff
@@ -11,20 +11,18 @@ WatchSidebar = require './watch-sidebar'
 
 module.exports =
 class Kernel
-    constructor: (@kernelSpec, @grammar, @config, @configPath, @onlyConnect = false) ->
+    constructor: (@onDestroy, @kernelName, @kernelSpec, filePath, @config, @configPath, @onlyConnect = false) ->
         console.log 'Kernel spec:', @kernelSpec
         console.log 'Kernel configuration:', @config
         console.log 'Kernel configuration file path:', @configPath
-        @kernelName = @kernelSpec.display_name
+        @displayName = @kernelSpec.display_name
```
n-riesco: Rather than rename it, I would remove it.
No need to duplicate the info already provided in the kernel spec.

lib/kernel.coffee
```diff
@@ -11,20 +11,18 @@ WatchSidebar = require './watch-sidebar'
 
 module.exports =
 class Kernel
-    constructor: (@kernelSpec, @grammar, @config, @configPath, @onlyConnect = false) ->
+    constructor: (@onDestroy, @kernelName, @kernelSpec, filePath, @config, @configPath, @onlyConnect = false) ->
```
n-riesco: we need to reduce the number of arguments this constructor is taking.

I think `filePath` is missing a `@`

lib/kernel.coffee
```diff
@@ -496,3 +494,5 @@ class Kernel
 
         unless @onlyConnect
             @kernelProcess.kill 'SIGKILL'
+
+        @onDestroy(this)
```
n-riesco: ` @onDestroy?(this)`

lib/main.coffee
```diff
@@ -11,6 +11,8 @@ Config = require './config'
 KernelManager = require './kernel-manager'
 Inspector = require './inspector'
 AutocompleteProvider = require './autocomplete-provider'
+zmq_services = require './zmq-services'
```
n-riesco: We just refactor `KernelManager` into a class so that we don't have a singleton.
Please, don't make `ZMQServices` a singleton.

lib/kernel-manager.coffee
nikitakit: Consequence of singleton status, will comment there

lib/kernel-manager.coffee
nikitakit: Gateway kernel spawning works using the kernel name (not display_name), which is not part of the kernelspec.

This whole method gets rewritten in patch 2, where I address this in a more comprehensive way.

lib/kernel.coffee
nikitakit: Actually, nothing outside the constructor needs it.

I think I'll keep it constructor-local, and fix the misreference below.

lib/kernel.coffee
```diff
@@ -11,20 +11,18 @@ WatchSidebar = require './watch-sidebar'
...
         @watchCallbacks = []
 
         @watchSidebar = new WatchSidebar(this)
-        @statusView = new StatusView(@kernelName)
+        @statusView = new StatusView(@displayName)
 
-        projectPath = path.dirname(
-            atom.workspace.getActiveTextEditor().getPath()
-        )
+        projectPath = path.dirname(@filePath)
```
nikitakit: Extra `@`: should be local variable instead

lib/kernel.coffee
nikitakit: `onDestroy` should never be null, because that would be a resource leak. (It's needed to de-list the session from the ZMQGateway. A callback is the easiest way to avoid circular references)

lib/main.coffee
nikitakit: Seems doable in retrospect, but I'll need to check after writing the gateway code.

`jupyter-js-services` provides a number of functions as globals. I'll need to make sure that any place those globals are accessed has a reasonable way to obtain a `ZMQGateway` instance from either main or the kernel manager.

lib/kernel-manager.coffee
```diff
@@ -73,39 +90,13 @@ class KernelManager
         return grammar?.name.toLowerCase()
 
 
-    getAllKernelSpecs: ->
-        # TODO(nikita): switch to async and jupyter-js-services API
-        # Note that using sync here is a temporary regression, because it forces
-        # a manual call to update-kernels by the user.
-        # This will be fixed in a later patch, that adds proper async back
-        kernelSpecs = zmq_services.getZMQGateway().getKernelSpecsSync()
```
nikitakit: Remember to delete `getKernelSpecsSync`, which is no longer needed/used

n-riesco:
Here's some feedback for the first commit 	a3dc04b  :

* I don't see the need to split `KernelManager` into `KernelManager` and `ZMQServices`. Why do you think this is necessary?

* In Hydrogen-speak, a `Kernel` instance is what this PR calls a `Session`. I don't mind to rename Kernel as Session, but I would do it in a different PR, so that here we can focus on the gateway functionality.

---

I haven't got to review the other 2 commits yet. But, just to help me have an idea of what approach you're trying to implement, here are some questions:

* Currently `KernelManager` is the class responsible for creating a `Kernel` instance that can handle requests for a given grammar. To achieve this, `KernelManager` uses kernel specs. Do gateway kernels have a kernel spec? Note that other parts in Hydrogen assume that a `Kernel` has been started using a kernel spec and that this kernel spec is accessible through `Kernel#kernelSpec`.

* Is the gateway able to provide a list of available kernel specs? If this is the case, then I reckon the only changes necessary in Hydrogen would be:
  * implement a `KernelManager#getKernelSpecsFromGateway`
  * a new class, e.g. `GatewayKernel`, equivalent to the existing `Kernel`.
  * and update `KernelManager#startKernel` to instanciate `GatewayKernel` or `Kernel` as required by the kernel spec.

---

I'll try to go through the other 2 commits tomorrow. But I'll be away and my access to internet may be limited.

lib/kernel-manager.coffee
n-riesco: Hydrogen uses `display_name`s that are part of the kernel spec (see [here](http://jupyter-client.readthedocs.io/en/latest/api/kernelspec.html)).

We've gone through a lot of trouble to minimise the names used in Hydrogen. Currently, we use only two:
- lowercased grammar languages (to determine what kernels can run a file)
- display_names (this is the name shown to the user to identify the kernel)

Can we make do without using that kernel name?

lib/kernel.coffee
n-riesco: I find unusual to have `@onDestroy` as the first argument.

lib/kernel.coffee
n-riesco: I don't see why *never*. Current `Kernel` doesn't need an `onDestroy`.

lib/main.coffee
n-riesco: The way this is done with `KernelManager` is:
- main creates an instance called `@kernelManager`
- any other classes that need the kernel manager, get a copy of `@kernelManager` through the constructor.

lib/kernel-manager.coffee
n-riesco: Can we call it something less ambiguous?
I don't want people to start using `spec.name` for anything else than spawning a gateway kernel.
Perhaps `spec._name`?

lib/main.coffee
n-riesco: @rgbkrk Could you comment about the globals defined by `jupyter-js-services`?
Does this interfere with any of the globals that Atom itself overwrites?
Is `nteract` already using `jupyter-js-services` as dependency?

nikitakit:
The approach you're suggesting is similar to what I tried in the first version of this PR, but there are enough API mismatches that I think merit additional abstraction.

* `KernelManager` is currently a combination of grammar-to-kernel mapping and subprocess/filesystem operations, which are conceptually different roles. This makes sense for local kernels, because hydrogen oversees their creation, destruction, and resource management. But for remote gateways, it is vital to support attaching to and detaching from existing kernels. I think it is good abstraction to separate the resource management (which applies only to local kernels) and grammar-to-kernel mappings into different classes.
* Implementing an independent `GatewayKernel` class will result a fair bit of copy-pasted code. Things like the status view, watch sidebar, and ensuring that watches get called at the right time would all be duplicated. My plan (not yet fully implemented) is to put all the shared bits into `KernelInterface`, and move all the ZMQ-specific stuff into `ZMQSession`. I don't particularly care what the classes are named, but the class specific to `jupyter-js-services` is already called `Session`.

The use of kernel specs throughout Hydgrogen is also a point of API mismatch. Kernel spawning operations in `jupyter-js-services` take a kernel name, which is not a part of the kernel spec. There also needs to be a parameter that selects between different gateways. For consistency reasons, I've decided to update most of Hydrogen to use an alternative datastructure that includes the spec, but also has the kernel name and gateway information. (I'm calling it the Kernel Model to avoid naming confusion). This is started in patch 2, and will continue when remote gateways are actually added.

lib/main.coffee
rgbkrk: One package in nteract is using `jupyter-js-services` as a prototype for using nteract UIs on top of Jupyter notebook backends. I can't comment much on `jupyter-js-services` as I haven't been using it. @jdfreder may know better.

lib/kernel-manager.coffee
nikitakit: This line gets deleted in the very next patch. It is here only to allow running this patch standalone without applying the next.

lib/kernel.coffee
nikitakit: The current kernel doesn't need `onDestroy` because it is killed by calling KernelManager (now ZMQGateway), not the kernel itself.

I am changing this such that calling the kernel's destroy directly is sufficient. To do this the kernel must know how to unlist itself from the ZMQGateway, which requires either a pointer to the ZMQGateway or a callback.

nikitakit:
Thanks for the feedback!

It'll probably be another day before I can get to making the actual gateway commits, but that leaves time to mull things over in the meanwhile.

n-riesco:
On 15/07/16 05:55, Nikita Kitaev wrote:
> Thanks for the feedback!
> 
> It'll probably be another day before I can get to making the actual gateway commits, but that leaves time to mull things over in the meanwhile.

I wouldn't commit anything else for the time being.

My current view is that this PR is touching code in Hydrogen and breaking conventions that it shouldn't.

I do believe we would be more effective here if we discuss first how to implement this PR. We can point you to the places we think changes could go, and make you aware of the conventions we are using (especially since the code is so scarcely documented).

n-riesco:
@nikitakit I'm sorry to be a pain, but after having a look at `jupyter-js-services`, I feel more convinced that this PR doesn't take full advantage of what `KernelManager` already provides.

This is how I think we could proceed:

1. In `main.coffee`, launch the gateway when Hydrogen is activated (the example launch script [here](https://github.com/jupyter/jupyter-js-services/blob/master/examples/node/main.py) seems enough).

2. In `main.coffee`, kill the gateway when Hydrogen is deactivated.

3. Create a `KernelManager#getKernelSpecsFromGateway` that uses `require('jupyter-js-services').getKernelSpecs` to get the list of kernel specs (similarly to `KernelManager#getKernelSpecsFromJupyter`).

4. Here's the most invasive step that I think is best to attack last. As you said, there is functionality in `Kernel` that is common to both ZMQ and gateway kernels. Let's keep this common functionality in `Kernel` and create two extended classes, e.g. `KernelZMQ` (with the old functionality) and `KernelXXX` (with the new functionality provided by `jupyter-js-services`).

Since @lgeiger is working on `KernelManager` (see #351), I'd start with 1 and 2. Hopefully by then, #351 will be merged, and you could continue with 3. We can tackle 4, once we're happy with the solution for 3.

nikitakit:
Looks like @lgeiger has addressed a some of the functionality currently in this PR, especially in commit 2 (in particular: making kernelspecs async, plus a few other things). I'll rebase on top of those changes once they get merge. That will decrease the overall diff for this PR.

#### Splitting up Kernels
Regarding point 4: that's essentially what I'm starting in commit 3 here. The kernel class is split up into `ZMQSession` and `jupyter-js-services::Session`, with any common functionality moved into `KernelInterface`.

#### Desired functionality
With regards to your points 1-3, I think the reason for confusion is that I haven't articulated the user workflow I'm striving for. So I'll first take a step back and list the features I would like to support (which have led to my deliberate decision to change some of the architecture)

Desired functionality:
* Connect to any number of remote gateways by specifying their URL + auth tokens
* Attach to and detach from remote kernels
* Actions should **never** kill remote kernels or shut down remote gateways. (Exception: a choice to activate "restart kernel" or "shutdown kernel" by the user, from the menu)
* Switch between multiple remote kernels per language/grammar (but have no more than 1 active at a time)

`getKernelSpecsFromGateway` is not sufficient to attach to an already existing remote kernel. Neither is `connection.json`, which is both ZMQ-only and inconvenient. Instead, I have a nice session-picker widget that can return a `jupyter-js-services::Session`, which can then be wrapped by `KernelManager`.

#### Splitting up KernelManager
`KernelManager` currently has code in two functional categories.

Category 1:
* Sending OS signals to kernels
* Calling `ipython`/`jupyter` as child processes
* Reading local kernelspecs from settings
* Managing ZMQ connection files.
* Killing kernels (again, it is vital that remote kernels are **never** accidentally killed)

Category 2:
* Associating running kernels to grammars
* Filtering kernelspecs by language
* Starting new kernels based on current grammar
* (To be implemented in this PR) Attaching to existing kernels

I decided to split these two categories into two different classes: `ZMQGateway` for category 1, and `KernelManager` for category 2. I don't see what the big deal with this is.

nikitakit:
I should add regarding your points 1 and 2: I'm leaving those for a future PR. As long as ZMQ functionality is not removed, it can serve local kernels well. As for remote kernels, the whole point is that they get spawned externally.

n-riesco:
On 15/07/16 22:30, Nikita Kitaev wrote:
> Desired functionality:
> 
>   * Connect to any number of remote gateways by specifying their URL + auth tokens
>   * Attach to and detach from remote kernels
>   * Actions should *never* kill remote kernels or shut down remote gateways. (Exception: a choice to activate "restart kernel" or "shutdown kernel" by the user, from the menu)
>   * Switch between multiple remote kernels per language/grammar (but have no more than 1 active at a time)

This is very ambitious for a single PR.

What is the most basic use-case that need?

lgeiger:
> Looks like @lgeiger has addressed a some of the functionality currently in this PR, especially in commit 2 (in particular: making kernelspecs async, plus a few other things). I'll rebase on top of those changes once they get merged. That should decrease the overall diff for this PR.

Sorry for interfering with your PR. I started working on some of the things discussed in #272 and thought it would make things a bit clearer. Turns out there was a fair amount of rewriting required in my PR.

nikitakit:
The most basic use-case is to attach to and detach from notebooks running on an existing notebook server, without ever killing any of them.

Yes, it's ambitious. But version 1 of this PR already had all of these features.

FYI no need to review the current set of changes, because I'm rewriting a number of things. I'm switching to Kernel/ZMQKernel/GatewayKernel inheritance architecture, which also means that I need to approach the KernelManager differently. I should have commits up in a few hours (I'm also rebasing on top of #351 as I go)

n-riesco:
I don't know how else I can say this:
- I'm reluctant to merge large PRs
- It's exhausting to review large PRs
- Please, let's agree on a small change.

Let's focus on the use-case  *"attach to and detach from notebooks running on an existing notebook server"*:

- Does this mean you want to implement a command like `hydrogen:connect-to-jupyter-service`?

- How would this command work? Would the user be asked for a URL?

If so, I think this PR could start implementing such a command. The initial implementation doesn't have to be fully functional. It'd be enough:
- prompt the user for an URL
- connect to the jupyter service
- show a menu with the running kernels

Once this is solved, we can design how to create a `Kernel` class that uses a Jupyter session.

nikitakit:
Done with rewrites for today.

I was busy coding, so sorry I couldn't respond to your message earlier. Most of the stuff I said above is now superceded.

The thing is that I have to write the hard bits first, not last, to actually end up with an architecture that makes sense.

`Kernel` class is now split up into an inheritance tree (as you suggested), and there's a new command for connecting to remote gateways.

The session picker lets you select from URLs you saved in the settings, rather than prompting each time. It then shows you a list of running kernels (or lets you start a new one). A manual URL prompt would be nice eventually, but this is the simpler change.

Changes to KernelManager are now just adding one function. The rest have been superceded by #351, or dropped (I no longer have plans to support "switch kernel" for gateways in this PR, among other features).

lgeiger:
Thanks a lot for your rewriting! 👍 

As I'm still fairly new to coffeescript I'll start just from a users perspective. I know this PR isn't completely finished yet but here are some things I noticed:

- I didn't found any problems on the `ZMQ` side 🎉 👍 .
- Getting the kernelspecs from a gateway isn't working for me.
  - If I use `jupyter kernelgateway` to start a gateway, `services.getKernelSpecs` throws `API request failed (403):  null` and `Uncaught (in promise) TypeError: specmodels.map is not a function`
  - If I use `jupyter notebook` to start a gateway. Only  `Uncaught (in promise) TypeError: specmodels.map is not a function` is thrown.
- The status bar isn't updated after connecting to a kernel. It has to be set similar to [this](https://github.com/nteract/hydrogen/blob/master/lib/main.coffee#L279).

lib/config.coffee
```diff
@@ -35,6 +23,23 @@ module.exports = Config =
...
+            [{
+                "name": "Remote notebook",
+                "options": {
+                        "baseUrl": "http://mysite.com:8888"
+                }
+            }]
+            ```
+            '
+            type: 'string'
+            default: '[{}]'
```
lgeiger: It should default to `[]`. Otherwise the session picker will always show an empty item.

lib/session-picker.coffee
```diff
@@ -0,0 +1,112 @@
...
+class SessionPicker
+    constructor: (onChosen) ->
+        @_onChosen = onChosen
+
+    toggle: ->
+        @_path = atom.workspace.getActiveTextEditor().getPath() + '-' + uuid.v4()
+        gatewayListing = new CustomListView('No gateways available', @onGateway.bind this)
+        gatewayListing.setLoading 'Loading gateways...'
+        items = Config.getJson('gateways')
+        if items?
```
lgeiger: `if not _.isEmpty items` would work here since `Config.getJson` returns an empty array on default.

lib/session-picker.coffee
```diff
@@ -0,0 +1,112 @@
...
+            gatewayListing.setItems items
+            gatewayListing.setError 'Select a gateway' # TODO(nikita): maybe don't misuse error
+        else:
```
lgeiger: `else:` isn't working in coffescript. Use `else` instead.

nikitakit: Right. You can probably tell which language I normally program in :wink:

lib/gateway-kernel.coffee
```diff
@@ -0,0 +1,97 @@
...
+        # TODO(nikita): implement watches
+        # Also consider switching to user_expressions for watches
+        console.log 'Watches are not supported by remote kernels'
+
+    complete: (code, onResults) ->
+        @session.kernel.complete(
+            code: code
+            cursor_pos: code.length
+        ).then((message) ->
+            onResults message.content.matches
```
nikitakit: Should be `onResults message.content` instead, otherwise autocomplete doesn't work.

nikitakit:
@lgeiger Thanks for testing!

You're right, at some point I stopped testing creating new sessions, and it's currently broken. The fix is to change line `session-picker.coffee:95` to say `items = _.map specmodels.kernelspecs, (specmodel) =>` instead.

The 403 error is normal for gateways: they return access denied when you attempt to list running sessions. I have an error handler that addresses this case, but I think it gets logged anyway somewhere inside `jupyter-js-services`.

lib/session-picker.coffee
```diff
@@ -0,0 +1,112 @@
...
+                options: spec.options
+
+    onSession: (spec) ->
+        unless spec.model?
+            kernelListing = new CustomListView('No kernel specs available', @startSession.bind this)
+            kernelListing.setLoading 'Loading kernel specs...'
+
+            services.getKernelSpecs(spec.options)
+            .then (specmodels) =>
+                items = specmodels.map (specmodel) =>
```
nikitakit: Should say `items = _.map specmodels.kernelspecs, (specmodel) =>`

lib/kernel-manager.coffee
```diff
@@ -79,11 +73,17 @@ class KernelManager
...
+                kernel = new ZMQKernel(
                     kernelSpec, grammar, config, filepath, onlyConnect = false
                 )
                 finishKernelStartup kernel
 
+    attachToGatewaySession: (grammar, session, onAttach) ->
+        language = @getLanguageFor grammar
+        gatewayKernelFromSession session, (kernel) =>
+            # No startup code is run when attaching to existing sessions
+            @_runningKernels[language] = kernel
```
nikitakit: Make sure to call `onAttach?(kernel)`

lib/main.coffee
```diff
@@ -349,6 +363,14 @@ module.exports = Hydrogen =
                     @watchSidebar.show()
         @watchKernelPicker.toggle()
 
+    showSessionPicker: ->
+        unless @sessionPicker?
+            @sessionPicker = new SessionPicker (session) =>
+                grammar = @editor.getGrammar()
+                @kernelManager.attachToGatewaySession grammar, session
```
nikitakit: Use onAttach callback argument to call `@setStatusBarElement()`

lgeiger:
I tested it and it works great 👍 

I would suggest you add `hydrogen:connect-to-remote-session` to the [activation commands](https://github.com/nteract/hydrogen/blob/master/package.json#L20).

nikitakit:
:tada:

I was thinking about adding it to activation commands, too (I'll do that next time I rebase).

I also think that command needs a different name, maybe something like "Connect to Jupyter Gateway". It's used rarely enough to not merit a keyboard shortcut, but as it stands I constantly forget what keywords to input into the command fuzzy-finder. Selecting kernels is nicer in that the word "kernel" alone brings that command into the top 3 fuzzy-find results. I'd love to get suggestions on alternate names!

Other than this small issue, the code has reached a state where I'm comfortable signing off on it, so I've removed the [DEMO] tag from the PR title.

lib/kernel.coffee
```diff
@@ -11,312 +11,31 @@ WatchSidebar = require './watch-sidebar'
...
-        console.log 'Kernel configuration:', @config
-        console.log 'Kernel configuration file path:', @configPath
-        @kernelName = @kernelSpec.display_name
+    constructor: (@kernelSpec) ->
         @executionCallbacks = {}
         @watchCallbacks = []
 
         @watchSidebar = new WatchSidebar(this)
-        @statusView = new StatusView(@kernelName)
+        @statusView = new StatusView(@kernelSpec.display_name)
```
n-riesco: Thanks for cleaning this up!

lib/kernel.coffee
```diff
@@ -11,312 +11,31 @@ WatchSidebar = require './watch-sidebar'
...
-                allow_stdin: false
-
-        message =
-                header: header
-                content: content
-
-        @executionCallbacks[requestId] = onResults
-
-        @shellSocket.send new jmp.Message message
+        console.log 'Abstract method'
```
n-riesco: `console.error 'Kernel: interrupt method not implemented'`

or

`throw new Error 'Kernel: interrupt method not implemented'`

I think throwing is better because we can always catch an exception.

lib/kernel.coffee
```diff
@@ -11,312 +11,31 @@ WatchSidebar = require './watch-sidebar'
...
 
     execute: (code, onResults) ->
-        console.log 'Kernel.execute:', code
-
-        requestId = 'execute_' + uuid.v4()
-        @_execute(code, requestId, onResults)
+        console.log 'Abstract method'
```
n-riesco: ditto

lib/kernel.coffee
```diff
@@ -11,312 +11,31 @@ WatchSidebar = require './watch-sidebar'
...
 
     executeWatch: (code, onResults) ->
-        console.log 'Kernel.executeWatch:', code
-
-        requestId = 'watch_' + uuid.v4()
-        @_execute(code, requestId, onResults)
+        console.log 'Abstract method'
```
n-riesco: ditto

lib/kernel.coffee
```diff
@@ -11,312 +11,31 @@ WatchSidebar = require './watch-sidebar'
...
-
-        message =
-                header: header
-                content: content
-
-        @executionCallbacks[requestId] = onResults
-
-        @shellSocket.send new jmp.Message message
-
+        console.log 'Abstract method'
```
n-riesco: ditto

lib/kernel.coffee
```diff
@@ -11,312 +11,31 @@ WatchSidebar = require './watch-sidebar'
...
-            console.log 'Invalid message: Missing header.msg_id'
-            return false
-
-        unless message.header.msg_type?
-            console.log 'Invalid message: Missing header.msg_type'
-            return false
-
-        return true
-
+        console.log 'Abstract method'
```
n-riesco: ditto

lib/kernel.coffee
```diff
@@ -463,36 +182,5 @@ class Kernel
...
-
-        if @onlyConnect
-            detail = 'Shutdown request sent to custom kernel connection in ' +
-                @configPath
-            atom.notifications.addInfo 'Custom kernel connection:',
-                detail: detail
-
-        unless @onlyConnect
-            @kernelProcess.kill 'SIGKILL'
+        console.log 'Destroying base kernel'
```
n-riesco: ditto

lib/zmq-kernel.coffee
```diff
@@ -0,0 +1,343 @@
...
+
+Kernel = require './kernel'
+
+
+module.exports =
+class ZMQKernel extends Kernel
+    constructor: (kernelSpec, @grammar, @config, @configPath, onlyConnect = false) ->
+        super kernelSpec
+
+        @executionCallbacks = {}
```
n-riesco: `@executionCallbacks` is already set by the parent constructor.

lib/zmq-kernel.coffee
```diff
@@ -0,0 +1,343 @@
...
+jmp = require 'jmp'
+uuid = require 'uuid'
+zmq = jmp.zmq
+
+Kernel = require './kernel'
+
+
+module.exports =
+class ZMQKernel extends Kernel
+    constructor: (kernelSpec, @grammar, @config, @configPath, onlyConnect = false) ->
```
n-riesco: `@onlyConnect` needs to be an instance property, otherwise `@destroy` won't work.

lib/jupyter-js-services-shim.coffee
```diff
@@ -0,0 +1,18 @@
+# jupyter-js-services apparently require overriding globals
+# This file exists to
+# a) Make sure globals are only ever overridden once
+# b) In the future, try to make the global overrides optional if gateways are
+#    not used, or have been pre-configured to avoid CORS issues
+
+services = require('jupyter-js-services')
```
n-riesco: `module.exports = require 'jupyter-js-services'`

lib/jupyter-js-services-shim.coffee
```diff
@@ -0,0 +1,18 @@
...
+
+# Set up global overrides, needed by jupyter-js-services
+# TODO(nikita): would like to not override globals, but there are CORS issues
```
n-riesco: Is there a link/reference that you could add to this comment?

lib/jupyter-js-services-shim.coffee
```diff
@@ -0,0 +1,18 @@
+# jupyter-js-services apparently require overriding globals
+# This file exists to
+# a) Make sure globals are only ever overridden once
+# b) In the future, try to make the global overrides optional if gateways are
+#    not used, or have been pre-configured to avoid CORS issues
+
```
n-riesco: Thanks for all the comments.

package.json
```diff
@@ -42,7 +42,11 @@
     "lodash": "^4.12.0",
     "transformime": "^3.0.0",
     "transformime-marked": "0.0.1",
-    "uuid": "^2.0.1"
+    "uuid": "^2.0.1",
+    "jupyter-js-services": "^0.11.4",
+    "requirejs": "^2.2.0",
+    "ws": "^1.1.1",
+    "xmlhttprequest": "^1.8.0"
```
n-riesco: Keep in alphabetical order, to avoid duplicates.

n-riesco:
@rgbkrk A question of nomenclature.

In this PR we're calling `GatewayKernel` the kernel class that makes use of `jupyter-js-services`. I've also noticed the proposal for a [Kernel Gateway](https://github.com/jupyter/kernel_gateway). Are they related?

How do you think? Is it OK to name `GatewayKernel` the kernels created using `jupyter-js-services`? Is there a better name?

lib/gateway-kernel.coffee
```diff
@@ -0,0 +1,94 @@
...
+path = require 'path'
+
+_ = require 'lodash'
+uuid = require 'uuid'
+services = require('./jupyter-js-services-shim')
+
+Kernel = require './kernel'
+
+class GatewayKernel extends Kernel
+    constructor: (kernelSpec, session) ->
```
n-riesco: `constructor: (kernelSpec, @session) ->`

lib/gateway-kernel.coffee
```diff
@@ -0,0 +1,94 @@
...
+        super kernelSpec
+
+        @session = session
+        @session.statusChanged.connect(@_onStatusChange, this)
```
n-riesco: Would this work?
```coffeescript
@session.statusChanged.connect => @_onStatusChange()
```

lib/gateway-kernel.coffee
```diff
@@ -0,0 +1,94 @@
...
+
+        if onResults?
+            console.log('got msg')
+            console.log(message)
+            result = @_parseIOMessage(message)
+            if result?
+                onResults result
+
+
+    execute: (code, onResults) ->
```
n-riesco: The code in `ZMQKernel.execute` and `ZMQKernel.executeWatch` should be in the parent class `Kernel`.
And `ZMQKernel._execute` and `GatewayKernel._execute` should have their own implementations.

In this way, you don't need to do anything to support kernel watches.

lib/gateway-kernel.coffee
```diff
@@ -0,0 +1,94 @@
...
+            code: code
+            cursor_pos: cursor_pos
+            detail_level: 0
+        ).then((message) ->
+            onResults
+                data: message.content.data
+                found: message.content.found
+            )
+
+    addWatchCallback: (watchCallback) ->
```
n-riesco: No need for this method.

lib/gateway-kernel.coffee
```diff
@@ -0,0 +1,94 @@
...
+        console.log 'WARNING: Watches are not supported by remote kernels'
+        @watchCallbacks.push(watchCallback)
+
+    destroy: ->
+        super
```
n-riesco: Here it doesn't matter, but in most cases the call to `super` in the destructor is last.

lib/gateway-kernel.coffee
```diff
@@ -0,0 +1,94 @@
...
+        console.log 'destroying jupyter-js-services Session'
+        @session.dispose()
+
+module.exports =
+    gatewayKernelFromSession: (session, callback) ->
```
n-riesco: We could make do without this function.
Also since it's a function, and to be consistent with the rest of the code, I would name it `getGatewayKernelFromSession`.

n-riesco:
@nikitakit I haven't reviewed f955c822d149b8c1130cefc7988eebd918d8f561 and d041a910bf9f5d76be50615071abedfc211d8ff6 yet. I will continue tomorrow.

rgbkrk:
> In this PR we're calling GatewayKernel the kernel class that makes use of jupyter-js-services. I've also noticed the proposal for a Kernel Gateway. Are they related?

> How do you think? Is it OK to name GatewayKernel the kernels created using jupyter-js-services? Is there a better name?

This is related. I'd probably call them `JupyterWebsocketKernel` (or `WSKernel` if you want something shorter) as the remote kernel setup done here works interchangeably with the kernel gateway and the standalone notebook.

lib/gateway-kernel.coffee
nikitakit: I don't see how this makes sense. `execute` (without the underscore) creates a request ID, which is ZMQ-specific. The gateway-based implementation doesn't use request ID's (or rather, all request ID's are managed internally to `jupyter-js-services`, with all outside access using higher-level APIs)

lib/gateway-kernel.coffee
nikitakit: I considered not having this function, but then there will be calls to the `jupyter-js-services` API directly from `KernelManager`. I decided it was a better idea to keep any calls to the external APIs confined to just a few files, instead of spread out throughout the codebase.

I'll rename it.

nikitakit:
`WSKernel` sounds good. I'll rename it next rebase.

I'm slightly busier these days, so I'll do a rebase once comments are in for the full patch series. (It will also be easier for me to resolve conflicts with new commits all in one sitting, instead of spread out over multiple rebases)

lib/zmq-kernel.coffee
nikitakit: Nice catch.

The correct fix is to not set it in the parent. (Execution callbacks are ZMQ-specific, because `jupyter-js-services` does its own callback management via promises)

lib/gateway-kernel.coffee
nikitakit: I'll double-check, but I remember looking into this when writing the code.

Apparently the second argument (the binding for this) is required by the signalling API being used. If `this` is not passed I'm pretty sure it will cause an error to be thrown.

n-riesco: On 21/07/16 14:40, Nikita Kitaev wrote:
> Apparently the second argument (the binding for this) is required by the signalling API being used. If |this| is not passed I'm pretty sure it will cause an error to be thrown.

This doesn't sound right.

In Javascript, if a callback requires `this` to be set by the caller. This is usually achieved by binding the function.

In Coffeescript, this is achieved by fat arrows (=>). Unbound functions are defined using (->).

n-riesco:
On 21/07/16 14:35, Nikita Kitaev wrote:
> I'm slightly busier these days, so I'll do a rebase once comments are in for the full patch series. (It will also be easier for me to resolve conflicts with new commits all in one sitting, instead of spread out over multiple rebases)

No worries. I'll try to review the remaining commits today.

There are potentially two PRs touching the same code: #355 and #342 .

Give us a shout before rebasing and I'll update you with the progress on those PRs.

lib/gateway-kernel.coffee
n-riesco: @rgbkrk Kyle, is this correct? Does `jupyter-js-services` remove the user's ability to set `msg_id`? Isn't `msg_id` part of the [Jupyter Messaging Protocol](http://jupyter-client.readthedocs.io/en/latest/messaging.html#general-message-format)?

lib/gateway-kernel.coffee
n-riesco: I don't follow the reasoning here.
Don't we need to use `jupyter-js-services` to get an instance of `Session`?

nikitakit: The instance of `Session` is created inside `session-picker.coffee`, but it needs to be converted into a `GatewayKernel` somewhere down the line.

I decided that the `GatewayKernel` constructor should be called from `KernelManager`. But constructors must be sync while `getKernelSpec` is async (and needs to happen before the Kernel is considered fully constructed). Hence this method, which is essentially an async constructor.

If you want to get rid of this method, where do you suggest inlining this code? If it's inlined into `KernelManager`, suddenly there will be lines of code in `KernelManager` that can't be understood without reading the `jupyter-js-services` API docs. I chose to limit `jupyter-js-services` API access to just two files (`session-picker.coffee` and `gateway-kernel.coffee`), instead of spread out all over the place.

lib/gateway-kernel.coffee
nikitakit: I checked again, and the code you suggested works.

Apparently the `TypeError`s I was getting before were unrelated to omitting the second argument.

lib/gateway-kernel.coffee
n-riesco: Could it go into `session-picker.coffee`? If `KernelManager` needs both `session` and `kernelSpec` to create a `GatewayKernel`, it'd be more convenient that `session-picker.coffee` returns both.

lib/kernel-manager.coffee
```diff
@@ -77,6 +78,13 @@ class KernelManager
                 )
                 finishKernelStartup kernel
 
+    attachToGatewaySession: (grammar, session, onAttach) ->
```
n-riesco: Rename to `startWSKernelFor` (to be consistent with the existing naming of `startKernel` and `startKernelFor`).

lib/main.coffee
```diff
@@ -356,6 +358,15 @@ module.exports = Hydrogen =
                     @watchSidebar.show()
         @watchKernelPicker.toggle()
 
+    showSessionPicker: ->
```
n-riesco: Hydrogen doesn't really need to know about sessions.

We want users to select kernels (not just sessions).
Rather than a `showSessionPicker`, we need a `showWSKernelPicker`.
This change will help get rid of `gatewayKernelFromSession`.

lib/main.coffee
```diff
@@ -49,6 +50,7 @@ module.exports = Hydrogen =
             'hydrogen:toggle-watches': => @toggleWatchSidebar()
             'hydrogen:select-watch-kernel': => @showWatchKernelPicker()
             'hydrogen:select-kernel': => @showKernelPicker()
+            'hydrogen:connect-to-remote-session': => @showSessionPicker()
```
n-riesco: `hydrogen:connect-to-remote-kernel`

lib/session-picker.coffee
```diff
@@ -0,0 +1,112 @@
+{SelectListView} = require 'atom-space-pen-views'
```
n-riesco: As I commented above, I think the user should choose kernels rather than sessions.

The current workflow is that users choose a ZMQ kernel amongst those that are able to handle the current grammar.

This PR should aim at the same workflow for WS kernels, i.e. the user should be shown a list of WS kernels that are able to handle the current grammar.

lib/signal-list-view.coffee
```diff
@@ -58,18 +58,21 @@ class SignalListView extends SelectListView
...
-                return {
-                    name: 'Switch to ' + spec.display_name
-                    value: 'switch-kernel'
-                    grammar: grammar
-                    language: language
-                    kernelSpec: spec
-                }
-
-            @setItems _.union basicCommands, switchCommands
+        if kernel.isRemote
```
n-riesco: `if kernel instanceof GatewayKernel`

nikitakit:
I have some time today, so I'm starting another rebase. I'll be checking off the comments with thumbs up emoji as I address them.

n-riesco:
On 24/07/16 16:28, Nikita Kitaev wrote:
> I have some time today, so I'm starting another rebase. I'll be checking off the comments with thumbs up emoji as I address them.

OK! I'm already online.

nikitakit:
Done with rebase.

I notice that ZMQ kernels now support prompting for input on stdin, but I'd like to consider the same functionality for remote kernels as out-of-scope for this PR. Rebases are starting to get unwieldy for such a large change, so I'd rather merge what's here first and work towards feature parity in future commits.

package.json
```diff
@@ -39,10 +40,14 @@
     "atom-space-pen-views": "^2.0.5",
     "escape-string-regexp": "^1.0.5",
     "jmp": "0.4.x",
+    "jupyter-js-services": "^0.11.4",
```
lgeiger: Is there a specific reason why you aren't using the latest version `0.16.3`?

nikitakit: Not really. I think I copied the dependencies from somewhere (probably either the example code of the atom notebook), and I don't actually know package.json syntax.

npm is pretty bandwidth-hungry, so I'd rather not be downloading alternative versions with my current (poor) internet connection.

lgeiger: No worries. I'm just curios. I can upgrade it in a future PR.

lib/zmq-kernel.coffee
```diff
@@ -0,0 +1,383 @@
...
+        super kernelSpec
+
+        @executionCallbacks = {}
+
+        projectPath = path.dirname(
+            atom.workspace.getActiveTextEditor().getPath()
+        )
+
+        @_connect()
+        if onlyConnect
```
n-riesco: `if @onlyConnect`

lib/ws-kernel.coffee
```diff
@@ -0,0 +1,84 @@
...
+        @statusView.setStatus @session.status
+
+    _onIOPub: (message, onResults) ->
+        msg_type = message.header.msg_type
+
+        # TODO(nikita): implement support for watches
+
+        if onResults?
+            console.log('got msg')
+            console.log(message)
```
n-riesco: console.log 'WSKernel: _onIOPub:', message

lib/ws-kernel.coffee
```diff
@@ -0,0 +1,84 @@
...
+            result = @_parseIOMessage(message)
+            if result?
+                onResults result
```
n-riesco: `onResults?(result)`

lib/ws-kernel.coffee
```diff
@@ -0,0 +1,84 @@
...
+            @_onIOPub(message, onResults)
+
+        future.onReply = (message) ->
+            if message.content.status is 'error'
+                return
+            result =
+                data: 'ok'
+                type: 'text'
+                stream: 'status'
+            onResults(result)
```
n-riesco: `onResults?(result)`

lib/ws-kernel.coffee
```diff
@@ -0,0 +1,84 @@
...
+
+    executeWatch: (code, onResults) ->
+        # TODO(nikita): implement watches
+        # Also consider switching to user_expressions for watches
+        console.log 'WARNING: Watches are not supported by remote kernels'
```
n-riesco: Use atom notifications.
---
Not every kernel supports user_expressions.

lib/ws-kernel.coffee
```diff
@@ -0,0 +1,84 @@
...
+
+    complete: (code, onResults) ->
+        @session.kernel.complete(
+            code: code
+            cursor_pos: code.length
+        ).then((message) ->
+            onResults message.content
+            )
+
```
n-riesco: no needs for parens:
```coffeescript
        @session.kernel.complete
            code: code
            cursor_pos: code.length
        .then (message) ->
            onResults message.content
```

lib/ws-kernel.coffee
```diff
@@ -0,0 +1,84 @@
...
+        # TODO(nikita): implement watches
+        # Also consider switching to user_expressions for watches
+        console.log 'WARNING: Watches are not supported by remote kernels'
+
+    complete: (code, onResults) ->
+        @session.kernel.complete(
+            code: code
+            cursor_pos: code.length
+        ).then((message) ->
+            onResults message.content
```
n-riesco: `onResults?(message.content)`

lib/ws-kernel.coffee
```diff
@@ -0,0 +1,84 @@
...
+            )
+
+    inspect: (code, cursor_pos, onResults) ->
+        @session.kernel.inspect(
+            code: code
+            cursor_pos: cursor_pos
+            detail_level: 0
+        ).then((message) ->
+            onResults
```
n-riesco:
```coffeescript
            onResults?(
                data: message.content.data
                found: message.content.found
            )
```

lib/ws-kernel.coffee
```diff
@@ -0,0 +1,84 @@
...
+        @session.kernel.complete(
+            code: code
+            cursor_pos: code.length
+        ).then((message) ->
+            onResults message.content
+            )
+
+    inspect: (code, cursor_pos, onResults) ->
+        @session.kernel.inspect(
+            code: code
```
n-riesco: no need for parens

lib/ws-kernel.coffee
```diff
@@ -0,0 +1,84 @@
...
+            cursor_pos: cursor_pos
+            detail_level: 0
+        ).then((message) ->
```
n-riesco: no need for parens

lib/signal-list-view.coffee
```diff
@@ -58,18 +60,21 @@ class SignalListView extends SelectListView
...
-                return {
-                    name: 'Switch to ' + spec.display_name
-                    value: 'switch-kernel'
-                    grammar: grammar
-                    language: language
-                    kernelSpec: spec
-                }
-
-            @setItems _.union basicCommands, switchCommands
+        if kernel instanceof WSKernel
```
n-riesco: Why isn't it possible to switch from remote kernels to other kernels?

n-riesco:
The description of commit 560bc13f is a bit deceptive. If I understand correctly, this commit does to things:
- handle restart of remote kernels
- disable switching from remote kernels

lib/kernel-manager.coffee
```diff
@@ -77,6 +77,14 @@ class KernelManager
                 )
                 finishKernelStartup kernel
 
+    attachKernel: (grammar, kernel, onAttach) ->
```
n-riesco: As it stands there is no need for the `onAttach` callback (because this method isn't asynchronous).

lib/ws-kernel-picker.coffee
```diff
@@ -0,0 +1,118 @@
...
+    cancelled: ->
+        @panel?.destroy()
+        @panel = null
+        @editor = null
+
+    confirmed: (item) ->
+        @onConfirmed?(item)
+        @cancel()
+
+    getEmptyMessage: ->
```
n-riesco: This method is unused.

lib/ws-kernel-picker.coffee
```diff
@@ -0,0 +1,118 @@
...
+
+    viewForItem: (item) ->
+        element = document.createElement('li')
+        element.textContent = item.name
+        element
+
+    cancelled: ->
+        @panel?.destroy()
+        @panel = null
+        @editor = null
```
n-riesco: @editor is unsed

lib/ws-kernel-picker.coffee
```diff
@@ -0,0 +1,118 @@
...
+            gatewayListing.setError 'Use the package settings to specify gateway information'
+
+    onGateway: (spec) ->
+        console.log('Picked a gateway')
+        sessionListing = new CustomListView('No sessions available', @onSession.bind this)
+        sessionListing.setLoading 'Loading sessions...'
+
+        services.listRunningSessions(spec.options)
+        .then (sessionModels) ->
+            items = sessionModels.map (model) ->
```
n-riesco: A question for a future PR, is it possible to filter the sessions to ensure that only sessions able to run the current grammar are shown?

lib/ws-kernel-picker.coffee
```diff
@@ -0,0 +1,118 @@
...
+            items.unshift
+                name: '[new session]'
+                model: null
+                options: spec.options
+            sessionListing.setItems(items)
+
+        , (err) =>
+            # Gateways offer the option of never listing sessions, for security
+            # reasons.
+            # Assume this is the case and proceed to creating a new session.
```
n-riesco: Would it be possible to offer the possibility to create a new session?

lib/ws-kernel.coffee
nikitakit: I see about user_expressions

In terms of atom notifications, this isn't the right place (IIRC this method isn't even called anywhere, because watches are also disabled in several other places, including the one which causes executeWatch would be called). I'll figure out the entire business with watches in a future PR.

lib/ws-kernel.coffee
nikitakit: Note that this is inside an `if onResults?` block. Is double-checking the accepted coding style here?

lib/ws-kernel-picker.coffee
nikitakit: This is an override that's used by the parent class.

lib/ws-kernel-picker.coffee
nikitakit: I'm confused about your question here, because the possibility of creating a new session exists in this very file!

lib/ws-kernel-picker.coffee
nikitakit: Filtering by grammar isn't too hard, but will need some careful attention in a future PR.

n-riesco:
@nikitakit A question about the last commit 2b98a7917a35632d6d6660a7f62af7949b4e8d61 . Let's imagine we've started a session with a remote kernel using `hydrogen:connect-to-remote-kernel`. Then, we invoke `hydrogen:connect-to-remote-kernel` again to select another remote kernel. What happens to the previous kernel?

lib/ws-kernel-picker.coffee
n-riesco: Maybe I'm reading the code incorrectly, what does the following line do?

```coffeescript
            @onSession
                name: null
                model: null
                options: spec.options
```

nikitakit:
@n-riesco Regarding connecting again:

If you select `hydrogen:connect-to-remote-kernel` a second time, you will be offered the option to connect to the same kernel (again) or to connect to another one. If you select another kernel, the first one remains on standby until it is needed again.

This follows from the guiding principle of never accidentally killing remote kernels. Currently, the only way to kill a remote kernel is to select "restart kernel" or to shut down the gateway server. (In the future, I'll investigate "switch kernel" functionality, and adding a "clean shutdown" option to the signal switcher view)

I use multiple kernels on standby all the time at work (they each run the same code, but with different parameters specified)

lib/signal-list-view.coffee
nikitakit: This is tricky to implement, so I left it for the future. The challenges are:
* Offering a selection of kernelspecs from the same gateway (there could be multiple gateways)
* Deciding whether to offer any local kernelspec options (if so, the UI must clearly distinguish them from any remote options)
* Cleanly shutting down the old kernel ("switch kernel" should kill the old kernel, by design)

lib/ws-kernel-picker.coffee
nikitakit: They delegate to the `unless` clause of `onSession`, which is responsible for offering the user a choice of kernelspecs, and then starting the chosen kernel. The result of selecting the "[new session]" option are the same.

(model is a variable that contains information about an existing session. It is set to null to indicate that a new session/kernel should be created instead)

n-riesco:
@nikitakit I understand the wish to keep remote kernels on standby. My concern is whether we are leaking instances of `WSKernel` and what that means to the gateway?

lib/ws-kernel-picker.coffee
n-riesco: Shouldn't it be `name: '[new session]'` in this case too?

nikitakit:
@n-riesco Good catch about leaking kernel instances. I'll push a fix shortly (it will add `destroyRunningKernel` and `clearResultBubbles` around `main.coffee:385`.)

As for the gateway, it has its own resource management. Notebooks have a "running kernels" panel UI that allows killing kernels, and I'll eventually add an option like this into the hydrogen signal list menu. Moreover, the jupyter team is implementing functionality to automatically kill kernels that are inactive longer than a certain threshold.

lib/ws-kernel-picker.coffee
```diff
@@ -0,0 +1,117 @@
...
+Config = require './config'
+services = require './jupyter-js-services-shim'
+WSKernel = require './ws-kernel'
+uuid = require 'uuid'
+
+class CustomListView extends SelectListView
+    initialize: (@emptyMessage, @onConfirmed) ->
+        super
+        @addClass('watch-language-picker')
+        @storeFocusedElement()
```
lgeiger: Restoring the focused pane isn't working for me. I think it has something to do with the location where `@storeFocusedElement()` is called.

lib/ws-kernel-picker.coffee
```diff
@@ -0,0 +1,117 @@
...
+
+Config = require './config'
+services = require './jupyter-js-services-shim'
+WSKernel = require './ws-kernel'
+uuid = require 'uuid'
+
+class CustomListView extends SelectListView
+    initialize: (@emptyMessage, @onConfirmed) ->
+        super
+        @addClass('watch-language-picker')
```
lgeiger: This line is unnecessary since hydrogen doesn't use the `watch-language-picker` class anymore.

nikitakit: You're right, I got rid of it.

lib/ws-kernel-picker.coffee
nikitakit: Thanks for pointing this out.

I'm not sufficiently familiar with atom to know a fix offhand, so I've made a note to look into this in the future (i.e. another PR).

lgeiger: I had a quick look into it:
The problem is the `CustomListView`s created in `onGateway` and `onSession` try to restore the focus to the input field of `CustomListView`.

I have an easy fix for that and submit a PR after this is merged.

lgeiger:
@nikitakit Thanks a lot for your work. This is an awesome addition to hydrogen.

rgbkrk:
Woo!

PR #332, opened 2016-07-09
Add select-kernel to list of activation commands

n-riesco:
I've noticed that 0a2bd2e also fixes a bug triggered when `hydrogen:select-kernel` was invoked for a file that had no running kernel.


lgeiger:
I wanted to propose this a while ago but somehow forgot. Thanks!

PR #333, opened 2016-07-09
CellManager: Fix regexp and off-by-one bug

n-riesco:


lgeiger:
I counted the row with the cell comment to the cell below on purpose. I think this is more intuitive but it would break [in this case](https://github.com/nteract/hydrogen/pull/328#discussion_r70044412).

I think this is a good way to prevent this issue.

PR #335, opened 2016-07-10
Refactor KernelManager as a class

n-riesco:
Originally, `KernelManager` was implemented using a global plain object.

This PR refactors `KernelManager` as a class so that `coffeelint` can check the use of arrows more reliably.

This PR replaces the use of the global object `KernelManager` by an instance of `KernelManager`, which is passed to wherever this object was used. This means that:
- the function that creates `AutocompleteProvider` now takes an instance of `KernelManager` as an argument;
- `SignalListView` takes an instance of `KernelManager` in the constructor;
- `Inspector` had to be converted into a class with a constructor that takes an instance of `KernelManager`;
- ...

I've also taken this opportunity to simplify the naming in `KernelManager`. Now there is only one type of language: that obtained using `KernelManager::getLanguageFor(grammar)`.

I've also fixed a  bug in `hydrogen:select-watch-kernel`.

I've tested this PR running simultaneously multiple kernels and watches and haven't found any problems.

---

TODO for  another PR: 
- There are two files, `main.coffee` and `autocomplete-provide.coffee` that still use '@' with plain objects.
- Update spec to use an instance of `KernelManager` instead of the `KernelManager::` hack.

### DIFF
--- a/lib/autocomplete-provider.coffee
+++ b/lib/autocomplete-provider.coffee
@@ -50,8 +50,8 @@ module.exports = (kernelManager) ->
                 return null
 
             grammar = editor.getGrammar()
-            grammarLanguage = kernelManager.getGrammarLanguageFor grammar
-            kernel = kernelManager.getRunningKernelFor grammarLanguage
+            language = kernelManager.getLanguageFor grammar
+            kernel = kernelManager.getRunningKernelFor language
             unless kernel?
                 return null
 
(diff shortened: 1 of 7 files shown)


lgeiger:
@n-riesco Thanks for refactoring KernelManager! This is a lot better now.

I tested this PR and think we can merge this.

PR #337, opened 2016-07-10
Don't save kernelspecs to settings

lgeiger:
As discussed in #272.

This PR updates specs to use an instance of `KernelManager` as introduced in #335.

### DIFF
--- a/lib/kernel-manager.coffee
+++ b/lib/kernel-manager.coffee
@@ -165,6 +165,7 @@ class KernelManager
                     dismissable: true
                 atom.notifications.addError message, options
             else
+                err = null
                 message = 'Hydrogen Kernels updated:'
                 options =
                     detail: (_.map @_kernelSpecs, 'spec.display_name').join('\n')
--- a/lib/kernel-manager.coffee
+++ b/lib/kernel-manager.coffee
@@ -11,6 +11,7 @@ module.exports =
 class KernelManager
     constructor: ->
         @_runningKernels = {}
+        @_kernelSpecs = []
 
 
     destroy: ->
@@ -91,8 +92,7 @@ class KernelManager
 
 
     getAllKernelSpecs: ->
-        kernelSpecs = _.map @parseKernelSpecSettings(), 'spec'
-        return kernelSpecs
+        return _.clone @_kernelSpecs
 
 
     getAllKernelSpecsFor: (language) ->
@@ -162,10 +162,10 @@ class KernelManager
         kernelSpecs = @parseKernelSpecSettings()
         _.assign kernelSpecs, newKernelSpecs
 
-        Config.setJson 'kernelspec', kernelspecs: kernelSpecs
+        @_kernelSpecs = _.map kernelSpecs, 'spec'
 
         message = 'Hydrogen Kernels updated:'
-        options = detail: (_.map kernelSpecs, 'spec.display_name').join('\n')
+        options = detail: (_.map @_kernelSpecs, 'display_name').join('\n')
         atom.notifications.addInfo message, options
 
 
--- a/lib/kernel-manager.coffee
+++ b/lib/kernel-manager.coffee
@@ -24,10 +24,15 @@ class KernelManager
 
 
     startKernelFor: (grammar, onStarted) ->
+        if _.isEmpty @_kernelSpecs
+            @updateKernelSpecs =>
+                @_startKernelFor grammar, onStarted
+        else
+            @_startKernelFor grammar, onStarted
+
+
+    _startKernelFor: (grammar, onStarted) ->
         language = @getLanguageFor grammar
-
-        console.log 'startKernelFor:', language
-
         kernelSpec = @getKernelSpecFor language
 
         unless kernelSpec?
@@ -38,6 +43,7 @@ class KernelManager
             atom.notifications.addError message, options
             return
 
+        console.log 'startKernelFor:', language
         @startKernel kernelSpec, grammar, onStarted
 
 
--- a/lib/kernel-manager.coffee
+++ b/lib/kernel-manager.coffee
@@ -11,7 +11,7 @@ module.exports =
 class KernelManager
     constructor: ->
         @_runningKernels = {}
-        @_kernelSpecs = []
+        @_kernelSpecs = @getKernelSpecsFromSettings()
 
 
     destroy: ->
@@ -92,7 +92,7 @@ class KernelManager
 
 
     getAllKernelSpecs: ->
-        return _.clone @_kernelSpecs
+        return _.map @_kernelSpecs, 'spec'
 
 
     getAllKernelSpecsFor: (language) ->
@@ -129,7 +129,7 @@ class KernelManager
         return kernelLanguage.toLowerCase() is language
 
 
-    parseKernelSpecSettings: ->
+    getKernelSpecsFromSettings: ->
         settings = Config.getJson 'kernelspec'
 
         unless settings.kernelspecs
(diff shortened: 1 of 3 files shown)


lib/kernel-manager.coffee
```diff
@@ -91,8 +92,7 @@ class KernelManager
 
 
     getAllKernelSpecs: ->
-        kernelSpecs = _.map @parseKernelSpecSettings(), 'spec'
-        return kernelSpecs
+        return @_kernelSpecs
```
n-riesco: We should return a clone, to avoid side-effects. See [getAllRunningKernels](https://github.com/lgeiger/hydrogen/blob/fd07870dce47987d46ea82ea1266d9cba85d4ae3/lib/kernel-manager.coffee#L83)

lib/kernel-manager.coffee
```diff
@@ -11,6 +11,7 @@ module.exports =
 class KernelManager
     constructor: ->
         @_runningKernels = {}
+        @_kernelSpecs = []
```
n-riesco: I think `@_kernelSpecs` should be initialised with the user settings; i.e. `@_kernelSpecs = @parseKernelSpecSettings()`. Is there a case when we wouldn't this to happen?

n-riesco:
We don't want to update the kernel specs synchronously, because it may be too slow (IPython introduced a 5-second deprecation delay in `ipython notebook`).

spec/hydrogen-spec.coffee
```diff
@@ -106,94 +106,86 @@ describe 'Kernel manager', ->
...
             expect(kernelSpecForJulia).toBeUndefined()
 
     it 'should read lower case name from grammar', ->
         grammar = atom.grammars.getGrammars()[0]
-        expect(KernelManager::getLanguageFor grammar).toEqual('null grammar')
+        expect(@kernelManager.getLanguageFor grammar).toEqual('null grammar')
 
     it 'should update kernelspecs', ->
-        KernelManager::updateKernelSpecs()
+        @kernelManager.updateKernelSpecs()
```
n-riesco: I think  we should add a callback to `updateKernelSpecs` rather than making it synchronous.
:) Javascript is all about asynchronous code!

lib/kernel-manager.coffee
lgeiger: Thanks a lot for your review! I really appreciate your feedback.

> @n-riesco I think @_kernelSpecs should be initialised with the user settings; i.e. @_kernelSpecs = @parseKernelSpecSettings(). Is there a case when we wouldn't this to happen?

The parsing of user settings should happen in `updateKernelSpecs`, otherwise `hydrogen:update-kernels` wouldn't help if the settings are changed by the user.

lgeiger:
> We don't want to update the kernel specs synchronously, because it may be too slow (IPython introduced a 5-second deprecation delay in ipython notebook).

Thats why we try `jupyter kernelspec list --json` beforehand.

I opted for synchronous code because hydrogen has to wait anyway for `updateKernelSpecs` before any code can be executed. The asynchronous code before was fine because the specs where cached in the settings.

lib/kernel-manager.coffee
n-riesco: On 11/07/16 10:55, Lukas Geiger wrote:
>>    @n-riesco <https://github.com/n-riesco> I think @_kernelSpecs should be initialised with the user settings; i.e. @_kernelSpecs = @parseKernelSpecSettings(). Is there a case when we wouldn't this to happen?
> 
> The parsing of user settings should happen in |updateKernelSpecs|, otherwise |hydrogen:update-kernels| wouldn't help if the settings are changed by the user.

One thing doesn't preclude the other, but it makes me think that `@updateKernelSpecs` could be split into two methods `@getKernelSpecsFromSettings` and `@getKernelSpecsFromJupyter`.

The constructor would call `@getKernelSpecsFromSettings`.
And `@updateKernelSpecs` woudl call `@getKernelSpecsFromSettings` and `@getKernelSpecsFromJupyter`.

lib/kernel-manager.coffee
lgeiger: Fixed :+1:

n-riesco:
On 11/07/16 11:01, Lukas Geiger wrote:
>     We don't want to update the kernel specs synchronously, because it may be too slow (IPython introduced a 5-second deprecation delay in ipython notebook).
> 
> Thats why we try |jupyter kernelspec list --json| beforehand.
> 
> I opted for synchronous code because hydrogen has to wait anyway for |updateKernelSpecs| before any code can be executed. The asynchronous code before was fine because the specs where cached in the settings.

OK, I chose a bad counter-example.

As a rule of thumb, I wouldn't use `execSync` in Atom because it is potentially slow (a terminal has to be opened to run it; even commands that do nothing but fail are slow to run; in some OS like Ubuntu, when a command is not found, the repository database is searched to advise the user what package to install, this search is very slow; ...).

`execSync` is a convenience function when Node.js is used to write scripts, but it goes against Javascript's philosophy of running small asynchronous tasks (the event loop).

I think the strongest argument against using `execSync` in this case is that some users cannot run `jupyter kernelspec`, and they shouldn't be penalised for that.

lib/kernel-manager.coffee
lgeiger: I don't see why initialising `@_kernelSpecs = @parseKernelSpecSettings()` would be an improvement. Since on activation of Hydrogen `@updateKernelSpecs` is called anyway.

> One thing doesn't preclude the other, but it makes me think that `@updateKernelSpecs` could be split into two methods `@getKernelSpecsFromSettings` and `@getKernelSpecsFromJupyter`.

This is actually the case right now. The only difference are the method names:
`@updateKernelSpecs` is now called `@saveKernelSpecs`,
`@getKernelSpecsFromJupyter` is `@updateKernelSpecs` and
`@getKernelSpecsFromSettings` is `@parseKernelSpecSettings`.

We can rename those methods if you think this would make it clearer.

lgeiger:
I totally understand your point. But if you try it async (checkout [`ffb242d`](https://github.com/nteract/hydrogen/pull/337/commits/ffb242d926f138b5094429f68852d32631a9502d)) hydrogen would throw an error on every startup since no kernelspec is available. Hence we have to always wait for `updateKernelSpecs` to finish.

lib/kernel-manager.coffee
n-riesco: On 11/07/16 12:59, Lukas Geiger wrote:
> In lib/kernel-manager.coffee <https://github.com/nteract/hydrogen/pull/337#discussion_r70244433>:
> 
>> @@ -11,6 +11,7 @@ module.exports =
>>  class KernelManager
>>      constructor: ->
>>          @_runningKernels = {}
>> +        @_kernelSpecs = []
> 
> I don't see why initialising |@_kernelSpecs = @parseKernelSpecSettings()| would be an improvement. Since on activation of Hydrogen |@updateKernelSpecs| is called anyway.

It's an improvement for users like me that cannot run `ipython kerlnelspec`.

It means I can start Hydrogen without waiting for `jupyter kernelspec` and `ipython kernelspec` to fail (which is the current behaviour).

> 
>     One thing doesn't preclude the other, but it makes me think that |@updateKernelSpecs| could be split into two methods |@getKernelSpecsFromSettings| and |@getKernelSpecsFromJupyter|.
> 
> This is actually the case right now. The only difference are the method names:
> |@updateKernelSpecs| is now called |@saveKernelSpecs|,
> |@getKernelSpecsFromJupyter| is |@updateKernelSpecs| and
> |@getKernelSpecsFromSettings| is |@parseKernelSpecSettings|.
> 
> We can rename those methods if you think this would make it clearer.

It makes sense to me. How do you think?

n-riesco:
On 11/07/16 13:53, Lukas Geiger wrote:
> I totally understand your point. But if you try it async (checkout |ffb242d| <https://github.com/nteract/hydrogen/pull/337/commits/ffb242d926f138b5094429f68852d32631a9502d>) hydrogen would throw an error on every startup since no kernelspec is available. Hence we have to always wait for |updateKernelSpecs| to finish.

This is the reason why I think `updateKernelSpecs` should accept a callback.

Currently, one of the specs fails in machines where `jupyter` isn't available:

```
$ apm test
[30876:0711/141009:ERROR:browser_main_loop.cc(203)] Running without the SUID sandbox! See https://code.google.com/p/chromium/wiki/LinuxSUIDSandboxDevelopment for more information on developing with the sandbox on.
[30876:0711/141011:INFO:audio_manager_pulse.cc(174)] Not implemented!
[30876:0711/141011:INFO:audio_manager_pulse.cc(174)] Not implemented!
[30876:0711/141012:INFO:CONSOLE(52)] "Window load time: 1321ms", source: file:///home/user/share/atom/resources/app.asar/static/index.js (52)
...........[30876:0711/141012:INFO:CONSOLE(162)] "saveKernelSpecs:", source: /home/user/src/github.com/hydrogen/lib/kernel-manager.coffee (162)
.[30876:0711/141012:INFO:CONSOLE(162)] "saveKernelSpecs:", source: /home/user/src/github.com/hydrogen/lib/kernel-manager.coffee (162)
.[30876:0711/141012:INFO:CONSOLE(162)] "saveKernelSpecs:", source: /home/user/src/github.com/hydrogen/lib/kernel-manager.coffee (162)
.[30876:0711/141012:INFO:CONSOLE(162)] "saveKernelSpecs:", source: /home/user/src/github.com/hydrogen/lib/kernel-manager.coffee (162)
....[30876:0711/141012:INFO:CONSOLE(167)] "[object Object]", source: /home/user/src/github.com/hydrogen/spec/hydrogen-spec.coffee (167)
...[30876:0711/141012:INFO:CONSOLE(198)] "updateKernelSpecs: `jupyter kernelspec` failed", source: /home/user/src/github.com/hydrogen/lib/kernel-manager.coffee (198)
[30876:0711/141012:INFO:CONSOLE(204)] "updateKernelSpecs: `ipython kernelspec` failed", source: /home/user/src/github.com/hydrogen/lib/kernel-manager.coffee (204)
F

Kernel manager
it should update kernelspecs
SyntaxError: Unexpected end of input
SyntaxError: Unexpected end of input
at Object.parse (native)
at [object Object].<anonymous> (/home/user/src/github.com/hydrogen/spec/hydrogen-spec.coffee:197:31)


Finished in 3.515 seconds
22 tests, 41 assertions, 1 failure, 0 skipped
```

n-riesco:
I'm so late!!! I'll test this tomorrow.

lgeiger:
OK I changed the code back to async.

Is there any way of preventing [this error](https://github.com/nteract/hydrogen/blob/master/lib/kernel-manager.coffee#L33) if the `updateKernelSpecs` command isn't finished? This problem was the reason why I originally went for `execSync`.

n-riesco:
The convention in Node.js for callbacks is something like this:

```coffeescript
myAsynchronousFunction = (arg1, arg2, argN, callback) ->
    # do something with arg1, arg2, argN
    callback? err, result1, result2, resultN
```

so that we can do:

```coffeescript
myAsynchronousFunction arg1, arg2, argN, (err, result1, result2, resultN) ->
    if err
        # handle error
        return
    # do something with result1, result2, resultN
```

Also note that if a function invokes an asynchronous function, then this function is also asynchronous and should accept a callback.

In this PR's case, we have `@updateKernelSpecs` that invokes  `@getKernelSpecsFromJupyter` and `@getKernelSpecsFromSettings`.

`@getKernelSpecsFromSettings` is synchronous and doesn't need a callback.

`@getKernelSpecsFromJupyter` is asynchronous and needs a callback. This also means `@updateKernelSpecs` is asynchronous and needs a callback.

All the above could be implemented like this:

```coffeescript
    mergeKernelSpecs: (kernelSpecs) ->
        _.assign @_kernelSpecs, kernelSpecs

    updateKernelSpecs: (callback) ->
        @getKernelSpecsFromJupyter (err, kernelSpecsFromJupyter) =>
            unless err
                @mergeKernelSpecs kernelSpecsFromJupyter

            kernelSpecsFromSettings = @getKernelSpecsFromSettings
            @mergeKernelSpecs kernelSpecsFromSettings

            if @_kernelSpecs.length is 0
                err = new Error 'updateKernelSpecs: no kernel specs found'

            callback? null, @_kernelSpecs

    getKernelSpecsFromJupyter: (callback) ->
        jupyter = 'jupyter kernelspec list --json --log-level=CRITICAL'
        ipython = 'ipython kernelspec list --json --log-level=CRITICAL'

        getKernelSpecsFrom jupyter, (jupyterError, kernelSpecs) ->
            unless jupyterError
                callback? jupyterError, kernelSpecs
                return

            getKernelSpecsFrom ipython, (ipythonError, kernelSpecs) ->
                unless ipythonError
                    callback? ipythonError, kernelSpecs
                else
                    callback? jupyterError, kernelSpecs

    getKernelSpecsFrom: (command, callback) ->
        child_process.exec command, (err, stdout, stderr) ->
            unless err
                kernelSpecsOrErr = JSON.parse(stdout).kernelspecs
                if kernelSpecsOrErr instanceof Error
                    err = kernelSpecsOrErr
                else
                    kernelSpecs = kernelSpecsOrErr

            callback? err, kernelSpecs

    parseJupyterStdout: (stdout) ->
        try
                kernelSpecsOrErr = JSON.parse(stdout).kernelspecs
        catch err
                kernelSpecsOrErr = err
                console.log 'parseJupyterStdout:', err
        return kernelSpecsOrErr
```

Using the above, the kernel spec could be implemented like this:

```coffeescript
    it 'should update kernelspecs', (done) ->
        @kernelManager.getKernelSpecsFromJupyter (err, specs) ->
            expect(specs).toEqual(jasmine.any(Object))
            done()
```

Warning: I haven't tested the code above.

lgeiger:
Thanks a lot for the great explanation this helps a lot! I'm still New to the concept oft async programming so bare with me ;).
I'll implement and test your code this evening.

lgeiger:
@n-riesco Thanks a lot for your help! I've updated the PR.

In [d90cef9](https://github.com/nteract/hydrogen/pull/337/commits/d90cef9ab88ff0102b02539ce7be93eccbd4d386) the updating of kernelspecs was moved to kernel startup, in order to prevent the `language not found` error on startup.

lib/kernel-manager.coffee
```diff
@@ -164,7 +170,7 @@ class KernelManager
                     detail: (_.map @_kernelSpecs, 'spec.display_name').join('\n')
                 atom.notifications.addInfo message, options
 
-            callback? err, @_kernelSpecs
+            callback? null, @_kernelSpecs
```
n-riesco: I don't understand this.

Shouldn't `updateKernelSpecs` fail if `err` in `@getKernelSpecsFromJupyter` and `@getKernelSpecsFromSettings` is empty?

lgeiger: Good catch, I didn't change that on purpose.

lib/kernel-manager.coffee
```diff
@@ -140,55 +146,59 @@ class KernelManager
...
+
+            if _.isEmpty @_kernelSpecs
+                message = 'No kernel specs found'
+                options =
+                    detail: 'Use kernelSpec option in Hydrogen or update
+                    IPython/Jupyter to a version that supports: `jupyter
+                    kernelspec list --json` or `ipython kernelspec list --json`'
+                    dismissable: true
+                atom.notifications.addError message, options
+            else
```
n-riesco: I would add `err = null` here so that an error will be returned only if `@_kernelSpecs` is empty

n-riesco:
@lgeiger Kudos! A very nice PR!

lgeiger:
@n-riesco Thanks for you awesome review. I learned a lot 🎉

PR #338, opened 2016-07-10
Close Inspector if invoked a second time on the same position

lgeiger:
- This PR removes the  `hydrogen:toggle-inspector-size` and `hydrogen:close-inspector` commands because I think they aren't necessary anymore.

- `hydrogen:inspect` was renamed to `hydrogen:toggle-inspector`

### DIFF
--- a/lib/inspector.coffee
+++ b/lib/inspector.coffee
@@ -7,14 +7,14 @@ class Inspector
         @editor = atom.workspace.getActiveTextEditor()
         @_lastInspectionResult = ''
 
-    inspect: ->
+    toggle: ->
         @editor = atom.workspace.getActiveTextEditor()
         grammar = @editor.getGrammar()
         language = @kernelManager.getLanguageFor grammar
         kernel = @kernelManager.getRunningKernelFor language
         unless kernel?
             atom.notifications.addInfo 'No kernel running!'
-            @closeInspector()
+            @view?.close()
             return
 
         [code, cursor_pos] = @getCodeToInspect()
@@ -28,10 +28,14 @@ class Inspector
                     firstline = lines[0]
                     lines.splice(0, 1)
                     message = lines.join('\n')
-                    if @_lastInspectionResult is message
-                        @closeInspector()
+
+                    @view ?= new MessagePanelView
+                        title: 'Hydrogen Inspector'
+                        closeMethod: 'destroy'
+
+                    if @_lastInspectionResult is message and @view.panel?
+                        @view?.close()
                     else
-                        @initializeInspector()
                         @addInspectResult(firstline, message)
                         @_lastInspectionResult = message
 
@@ -43,7 +47,7 @@ class Inspector
 
             else
                 atom.notifications.addInfo 'No introspection available!'
-                @closeInspector()
+                @view?.close()
 
     getCodeToInspect: ->
         if @editor.getSelectedText()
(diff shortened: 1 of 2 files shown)


lib/inspector.coffee
```diff
@@ -50,7 +56,7 @@ class Inspector
             cursor_pos = cursor.getBufferColumn()
         return [code, cursor_pos]
 
-    getInspector: ->
+    initializeInspector: ->
```
n-riesco: This method is only used by the constructor. Can it be moved to the constructor?
And I'd rename `@inspector` to `@view`.

lib/inspector.coffee
```diff
@@ -5,6 +5,7 @@ module.exports =
 class Inspector
     constructor: (@kernelManager) ->
         @editor = atom.workspace.getActiveTextEditor()
+        @_lastInspectionResult = ''
 
     inspect: ->
```
n-riesco: With this PR, it make sense to rename `@inspect` as `@toggle`.

lib/inspector.coffee
```diff
@@ -69,11 +75,8 @@ class Inspector
             className: 'inspect-message'
             raw: true
 
-    toggleInspectorSize: ->
-        if @inspector?
-            @inspector.toggle()
-
     closeInspector: ->
```
n-riesco: I'd remove this method or rename it to `@hide`.

lib/inspector.coffee
n-riesco: I see `initializeInspector` is now used to toggle the inspector view.
I'd still think we could rename `@inspector` to `@view` and move its creation to the constructor.
I'd also rename `@initializeInspector` to `@show`.

lgeiger: I'm OK with renaming `@inspector` to `@view`. But I don't think `@initializeInspector` should be renamed to `@show` since it just constructs or clears the view. The inspector is only shown if a message is added.

I didn't move the creation to the constructor because the `inspect` command is only used on occasions, hence I think the view should only be created on the first `inspect` request. Or am I missing something?

lib/inspector.coffee
lgeiger: I don't think we have to have a hydrogen command for hiding the view. This functionality is provided by an button on the view itself.

If you like to keep `hydrogen:toggle-inspector-size` and `hydrogen:close-inspector` I can add them back.

lib/inspector.coffee
n-riesco: On 11/07/16 11:40, Lukas Geiger wrote:
> In lib/inspector.coffee <https://github.com/nteract/hydrogen/pull/338#discussion_r70235799>:
> 
>> @@ -50,7 +56,7 @@ class Inspector
>>              cursor_pos = cursor.getBufferColumn()
>>          return [code, cursor_pos]
>>  
>> -    getInspector: ->
>> +    initializeInspector: ->
> 
> I'm OK with renaming |@inspector| to |@view|. But I don't think |@initializeInspector| should be renamed to |@show| since it just constructs or clears the view. The inspector is only shown if a message is added.
> 
> I didn't move the creation to the constructor because the |inspect| command is only used on occasions, hence I think the view should only be created on the first |inspect| request. Or am I missing something?

I find confusing that an `init` function is invoked multiple times.

I think the real problem is that `@initializeInspector` is used to accomplish two different goals: construct the view and clear the view.

Since the view is to be constructed on first call to `@toggle`, why not having a check in `@toggle` to create `@view`:

```
@view ?= new MessagePanelView, title: 'Hydrogen Inspector'
```

and replace the call to `@initializeInspector` with a `@view.clear()` ?

lib/inspector.coffee
n-riesco: On 11/07/16 11:43, Lukas Geiger wrote:
>>      closeInspector: ->
> 
> I don't think we have to have a hydrogen command for hiding the view. This functionality is provided by an button on the view itself.
> 
> If you like to keep |hydrogen:toggle-inspector-size| and |hydrogen:close-inspector| I can add them back.

To be clear, I'm only suggesting to rename the method `@closeInspector` to `@hide`.

As you said, having the command `hydrogen:toggle-inspector` makes |hydrogen:toggle-inspector-size| and |hydrogen:close-inspector| redundant.

lgeiger: Oh I see. I missunderstood that makes sense.

lgeiger:
I did tried to simplify the code. Thanks a lot for your feedback!

PR #339, opened 2016-07-10
SignalListView: Cleanup language names

lgeiger:

PR #340, opened 2016-07-10
Kernel: Remove hardcoded ipython kernel

### LINKED ISSUE #203: Kernel hardcode
Anybody knows is this lines actual?
https://github.com/nteract/hydrogen/blob/master/lib/kernel.coffee#L33-L45

This clobber should be removed.

lgeiger:
Fixes #203

### DIFF
--- a/lib/kernel.coffee
+++ b/lib/kernel.coffee
@@ -15,7 +15,7 @@ class Kernel
         console.log 'Kernel spec:', @kernelSpec
         console.log 'Kernel configuration:', @config
         console.log 'Kernel configuration file path:', @configPath
-        @language = @kernelSpec.grammarLanguage
+        @language = @kernelSpec.language
         @kernelName = @kernelSpec.display_name
         @executionCallbacks = {}
         @watchCallbacks = []
@@ -32,27 +32,13 @@ class Kernel
             atom.notifications.addInfo 'Using custom kernel connection:',
                 detail: @configPath
         else
-            if @language is 'python' and not @kernelSpec.argv?
-                commandString = 'ipython'
-                args = [
-                    'kernel',
-                    '--no-secure',
-                    "--hb=#{@config.hb_port}",
-                    "--control=#{@config.control_port}",
-                    "--shell=#{@config.shell_port}",
-                    "--stdin=#{@config.stdin_port}",
-                    "--iopub=#{@config.iopub_port}",
-                    '--colors=NoColor'
-                    ]
-
-            else
-                commandString = _.head(@kernelSpec.argv)
-                args = _.tail(@kernelSpec.argv)
-                args = _.map args, (arg) =>
-                    if arg is '{connection_file}'
-                        return @configPath
-                    else
-                        return arg
+            commandString = _.head(@kernelSpec.argv)
+            args = _.tail(@kernelSpec.argv)
+            args = _.map args, (arg) =>
+                if arg is '{connection_file}'
+                    return @configPath
+                else
+                    return arg
 
             console.log 'Kernel: Spawning:', commandString, args
             @kernelProcess = child_process.spawn commandString, args,


lib/kernel.coffee
```diff
@@ -15,7 +15,7 @@ class Kernel
         console.log 'Kernel spec:', @kernelSpec
         console.log 'Kernel configuration:', @config
         console.log 'Kernel configuration file path:', @configPath
-        @language = @kernelSpec.grammarLanguage
+        @language = @kernelSpec.language
```
lgeiger: `@language` was always `undefined`. Hence the hardcoded ipython kernel couldn't ever work.

I think we can remove it because nobody reported an issue 😉 .

n-riesco: We need to remove [this line](https://github.com/nteract/hydrogen/blob/e3927895dc903d408e5fe1624ff3e81cdaed3b59/lib/signal-list-view.coffee#L63) as well.

n-riesco:
We should provide a sample kernel spec for users like Ubuntu 16.04's that don't have `ipython kernelspec`. For example:

```json
{
  "kernelspecs": {
    "python2": {
      "spec": {
        "display_name": "Python 2",
        "language": "python",
        "argv": ["python2.7", "-m", "ipykernel", "-f", "{connection_file}"],
        "env": {}
      }
    },
    "python3": {
      "spec": {
        "display_name": "Python 3",
        "language": "python",
        "argv": ["python3.4", "-m", "ipykernel", "-f", "{connection_file}"],
        "env": {}
      }
    }
  }
}
```

lib/kernel.coffee
n-riesco: After some more thought and since `Kernel` gets a `grammar`, I think this line should be `@language =  KernelManager::getLanguageFor grammar`.

This would also suggest that  `getLanguageFor` should be a static method (I would leave that for another PR because `getLanguageFor` is used everywhere).

lgeiger: > We need to remove this line as well.
Done in #339 

> After some more thought and since Kernel gets a grammar, I think this line should be @language = KernelManager::getLanguageFor grammar.
Fixed!

lgeiger:
> We should provide a sample kernel spec for users like Ubuntu 16.04's that don't have `ipython kernelspec`.

Yes this should be added to the readme.

lib/kernel.coffee
```diff
@@ -15,7 +15,7 @@ class Kernel
         console.log 'Kernel spec:', @kernelSpec
         console.log 'Kernel configuration:', @config
         console.log 'Kernel configuration file path:', @configPath
-        @language = @kernelSpec.grammarLanguage
+        @language = KernelManager::getLanguageFor @grammar
```
n-riesco: We need to add `KernelManager = require './kernel-manager'` to the top of this file.

n-riesco: This isn't working.

I'm guessing this is caused by the dependency-cycle this statement creates between `KernelManager` and `Kernel`.

Since `Kernel::language` and `Kernel::grammar` aren't used anywhere in the code, I think it's best we remove them and update the code [here](https://github.com/lgeiger/hydrogen/blob/9a9726a677b8b91322457134717bd3b886dd8ee6/lib/kernel.coffee#L14), [here](https://github.com/lgeiger/hydrogen/blob/9a9726a677b8b91322457134717bd3b886dd8ee6/lib/kernel.coffee#L14), [here](https://github.com/nteract/hydrogen/blob/079608feca61a9c419cbcc4445cfc68252a271d7/lib/kernel-manager.coffee#L67) and [here](https://github.com/nteract/hydrogen/blob/079608feca61a9c419cbcc4445cfc68252a271d7/lib/kernel-manager.coffee#L76).

lgeiger: Oh I'm sorry I didn't restart atom while testing, so this error slipped through.

`Kernel::grammar` is used in this [line](https://github.com/nteract/hydrogen/blob/079608feca61a9c419cbcc4445cfc68252a271d7/lib/watch-view.coffee#L16). I wouldn't remove it.

I think we should switch back to `@language = @kernelSpec.language`

PR #344, opened 2016-07-11
README: Add example of kernelSpec setting

n-riesco:


lgeiger:
Great!

I think we should also update [this section](https://github.com/nteract/hydrogen/blob/master/README.md#dependencies) to prefer jupyter.

lgeiger:
I think to install the notebook `pip install jupyter` is enough.

lgeiger:
Awesome thanks!

PR #345, opened 2016-07-11
Kernel: Remove Kernel::language

n-riesco:
* Remove Kernel::language to avoid ambiguity.

* Use Kernel::getLanguageFor or Kernel::kernelSpec.display_name instead.


lgeiger:
`@language` is used [here](https://github.com/nteract/hydrogen/blob/master/lib/kernel.coffee#L83).

Either remove the usage or use `@kernelName`.

n-riesco:
Thx for catching that!

PR #347, opened 2016-07-13
Fix #336 (accept `text/html` inspection results)

### LINKED ISSUE #336: `hydrogen:inspect` is broken for `text/html` results
While testing the last PR I noticed that `hydrogen:inspect` has been broken for some time.

I won't be able to look at this until tonight or tomorrow.

n-riesco:
* Also added heuristics to guess a selection to inspect, when no text has been selected.

* Added two TODO items:
  * handle inspection errors
  * improve selection heuristics using `kernel.complete`

### DIFF
--- a/lib/inspector.coffee
+++ b/lib/inspector.coffee
@@ -17,37 +17,14 @@ class Inspector
             @view?.close()
             return
 
+        @view ?= new MessagePanelView
+            title: 'Hydrogen Inspector'
+            closeMethod: 'destroy'
+
         [code, cursor_pos] = @getCodeToInspect()
 
         kernel.inspect code, cursor_pos, (result) =>
-            console.log 'Inspector: Result:', result
-            found = result.found
-            if found is true
-                onInspectResult = ({mimetype, el}) =>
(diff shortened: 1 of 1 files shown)


lgeiger:
`13213ad` Works for me 👍 

I'm not sure about the heuristics. I thought they are handled by the kernel. [Jupyter messaging specs](https://jupyter-client.readthedocs.io/en/latest/messaging.html#introspection) says about the `cursor_position`:
> The cursor position within 'code' (in unicode characters) where inspection is requested

What improvement would the heuristics on our side bring for the inspect result? I get the same introspection results with and without `e86d241` on python.

n-riesco:
IJulia and IJavascript fail when `cursor_pos` is 0.

It is also up to the kernel how to handle cases like `Arr|ay`. Does it inspect `Arr` or `Array`?

lgeiger:
> IJulia and IJavascript fail when cursor_pos is 0.
It is also up to the kernel how to handle cases like Arr|ay. Does it inspect Arr or Array?

Oh I see. Then it makes sense!

PR #350, opened 2016-07-15
@showKernelPicker: Update kernelspecs if necessary

lgeiger:
In the future we should make the `getKernelSpecs` functions async as proposed in #330.

But for now I think it's ok to update the kernel specs if an activation command is triggered.


lgeiger:
I'll close this. Since I found a better solution for this problem

PR #351, opened 2016-07-15
Remove kernelMapping setting and handle kernelspecs asynchronous

lgeiger:
The first commit removes the `kernelMapping` setting. Instead the user is asked to chose a kernel to start.

The second commit introduces callbacks for all kernel specs handling. In order to update the kernel specs if necessary.

It probably makes sense to review each commit on its own.

### DIFF
--- a/lib/kernel-manager.coffee
+++ b/lib/kernel-manager.coffee
@@ -25,14 +25,6 @@ class KernelManager
 
 
     startKernelFor: (grammar, onStarted) ->
-        if _.isEmpty @_kernelSpecs
-            @updateKernelSpecs =>
-                @_startKernelFor grammar, onStarted
-        else
-            @_startKernelFor grammar, onStarted
-
-
-    _startKernelFor: (grammar, onStarted) ->
         language = @getLanguageFor grammar
         @getKernelSpecFor language, (kernelSpec) =>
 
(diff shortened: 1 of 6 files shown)


lib/kernel-manager.coffee
```diff
@@ -111,26 +112,23 @@ class KernelManager
...
+            @kernelPicker = new KernelPicker ->
+                return kernelSpecs
+            @kernelPicker.onConfirmed = ({kernelSpec}) ->
+                callback kernelSpec
+            @kernelPicker.toggle()
 
 
     kernelSpecProvidesLanguage: (kernelSpec, language) ->
         kernelLanguage = kernelSpec.language
-        mappedLanguage = Config.getJson('languageMappings')[kernelLanguage]
```
n-riesco: If `languageMappings` is removed, then users won't be able to use kernels with non-standard language names.

lib/main.coffee
```diff
@@ -105,17 +105,17 @@ module.exports = Hydrogen =
             kernel.interrupt()
 
         else if command is 'restart-kernel'
+            kernelSpec = kernel.kernelSpec
```
n-riesco: Neat!

lib/kernel-manager.coffee
```diff
@@ -98,33 +90,39 @@ class KernelManager
...
-            callback kernelSpecs[0]
-        else
-            @kernelPicker = new KernelPicker ->
-                return kernelSpecs
-            @kernelPicker.onConfirmed = ({kernelSpec}) ->
-                callback kernelSpec
-            @kernelPicker.toggle()
+        @getAllKernelSpecsFor language, (kernelSpecs) ->
+            if kernelSpecs.length <= 1
+                callback kernelSpecs[0]
```
n-riesco: `callback kernelSpecs[0] or []`

lib/kernel-manager.coffee
```diff
@@ -111,26 +112,23 @@ class KernelManager
...
         unless language?
             return null
 
-        kernelMapping = Config.getJson('kernelMappings')?[language]
-        if kernelMapping?
-            kernelSpecs = @getAllKernelSpecs().filter (spec) ->
-                return spec.display_name is kernelMapping
+        kernelSpecs = @getAllKernelSpecsFor language
+        if kernelSpecs.length <= 1
+            callback kernelSpecs[0]
```
n-riesco: `callback kernelSpecs[0] or []`

lib/kernel-manager.coffee
lgeiger: Ooops, I removed it by accident. Thanks for noticing!

lib/main.coffee
```diff
@@ -101,18 +101,25 @@ module.exports = Hydrogen =
         unless kernel
             kernel = @kernelManager.getRunningKernelFor language
 
+        message = "No running kernel for language `#{language}` found"
```
n-riesco: I think this kind of early optimisation makes the code less readable (please, read this comment just as an opinion).

lib/main.coffee
```diff
@@ -101,18 +101,25 @@ module.exports = Hydrogen =
...
+                atom.notifications.addError message
+                return
             kernelSpec = kernel.kernelSpec
             @kernelManager.destroyRunningKernel kernel
             @clearResultBubbles()
             @kernelManager.startKernel kernelSpec, grammar, =>
                 @setStatusBarElement()
 
         else if command is 'switch-kernel'
-            kernel = @kernelManager.getRunningKernelFor language
```
n-riesco: I think this line is potentially necessary, since the user can pass any kernel.

lib/main.coffee
lgeiger: I think I'll rename it to `errorMessage` which should make it clearer.

lib/main.coffee
lgeiger: `handleKernelCommands` is only used with the `command` and `kernelSpec` arguments.
@n-riesco Should I simplify the function and remove the unused arguments?

lgeiger: I'm wrong `SignalListView` uses all the arguments though I think this can be simplified in a future PR.

lgeiger:
@n-riesco Thanks for the feedback.

rgbkrk:
Thank you both!

PR #355, opened 2016-07-16
Select non standard kernel

lgeiger:
This PR is based on #351 

It allows the user to start any kernel, regardless of the editor grammar. Therefor a `select-non-standard-kernel` command was added to the `KernelPicker` and `SignalListView`.

### DIFF
--- a/lib/kernel-manager.coffee
+++ b/lib/kernel-manager.coffee
@@ -113,15 +113,16 @@ class KernelManager
         unless language?
             return null
 
-        @getAllKernelSpecsFor language, (kernelSpecs) ->
+        @getAllKernelSpecsFor language, (kernelSpecs) =>
             if kernelSpecs.length <= 1
                 callback kernelSpecs[0]
             else
                 unless @kernelPicker?
-                    @kernelPicker = new KernelPicker (onUpdated) ->
+                    @kernelPicker = new KernelPicker this, (onUpdated) ->
                         onUpdated kernelSpecs
                     @kernelPicker.onConfirmed = ({kernelSpec}) ->
                         callback kernelSpec
+
                 @kernelPicker.toggle()
 
 
--- a/lib/kernel-manager.coffee
+++ b/lib/kernel-manager.coffee
@@ -25,14 +25,6 @@ class KernelManager
 
 
     startKernelFor: (grammar, onStarted) ->
-        if _.isEmpty @_kernelSpecs
-            @updateKernelSpecs =>
-                @_startKernelFor grammar, onStarted
-        else
-            @_startKernelFor grammar, onStarted
-
-
-    _startKernelFor: (grammar, onStarted) ->
         language = @getLanguageFor grammar
         @getKernelSpecFor language, (kernelSpec) =>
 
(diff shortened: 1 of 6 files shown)


ajeebkp23:
Hi,
This is really interesting. Can I use any kernel ?
Can you please elaborate how to use with a kernel ?

I am really using it with django. It will be wonderful, if I can use it with the django-extension's notebook option. If you can, please tell me about using it with http://django-extensions.readthedocs.io/en/latest/shell_plus.html

I am really interested in this PR. Thanks for the PR.

ajeebkp23:
Please, Never mind about My shell_plus request. It is working fine with running a notebook with port 8888.

lib/kernel-manager.coffee
```diff
@@ -109,19 +109,21 @@ class KernelManager
             callback []
 
 
-    getKernelSpecFor: (language, callback) ->
+    getKernelSpecFor: (language, callback) =>
```
n-riesco: class methods don't use fat arrows (this is unlikely to be right)

lib/kernel-manager.coffee
```diff
@@ -109,19 +109,21 @@ class KernelManager
...
         unless language?
             return null
 
-        @getAllKernelSpecsFor language, (kernelSpecs) ->
+        @getAllKernelSpecsFor language, (kernelSpecs) =>
```
n-riesco: didn't coffeelint catch this one?

lib/kernel-manager.coffee
```diff
@@ -109,19 +109,21 @@ class KernelManager
...
             if kernelSpecs.length <= 1
                 callback kernelSpecs[0]
             else
                 unless @kernelPicker?
                     @kernelPicker = new KernelPicker (onUpdated) ->
                         onUpdated kernelSpecs
+                    , this
```
n-riesco: This looks awkward. Callbacks are usually last arguments:
```coffeescript
@kernelPicker = new KernelPicker this, (onUpdated) ->
    onUpdated kernelSpecs
```

lib/main.coffee
```diff
@@ -333,9 +333,12 @@ module.exports = Hydrogen =
                 language = @kernelManager.getLanguageFor grammar
                 @kernelManager.getAllKernelSpecsFor language, (kernelSpecs) ->
                     callback kernelSpecs
+            , @kernelManager
```
n-riesco: ditto

n-riesco:
@lgeiger  I'm not done with the review yet.

The description of commit 	6ddc1f1 should say `languageMappings`.

lgeiger:
@n-riesco Thanks for taking a look at the PR. I didn't mean to include [`6ddc1f1`](https://github.com/nteract/hydrogen/pull/355/commits/6ddc1f1b76ef01d7891bd9d74e8b41a9db11158a). There went something wrong with the rebase. I think we should keep the `languageMapping` setting.

lib/kernel-manager.coffee
lgeiger: I opted for this, since `@kernelManager` is optional but I'll change it.

lib/kernel-manager.coffee
lgeiger: This one is necessary because of the line 121

lgeiger:
@ajeebkp23 Hydrogen works in principle with every jupyter kernel. Hydrogen detects your installed kernels automatically if you have `jupyter` or a recent version of `ipython` installed. If not have al look at the `Kernel Spec` setting.

Hydrogen uses atoms grammar to determine which kernel to start. Some languages like `scala` use different kernel names. Therefor we provide the `Language Mappings` setting. This PR just allows you to start any kernel without editing the `Language Mappings` setting.

lgeiger:
Just some thoughts on my own PR:

In my workflow I don't really have the need to start a non standard kernel. Thus I'm fine if this doesn't get merged. At this point I'm not sure if we should have this feature or if it just clutters the menu.

PR #366, opened 2016-07-19
Keep watch sidebar in sync with editor

n-riesco:


lgeiger:
This is a great!

[This line](https://github.com/nteract/hydrogen/blob/master/lib/watch-sidebar.coffee#L32) has to be
```coffeescript
@toggleButton.onclick = ->
    editorView = atom.views.getView(editor)
    atom.commands.dispatch(editorView, 'hydrogen:toggle-watches')
```
in order to work consistent with the `@watchSidebarIsVisible` variable.

I think now we can also remove the `hydrogen:select-watch-kernel`command. What do you think?

It would be awesome to track the sidebar visibility on a kernel per kernel basis or only toggle the sidebar automatically if the sidebar isn't empty.

n-riesco:
On 19/07/16 14:24, Lukas Geiger wrote:
> This line <https://github.com/nteract/hydrogen/blob/master/lib/watch-sidebar.coffee#L32> has to be
> 
> @toggleButton.onclick = ->
>     editor = atom.workspace.getActiveTextEditor()
>     editorView = atom.views.getView(editor)
>     atom.commands.dispatch(editorView, 'hydrogen:toggle-watches')
> 
> in order to work consistent with the |@watchSidebarIsVisible| variable.

Again, good catch!

> It would be awesome to track the sidebar visibility on a kernel per kernel basis or only toggle the sidebar automatically if the sidebar isn't empty.

That's how I tried first, but I found it a bit inconvenient, because often I want the watch sidebar visible for all the running kernels.

At least to me, it makes more sense to set the visibility regardless the kernel (the inspect panel behaves this way too).

lgeiger:
> That's how I tried first, but I found it a bit inconvenient, because often I want the watch sidebar visible for all the running kernels.
At least to me, it makes more sense to set the visibility regardless the kernel (the inspect panel behaves this way too).

That's fine for me. We can revisit this behavior if we decide to add support for multiple kernels with the same language someday.

n-riesco:
I noticed `clearBubblesOnRow` hasn't been working for a while.
I've included a fix in this PR.

lgeiger:
Works well 👍 

I never noticed that `clearBubblesOnRow` wasn't working. Thanks for fixing that!

PR #370, opened 2016-07-19
Fix #369: html output size

### LINKED ISSUE #369: $$html$$ output size problem
Hi there,

When trying to run the following [IJavascript example](http://n-riesco.github.io/ijavascript/doc/graphics.ipynb.html) 

```
$$html$$ = "<div style='background-color:olive;width:50px;height:50px'></div>";
```

I get 

image

This is what happens if I add something in the div like:

```
$$html$$ = "<div style='background-color:olive;width:50px;height:50px'>test</div>";
```

image

The next `$$svg$$` example works but it looks like the canvas size is too big:

image

Cheers!

Jun

lgeiger:
@n-riesco You added these lines in #207. Are there any special cases in which we needed them?

### DIFF
--- a/styles/hydrogen.less
+++ b/styles/hydrogen.less
@@ -112,17 +112,12 @@
         overflow: auto;
     }
 
-    .bubble-result-container > * {
+    .bubble-result-container {
         width: fit-content !important;
         height: fit-content !important;
         max-width: unset !important;
         max-height: unset !important;
         overflow: hidden !important;
-    }
-
-    .bubble-result-container {
-        width: inherit;
-        height: inherit;
         white-space: normal;
         pre {
             font-size: @font-size;


n-riesco:
If I remember correctly those lines are meant to deal with large `DataFrame`s; e.g.
```python
import pandas as pd
import numpy as np
pd.DataFrame(np.random.randn(100, 16), columns=list('ABCDEFGHabcdefgh'))
```

I've just tested it. With this PR the horizontal scrollbar is only visible after scrolling to the bottom:

screenshot from 2016-07-20 14-48-02

and without, the horizontal scrollbar is always visible:

screenshot from 2016-07-20 14-50-21

lgeiger:
On my system the scrollbars are displayed correctly with this PR.

n-riesco:
@lgeiger Do you mean that you don't need to scroll down to the bottom of the DataFrame to see the horizontal scroll bar?

lgeiger:
> @lgeiger Do you mean that you don't need to scroll down to the bottom of the DataFrame to see the horizontal scroll bar?

Yes I see both.

n-riesco:
I'm using Pandas 0.17.1 and the DataFrame is styled like this `<div style="max-width:1500px;overflow:auto;">`.

lgeiger:
I tried with pandas `0.17.1` and everything works fine for me.
The DataFrame is styled like this `<div><table border="1" class="dataframe">`. The problem is caused by `overflow: auto`.

n-riesco:
I can reproduce this issue in Ubuntu 14.04 (using Pandas 0.13.1 and IPython 4.1.2) and Ubuntu 16.04 (using Pandas 0.17.1 and IPython 2.4.1).

lgeiger:
@n-riesco I updated the PR. Please test if it works for you.

styles/hydrogen.less
```diff
@@ -112,7 +112,7 @@
         overflow: auto;
     }
 
-    .bubble-result-container > * {
+    .bubble-result-container > {
```
n-riesco: Remove `>`

n-riesco: Also remove [these lines](https://github.com/lgeiger/hydrogen/blob/9dfc7c7a94d458c8971d14e36842b1070f712e0a/styles/hydrogen.less#L124-L125)

n-riesco: Tested and it works for me.

lgeiger:
Great! Works for me as well

PR #372, opened 2016-07-21
Update troubleshooting document

n-riesco:


lgeiger:
Awesome thanks for going through all the issues! 👍

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,87 @@
...
+  print sys.executable
+  ```
+
+  Let's say this script prints out `c:\Anaconda2\python.exe`. You can configure
+  Atom to use this executable by running
+  `apm config set python c:\Anaconda2\python.exe`. Then, Hydrogen can be
+  installed by running `apm install hydrogen`.
+
+
+### No kernel for language X found, but I have a kernel for that language.
```
lgeiger: Should we add an item about the `Language Mappings` setting if the kernel uses a different name (e.g. `scala211`)?

n-riesco: I haven't tested #355 yet. Shall we wait until then?

lgeiger: Oh I forgot about #355.

Issue #377, opened 2016-07-24
In[] or In[{number}] syntax problem


jmatsushita:
In the image below I've done `hydrogen:run-cell` in all cells. As you can see the cells with the `// %%` and `// <codecell>` syntax work just fine, but the `In[]` or `In[1]` syntax doesn't seem to be picked up. In fact when running `hydrogen:run-cell-and-move-down` in any of the 4 top cells, the cursor moves to the 5th cell.

![image](https://cloud.githubusercontent.com/assets/356097/17086602/a7d7c8aa-51c4-11e6-9811-2771fe373a80.png)

lgeiger:
The syntax for defining code cells is `// In[ ]:`. Notice the space (or a number) between the brackets and the colon at the end.
@n-riesco Should we remove the need for the space between the brackets?

@jmatsushita Which syntax theme are you using? There should be syntax highlighting for the correct breakpoints:
<img width="248" alt="bildschirmfoto 2016-07-25 um 08 17 08" src="https://cloud.githubusercontent.com/assets/13285808/17092153/d2a5af76-5240-11e6-9123-dcfddb5bbae7.png">

n-riesco:
@lgeiger Is the issue caused by the missing space between the brackets or the missing colon `:`?

+1 for making the space and numbers optional. I'd feel tempted to make the colon optional too.

lgeiger:
> Is the issue caused by the missing space between the brackets or the missing colon `:`?

Both.

The idea about the space and the colon was to adhere to the `jupyter notebook` export format. They are using `// In[ ]:` for not executed cells and `// In[1]:` for executed ones.

I'm OK with removing the need for the space but I think without the colon the syntax highlighting would look odd. However I don't feel strongly about this. What do you think?
<img width="237" alt="bildschirmfoto 2016-07-25 um 08 40 45" src="https://cloud.githubusercontent.com/assets/13285808/17092522/8fbd9658-5243-11e6-9d7d-ff3860750a13.png">

n-riesco:
How does spyder deal with `// In[ ]:`?

The reason why I thought optional spaces and colons are better is that users can easily make the mistake of leaving them out. Syntax highlight should've helped. @jmatsushita , do you know why in your screenshot the syntax highlight doesn't distinguish between normal comments and breakpoints?

The grammar highlight issue with optional colons would be solved by:
```cson
'match': '%%| %%| <codecell>| In\\[[0-9 ]+\\]:*'
```

[...]

PR #378, opened 2016-07-25
Run All: Add support for code cells

### LINKED ISSUE #375: Cell boundaries
Hi there,

> Please note I've only used Hydrogen with IJavascript so I'm not sure if my observations only apply to this combination.

I think I understand that compared to a Jupyter notebook, it's difficult for Hydrogen to determine the boundaries of cells to enable the display of intermediary computation results in Atom when the whole notebook is ran? Is that the reason why intermediary results aren't displayed or updated?

For languages (like Javascript) where an optional line delimiter (`;`) can be used, wouldn't it be possible to use the presence of the delimiter as a cue to render the cell?

For other cases, or in addition to this, maybe Hydrogen could remember where existing cell runs have been requested and when a `run all` command is issued, it would update these bubbles (it seems that currently - with IJavascript - its only updating the final checkmark).

I guess that the more general case is to try to introspect the code to define the appropriate cell boundaries. This seems almost like something that would fit in a Kernel extension if there was interest upstream in allowing this type of freeform notebook. 

Let me know if that's unclear and I can explain more visually what I'm suggesting.

Cheers!

Jun

lgeiger:
Closes #375

### DIFF
--- a/grammars/breakpoints.cson
+++ b/grammars/breakpoints.cson
@@ -3,7 +3,7 @@
 'injectionSelector': 'comment.line'
 'patterns': [
   {
-    'match': '%%| %%| <codecell>| In\\[[0-9 ]+\\]:'
+    'match': '%%| %%| <codecell>| In\\[[0-9 ]*\\]:?'
     'name': 'entity.name.section.breakpoint'
   }
 ]
--- a/lib/cell-manager.coffee
+++ b/lib/cell-manager.coffee
@@ -62,6 +62,6 @@ module.exports = CellManager =
             escapeStringRegexp commentStartString.trimRight()
 
         regexString =
-            escapedCommentStartString + '(%%| %%| <codecell>| In\[[0-9 ]+\]:)'
+            escapedCommentStartString + '(%%| %%| <codecell>| In\[[0-9 ]*\]:?)'
 
         return regexString
--- a/lib/cell-manager.coffee
+++ b/lib/cell-manager.coffee
@@ -4,24 +4,15 @@ module.exports = CellManager =
     getCurrentCell: ->
         editor = atom.workspace.getActiveTextEditor()
         buffer = editor.getBuffer()
-        scope = editor.getRootScopeDescriptor()
-
-        {row, column} = editor.getLastCursor().getBufferPosition()
         start = 0
         end = editor.getLastBufferRow()
+        regexString = @getRegexString editor
 
-        {commentStartString, commentEndString} =
-            editor.languageMode.commentStartAndEndStringsForScope(scope)
+        unless regexString?
+            return [start, end]
 
-        unless commentStartString
-            console.log 'CellManager: No comment string defined in root scope'
-            return
-
-        escapedCommentStartString =
-            escapeStringRegexp commentStartString.trimRight()
-        regex = new RegExp(
-            escapedCommentStartString + '(%%| %%| <codecell>| In\[[0-9 ]+\]:)'
-        )
+        regex = new RegExp regexString
+        {row, column} = editor.getLastCursor().getBufferPosition()
 
         if row > 0
             range = [[0, 0], [row - 1, 100]]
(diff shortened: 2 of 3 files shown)


lgeiger:
Fixes #377

grammars/breakpoints.cson
```diff
@@ -3,7 +3,7 @@
 'injectionSelector': 'comment.line'
 'patterns': [
   {
-    'match': '%%| %%| <codecell>| In\\[[0-9 ]+\\]:'
+    'match': '%%| %%| <codecell>| In\\[[0-9 ]*\\]:*'
```
n-riesco: My bad, I should've written:
```cson
'match': '%%| %%| <codecell>| In\\[[0-9 ]*\\]:?'
```

lib/cell-manager.coffee
```diff
@@ -59,7 +59,7 @@ module.exports = CellManager =
         escapedCommentStartString =
             escapeStringRegexp commentStartString.trimRight()
         regex = new RegExp(
-            escapedCommentStartString + '(%%| %%| <codecell>| In\[[0-9 ]+\]:)',
+            escapedCommentStartString + '(%%| %%| <codecell>| In\[[0-9 ]*\]:*)',
```
n-riesco: ditto

grammars/breakpoints.cson
lgeiger: Oh yes thats better

n-riesco:
We don't handle the case of inline comments  yet; e.g. : 
```coffeescript
a = 1 //%%
a += 1
//%%
```

But this is existing issue, so I'd leave the fix for another PR.

lib/cell-manager.coffee
```diff
@@ -37,3 +27,39 @@ module.exports = CellManager =
...
+                breakpoints.push range.start.row
+
+        breakpoints.push editor.getLastBufferRow()
+
+        console.log 'CellManager: Breakpoints:', breakpoints
+
+        return breakpoints
+
+
+    getRegex: (editor, option = '') ->
```
n-riesco: No need to act on this comment, but probably a `getRegexString` would've been a more versatile function.

lib/cell-manager.coffee
```diff
@@ -37,3 +27,39 @@ module.exports = CellManager =
...
+
+    getBreakpoints: ->
+        editor = atom.workspace.getActiveTextEditor()
+        buffer = editor.getBuffer()
+        breakpoints = [0]
+
+        regex = @getRegex editor, 'g'
+        if regex?
+            buffer.scan regex, ({range}) ->
+                breakpoints.push range.start.row
```
n-riesco: Again no need to act this comment, but to address the case of inline comments fully, here we may need to store `breakpoints.push range.start`/

lgeiger: I'll address this in a future PR. But probably not today.

lib/cell-manager.coffee
lgeiger: You're right.  I wasn't really happy with the name either.

lib/cell-manager.coffee
```diff
@@ -37,3 +27,39 @@ module.exports = CellManager =
...
+
+        breakpoints.push editor.getLastBufferRow()
+
+        console.log 'CellManager: Breakpoints:', breakpoints
+
+        return breakpoints
+
+
+    getRegexString: (editor, option = '') ->
```
n-riesco: What I meant is that this function would be more versatile if it returned a string, and let the caller decide whether they want a Regex and what kind of options the regex should take.

n-riesco: I'm sorry I didn't want to cause more unnecessary work for you.

lgeiger: Oh now I understand.

No problem (thanks to `git rebase -i`) 😄

PR #381, opened 2016-07-26
Code cells: Handle inline comments

### LINKED ISSUE #380: Code cells: Handle inline comments
We don't handle the case of inline comments yet; e.g. :
```coffeescript
a = 1 //%%
a += 1
//%%
```
Moved from [#378](https://github.com/nteract/hydrogen/pull/378#issuecomment-234909566)

lgeiger:
Fix #380

### DIFF
--- a/lib/cell-manager.coffee
+++ b/lib/cell-manager.coffee
@@ -12,18 +12,21 @@ module.exports = CellManager =
             return [start, end]
 
         regex = new RegExp regexString
-        {row, column} = editor.getLastCursor().getBufferPosition()
+        cursor = editor.getLastCursor().getBufferPosition()
 
-        if row > 0
-            range = [start, [row - 1, 9999999]]
-            buffer.backwardsScanInRange regex, range, ({range}) ->
+        while cursor.row < end.row and @isComment editor, cursor
+            cursor.row += 1
+            cursor.column = 0
+
+        if cursor.row > 0
+            buffer.backwardsScanInRange regex, [start, cursor], ({range}) ->
                 start = range.start
 
-        range = [[row, 0], end]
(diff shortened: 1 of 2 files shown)


lgeiger:
e.g.:
```python
print('Cell', 1)  # %%
print('Cell', 2)
# %%
print('Cell', 3)
```

lib/cell-manager.coffee
```diff
@@ -15,15 +15,13 @@ module.exports = CellManager =
         {row, column} = editor.getLastCursor().getBufferPosition()
 
         if row > 0
-            range = [[0, 0], [row - 1, 100]]
+            range = [start, [row - 1, 9999999]]
```
n-riesco: How about `range = [start, [row, column]]`?

lgeiger: That could lead to problems if the cursor is inside the comment defining the code cell.
But it would be more accurate otherwise.

n-riesco: On 26/07/16 10:38, Lukas Geiger wrote:
> In lib/cell-manager.coffee <https://github.com/nteract/hydrogen/pull/381#discussion_r72217505>:
> 
>> @@ -15,15 +15,13 @@ module.exports = CellManager =
>>          {row, column} = editor.getLastCursor().getBufferPosition()
>>  
>>          if row > 0
>> -            range = [[0, 0], [row - 1, 100]]
>> +            range = [start, [row - 1, 9999999]]
> 
> That could lead to problems if the cursor is inside the comment defining the code cell.
> But it would be more accurate otherwise.

is it possible to detect whether we're inside a comment? I mean using the grammar.

lgeiger: That's a great idea. I addressed it in the latest commit.

PR #395, opened 2016-08-03
Remove watch kernel picker

lgeiger:
This functionality is now provided by #366.

The sidebar will now look like this:
watches

### DIFF
--- a/lib/watch-sidebar.coffee
+++ b/lib/watch-sidebar.coffee
@@ -14,13 +14,9 @@ class WatchSidebar
         toolbar = document.createElement('div')
         toolbar.classList.add('toolbar', 'block')
 
-        languageDisplay = document.createElement('button')
-        languageDisplay.classList.add('btn', 'icon', 'icon-sync')
-        languageDisplay.innerText = "Watch: #{@kernel.kernelSpec.display_name}"
-        languageDisplay.onclick = ->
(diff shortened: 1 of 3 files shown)


lib/watch-sidebar.coffee
```diff
@@ -37,12 +33,10 @@ class WatchSidebar
         tooltips = new CompositeDisposable()
         tooltips.add atom.tooltips.add toggleButton,
             title: 'Toggle Watches'
```
n-riesco: How about `Toggle Watch Sidebar`?

n-riesco:
@lgeiger Here's an idea. Currently the `x` icon of the result view in the watch sidebar only closes the result view (it doesn't remove the watch). If the `x` icon removed the watch, then `removeButton` would be unnecessary.

I always have to think twice to remember what all those buttons do in the watch sidebar.

lgeiger:
>@lgeiger Here's an idea (for another PR). Currently the x icon of the result view in the watch sidebar only closes the result view (it doesn't remove the watch). If the x icon removed the watch, then removeButton would be unnecessary.

This is not that easy since we would have to add `removeButtons` to every watch.

Would a different icon (instead of `x`) make the functionality clearer?

n-riesco:
Obviously I didn't do a good job at explaining this idea. Nevermind, I've opened a new issue #397.

PR #391, opened 2016-08-03
Fix connection to existing kernel (issue #374)

n-riesco:

PR #396, opened 2016-08-03
Fix restarting kernels connected via `connection.json` (issue #394)

### LINKED ISSUE #394: Restarting kernels connected via `connection.json` is broken

n-riesco:
Fixes #394

### DIFF
--- a/lib/kernel-manager.coffee
+++ b/lib/kernel-manager.coffee
@@ -16,7 +16,8 @@ class KernelManager
 
 
     destroy: ->
-        _.forEach @_runningKernels, (kernel) => @destroyRunningKernel kernel
+        _.forEach @_runningKernels, (kernel) -> kernel.destroy()
+        @_runningKernels = {}
 
 
     setRunningKernelFor: (grammar, kernel) ->
@@ -27,9 +28,11 @@ class KernelManager
         @_runningKernels[language] = kernel
 
 
-    destroyRunningKernel: (kernel) ->
-        delete @_runningKernels[kernel.kernelSpec.language]
-        kernel.destroy()
+    destroyRunningKernelFor: (grammar) ->
+        language = @getLanguageFor grammar
+        kernel = @_runningKernels[language]
+        delete @_runningKernels[language]
+        kernel?.destroy()
 
 
     startKernelFor: (grammar, onStarted) ->
--- a/lib/kernel-manager.coffee
+++ b/lib/kernel-manager.coffee
@@ -51,6 +51,7 @@ class KernelManager
             return
 
         console.log 'KernelManager: restartRunningKernelFor: ignored', kernel
+        atom.notifications.addWarning 'Cannot restart this kernel'
         onRestarted?(kernel)
 
 
--- a/lib/kernel-manager.coffee
+++ b/lib/kernel-manager.coffee
@@ -5,6 +5,7 @@ fs = require 'fs'
 path = require 'path'
 
 Config = require './config'
+WSKernel = require './ws-kernel'
 ZMQKernel = require './zmq-kernel'
 KernelPicker = require './kernel-picker'
 
@@ -33,6 +34,24 @@ class KernelManager
         kernel = @_runningKernels[language]
         delete @_runningKernels[language]
         kernel?.destroy()
+
+
+    restartRunningKernelFor: (grammar, onRestarted) ->
+        language = @getLanguageFor grammar
+        kernel = @_runningKernels[language]
+
+        if kernel instanceof WSKernel
+            kernel.restart().then -> onRestarted?(kernel)
+            return
+
+        if kernel instanceof ZMQKernel and kernel.kernelProcess
+            kernelSpec = kernel.kernelSpec
+            @destroyRunningKernelFor grammar
+            @startKernel kernelSpec, grammar, (kernel) -> onRestarted?(kernel)
+            return
+
+        console.log 'KernelManager: restartRunningKernelFor: ignored', kernel
+        onRestarted?(kernel)
 
 
     startKernelFor: (grammar, onStarted) ->
--- a/lib/kernel-manager.coffee
+++ b/lib/kernel-manager.coffee
@@ -19,6 +19,14 @@ class KernelManager
         _.forEach @_runningKernels, (kernel) => @destroyRunningKernel kernel
 
 
+    setRunningKernelFor: (grammar, kernel) ->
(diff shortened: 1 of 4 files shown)


lib/zmq-kernel.coffee
```diff
@@ -17,12 +17,11 @@ class ZMQKernel extends Kernel
 
         @executionCallbacks = {}
 
-        projectPath = path.dirname(
```
n-riesco: leftover from #391

lgeiger:
This still isn't working for me. If I connect to an existing kernel and use the menu to restart it, I still get an error ([here](https://github.com/nteract/hydrogen/blob/master/lib/main.coffee#L200)) because the kernel can't be restarted.
I'm currently away the next two weeks so won't be able to fix this issue.

lib/zmq-kernel.coffee
```diff
@@ -44,15 +43,16 @@ class ZMQKernel extends Kernel
...
+                console.log 'ZMQKernel: stderr:', data
 
                 regexp = getKernelNotificationsRegExp()
                 if regexp?.test data
                     atom.notifications.addError @kernelSpec.display_name,
                         detail: data, dismissable: true
         else
-            atom.notifications.addInfo 'Using custom kernel connection:',
-                detail: @connectionFile
+            console.log 'ZMQKernel: connectionFile:', connectionFile
```
lgeiger: This is working but shouln't it be `@connectionFile`?

lib/zmq-kernel.coffee
```diff
@@ -93,15 +93,24 @@ class ZMQKernel extends Kernel
...
+            console.log 'ZMQKernel: cannot interrupt an existing kernel'
+            atom.notifications.addWarning 'Cannot interrupt this kernel'
+
+
+    shutdown: (restart) ->
+        requestId = 'shutdown_' + uuid.v4()
+        message = @_createMessage 'shutdown_request', requestId
+
+        message.content =
+            restart: not not restart
```
lgeiger: Why aren't you using a default value in the function definition?

lib/zmq-kernel.coffee
```diff
@@ -306,30 +320,27 @@ class ZMQKernel extends Kernel
...
-        console.log 'sending shutdown'
+        console.log 'ZMQKernel: destroy:', this
 
-        message = @_createMessage 'shutdown_request'
-        message.content =
-            restart: false
-
-        @shellSocket.send new jmp.Message message
+        if @kernelProcess?
+            @interrupt()
```
lgeiger: Is sending `@kernelProcess.kill 'SIGINT'` working correctly? Before we were using `@kernelProcess.kill 'SIGKILL'`.

lib/zmq-kernel.coffee
```diff
@@ -306,30 +320,27 @@ class ZMQKernel extends Kernel
...
+            fs.unlink @connectionFile
+        else
+            @shutdown()
```
lgeiger: I think we should send an shutdown message even if `@kernelProcess` is defined.

lib/zmq-kernel.coffee
```diff
@@ -306,30 +320,27 @@ class ZMQKernel extends Kernel
...
 
         @shellSocket.close()
+        @controlSocket.close()
         @ioSocket.close()
+        @stdinSocket.close()
```
lgeiger: 👍

lib/zmq-kernel.coffee
n-riesco: Good catch, `SIGINT` is for interrupting the kernel and `SIGKILL` is to kill it.

n-riesco:
We should create a label for `late-night-PR-review-with-care`. :disappointed:

lgeiger:
> We should create a label for `late-night-PR-review-with-care`. 😞

😄 😄 👍

rgbkrk:
We're sitting on 67 commits since the last release, do you think we'll be shipping a new version soon, possibly with this PR?

n-riesco:
[WIP] Not ready for merging

n-riesco:
PR is ready for review/merging.

lib/kernel-manager.coffee
```diff
@@ -35,6 +36,24 @@ class KernelManager
...
+            kernel.restart().then -> onRestarted?(kernel)
+            return
+
+        if kernel instanceof ZMQKernel and kernel.kernelProcess
+            kernelSpec = kernel.kernelSpec
+            @destroyRunningKernelFor grammar
+            @startKernel kernelSpec, grammar, (kernel) -> onRestarted?(kernel)
+            return
+
+        console.log 'KernelManager: restartRunningKernelFor: ignored', kernel
```
lgeiger: I think we should add an atom notification here.

lib/zmq-kernel.coffee
```diff
@@ -306,30 +329,27 @@ class ZMQKernel extends Kernel
 
 
     destroy: ->
```
lgeiger: I think we should add the `restart` option from `@shutdown`.

n-riesco: On 07/08/16 08:40, Lukas Geiger wrote:
> In lib/zmq-kernel.coffee <https://github.com/nteract/hydrogen/pull/396#discussion_r73797331>:
> 
>> @@ -306,30 +329,27 @@ class ZMQKernel extends Kernel
>>  
>>  
>>      destroy: ->
> 
> I think we should add the |restart| option from |@shutdown|.

I'd find this confusing. Because of its name, `destroy` makes me think that we expect it be used as a class destructor.

I wouldn't expect a class instance to be usable after invoking its destructor (and so, the restart option wouldn't make sense).

Do we have any use for `destroy(restart)`?

lgeiger: OK I see, it makes sense.
I was just wondering because `@shutdown` has an restart option which isn't used anywhere. But this can be left for an future PR.

lgeiger:
Thanks for this great PR! (And sorry for the late reply)

lib/zmq-kernel.coffee
n-riesco: Unfortunately, the use of the `shutdown` message is undermined by frontends like the notebook. When the notebook restarts a kernel, what actually happens is that the notebook kills the kernel process and an relaunches a new kernel.

Judging by [this proposal](https://github.com/jupyter/enhancement-proposals/pull/14), this behaviour won't change in the future.

PR #399, opened 2016-08-09
KernelManager: Set kernel working directory

### LINKED ISSUE #398: Importing modules from same dir broken
After updating to 0.12.0 I am not able to import a python module I wrote (it sits in the same directory as the executed script). Firing up an ipython interpreter in the terminal and importing said module works like a charm though. I am on OS X 10.11.6, atom 1.9.6.
`apm --version` gives this output:
```
app  1.10.0
nam  2.13.3
node 0.10.40
python 3.5.2
git 2.7.4
```
I am using python from within an anaconda virtual environment, in case this might play a role.

n-riesco:
Fixes #398

### DIFF
--- a/lib/kernel-manager.coffee
+++ b/lib/kernel-manager.coffee
@@ -108,17 +108,23 @@ class KernelManager
 
         console.log 'KernelManager: startKernelFor:', language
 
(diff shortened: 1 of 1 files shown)


lgeiger:
Thanks for the fast fix! 🎉

PR #405, opened 2016-08-13
Pass the environment in for spawnOptions

### LINKED ISSUE #404: Missing environment variables
Sorry if this has been covered before, can't seem to find a solution.

When I examine `os.environ` from `jupyter` or `ipython3` or `python3` I can see all of my shell environment variables.

However when I run the same using Hydrogen (Python 3 kernel) in the latest version of Atom, many variables are missing. Specifically, I'm having problems with some CUDA variables needed for Python libraries. This is true whether I start Atom via the command line or otherwise.

Anything I might be missing? Is there a way to manually pass environment variables to Hydrogen?

rgbkrk:
Closes #404

### DIFF
--- a/lib/kernel-manager.coffee
+++ b/lib/kernel-manager.coffee
@@ -113,6 +113,7 @@ class KernelManager
         )
         spawnOptions =
             cwd: projectPath
+            env: process.env
         launchSpec(kernelSpec, spawnOptions).
             then ({config, connectionFile, spawn}) =>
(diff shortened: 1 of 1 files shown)


rgbkrk:
I think it would be ok/appropriate to set this in spawnteract by default, while allowing it to be overridden using `Object.assign` on the `opts` object.

lgeiger:
Thanks! I'll go ahead and make a patch release.

PR #406, opened 2016-08-16
Bump spawnteract

lgeiger:


rgbkrk:
Excellent. Yay for good defaults in spawnteract.

Issue #383, opened 2016-07-28
TODOs: Remote kernels via kernel gateway


lgeiger:
This issue sums up the TODOs mentioned in #330. Feel free to add more items to the list.
cc. @nikitakit @n-riesco 

- [x] document functionality in readme.
- [x] implement support for watches
- [x] filter sessions depending on current grammar
- [ ] implement "switch to"
- [x] implement "shut down"
- [x] implement STDIN support

---

(Edited by @n-riesco)
- [x] if no gateways have been setup, do not show the WSKernel picker
- [x] restart doesn't clear the result bubbles (fixed in PR #396 )
- [x] add command to disconnect from a remote kernel
- [x] `Kernel` constructor should take the arguments `(@kernelSpec, @grammar)`, because `WatchView`s expect `Kernel` to have the property `grammar`.

lgeiger:
Wow I can't believe this list got checked off so fast 🎉💃.
Thanks a lot for your work @nikitakit!

I'm going to close this issue since I think the "switch to" command isn't really useful for remote kernels. There are multiple gateways and sessions so it would result in a quite confusing menu.

PR #408, opened 2016-08-17
Remote kernels: support watches

nikitakit:
Now that I'm back from vacation, it's time to address the backlog of issues in #383 

This PR adds watch support to WSKernel. `kernel.grammar` changes are also here because `WatchView` expects this property to be present.

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -351,10 +351,10 @@ module.exports = Hydrogen =
             @wsKernelPicker = new WSKernelPicker (kernel) =>
                 @clearResultBubbles()
 
-                grammar = @editor.getGrammar()
+                grammar = kernel.grammar
                 @kernelManager.destroyRunningKernelFor grammar
 
                 @kernelManager.setRunningKernelFor grammar, kernel
                 @onKernelChanged kernel
 
-        @wsKernelPicker.toggle()
+        @wsKernelPicker.toggle @editor.getGrammar()
--- a/lib/kernel.coffee
+++ b/lib/kernel.coffee
@@ -22,6 +22,11 @@ class Kernel
         @watchCallbacks.push(watchCallback)
 
 
+    _callWatchCallbacks: ->
+        @watchCallbacks.forEach (watchCallback) ->
+            watchCallback()
+
+
     interrupt: ->
         throw new Error 'Kernel: interrupt method not implemented'
 
--- a/lib/kernel.coffee
+++ b/lib/kernel.coffee
@@ -11,7 +11,7 @@ WatchSidebar = require './watch-sidebar'
 
 module.exports =
 class Kernel
-    constructor: (@kernelSpec) ->
+    constructor: (@kernelSpec, @grammar) ->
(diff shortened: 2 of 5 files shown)


lgeiger:
Awesome, great to see you back!

lib/main.coffee
```diff
@@ -351,10 +351,10 @@ module.exports = Hydrogen =
             @wsKernelPicker = new WSKernelPicker (kernel) =>
                 @clearResultBubbles()
 
-                grammar = @editor.getGrammar()
+                grammar = kernel.grammar
```
lgeiger: Why are you using `kernel.grammar` instead of `@editor.getGrammar()`?

nikitakit: I'm defending against a race condition where the active grammar changes between the time that the `WSKernelPicker` is activated, and the time a kernel is selected.

Technically there should never happen. However the `WSKernelPicker` can have significant delays over poor network connections, and perhaps some tab-switching from another plugin can sneak into the event loop during that time.

lgeiger: The problem is that `@kernelManager.destroyRunningKernelFor` isn't working with `kernel.grammar` because kernels are identified via `@editor.grammar` (e.g. `python` instead of `python 3`).

I think we should use `@editor.getGrammar()` for now.

nikitakit: As far as I can see from the code, `kernel.grammar` is set to the value of `editor.getGrammar()` at the time when the kernel is constructed (following the call chain from `main.coffee::createResultBubble`)

There is a separate `language` field, but I don't believe that it is in play here.

lgeiger: Oops, I'm sorry I didn't read `ps aux` correctly. I had an other python process running and thought it was related to hydrogen :wink:.

n-riesco: I haven't reviewed the whole PR yet, but here are my thoughts on this point.

I think this discussion has uncovered a design flaw in `KernelManager`:
- running kernels are indexed by grammar language
- when a kernel is created, the kernel is assigned the current `@editor.getGrammar()`
- when a kernel is indexed as a running kernel, the kernel is indexed by the language corresponding to the current `@editor.getGrammar()`

Since the time of creation and the time of registration as running kernel are different, it is possible that those values of `@editor.getGrammar()` may be inconsistent.

The call to `@wsKernelPicker.toggle @editor.getGrammar()` makes me think that they must be consistent, but I'm not certain. @lgeiger , can you trigger this inconsistency?

If there is no way to trigger this inconsistency `kernel.grammar` seems more appropriate to me. But I think we should look at removing `kernel.grammar` to avoid any inconsistencies (this may be difficult because of the way we handle watches).

lgeiger: I triggered an inconsistency like @nikitakit mentioned.
To trigger it, select a remote kernel (preferably over a poor internet connection) and quickly switch to an other tab with different grammar. `@editor.getGrammar()` and `kernel.grammar` will not be the same (I checked for equality at line 355).

PR #409, opened 2016-08-17
Remote kernels: support shutdown (plus shutdown menu item)

nikitakit:
Continuing with the issues listed in #383, this is another (disjoint) set of changes.

The first commit implements the "shut down" method for remote kernels.

The second adds "shut down" to the list of available kernel commands. It's accessible via both the status bar and the atom command list. This is especially useful for remote kernels, because they don't shut down automatically through any other code path (as intended).

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -70,6 +70,8 @@ module.exports = Hydrogen =
                 @handleKernelCommand command: 'interrupt-kernel'
             'hydrogen:restart-kernel': =>
                 @handleKernelCommand command: 'restart-kernel'
+            'hydrogen:shutdown-kernel': =>
+                @handleKernelCommand command: 'shutdown-kernel'
             'hydrogen:copy-path-to-connection-file': =>
                 @copyPathToConnectionFile()
 
@@ -190,6 +192,13 @@ module.exports = Hydrogen =
             @clearResultBubbles()
             @kernelManager.restartRunningKernelFor grammar, (kernel) =>
                 @onKernelChanged kernel
+
+        else if command is 'shutdown-kernel'
+            @clearResultBubbles()
+            # Note that destroy alone does not shut down a WSKernel
+            kernel.shutdown()
+            @kernelManager.destroyRunningKernelFor grammar
+            @onKernelChanged()
 
         else if command is 'switch-kernel'
             @clearResultBubbles()
--- a/lib/ws-kernel.coffee
+++ b/lib/ws-kernel.coffee
@@ -19,8 +19,11 @@ class WSKernel extends Kernel
     interrupt: ->
         @session.kernel.interrupt()
 
+    shutdown: ->
+        return @session.kernel.shutdown()
+
     restart: ->
-        @session.kernel.restart()
+        return @session.kernel.restart()
 
     _onStatusChange: ->
         @statusView.setStatus @session.status
--- a/lib/kernel.coffee
+++ b/lib/kernel.coffee
@@ -31,7 +31,7 @@ class Kernel
         throw new Error 'Kernel: interrupt method not implemented'
 
 
-    shutdown: (restart) ->
+    shutdown: ->
         throw new Error 'Kernel: shutdown method not implemented'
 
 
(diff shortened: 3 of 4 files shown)


lib/main.coffee
```diff
@@ -189,6 +191,13 @@ module.exports = Hydrogen =
             @kernelManager.restartRunningKernelFor grammar, (kernel) =>
                 @onKernelChanged kernel
 
+        else if command is 'shutdown-kernel'
+            @clearResultBubbles()
+            # Note that destroy alone does not shut down a WSKernel
+            kernel.shutdown()
+            @kernelManager.destroyRunningKernelFor grammar
+            @onKernelChanged()
```
nikitakit: Note that `@kernel` is set to `null` at startup, but is set to `undefined` by `onEditorChanged` whenever the current tab doesn't have a running kernel (see `KernelManager::getRunningKernelFor`).

lib/main.coffee
```diff
@@ -189,6 +191,13 @@ module.exports = Hydrogen =
             @kernelManager.restartRunningKernelFor grammar, (kernel) =>
                 @onKernelChanged kernel
 
+        else if command is 'shutdown-kernel'
+            @clearResultBubbles()
+            # Note that destroy alone does not shut down a WSKernel
+            kernel.shutdown()
+            @kernelManager.destroyRunningKernelFor grammar
```
lgeiger: Is it possible to modify `@kernelManager.destroyRunningKernelFor` to work for `WSKernels` as well?

lib/ws-kernel.coffee
```diff
@@ -18,8 +18,14 @@ class WSKernel extends Kernel
...
 
+    shutdown: (restart = false) ->
+        if restart
+            return @session.kernel.restart()
+        else
+            return @session.kernel.shutdown()
+
     restart: ->
-        @session.kernel.restart()
+        return @session.kernel.restart()
```
lgeiger: Sidenote:
```coffeescript
restart: ->
    @session.kernel.restart()
```
and
```coffeescript
restart: ->
    return @session.kernel.restart()
```

compile to the same javascript code.

lib/ws-kernel.coffee
```diff
@@ -18,8 +18,14 @@ class WSKernel extends Kernel
     interrupt: ->
         @session.kernel.interrupt()
 
+    shutdown: (restart = false) ->
+        if restart
```
lgeiger: The restart option in `@shutdown` for `ZMQKernel` is just a parameter in the [shutdown message](https://github.com/nteract/hydrogen/blob/master/lib/zmq-kernel.coffee#L119) and doesn't really restart the kernel. (I don't really know what it is doing instead, maybe @n-riesco knows.)

I think for `WSKernel` we don't need the restart option in `@shutdown`.

lgeiger:
Sorry for the late review. Besides the minor things I mentioned above this PR works very well. Great to have you onboard 🎉

lib/ws-kernel.coffee
nikitakit: I'm aware (though this came as a surprise to me, because I don't think hydrogen coding style generally uses implicit returns)

lgeiger: Great, no need to act on my comment. I mentioned it, because I was surprised too :)

lib/ws-kernel.coffee
nikitakit: After re-reading the jupyter messaging spec, it looks like you're right.

I'll remove this handling here to avoid confusion

lib/main.coffee
nikitakit: Please no. The current situation exists to avoid accidental data loss, which is a major concern for me.

If I'm just temporarily attaching to a long-running kernel hosted elsewhere, I don't want the detach operation to forcibly kill the kernel! (`destroyRunningKernelFor` is called anytime you switch between kernels.)

Please let me know if there's any additional comments I can add to the code to make sure that people refactoring the code in the future are aware of this.

lgeiger: Oh of course. Thanks for reminding me 👍

PR #410, opened 2016-08-17
Remote kernels: stdin support

nikitakit:
Continuing with the checklist from #383, here is stdin support for WSKernel.

### DIFF
--- a/lib/ws-kernel.coffee
+++ b/lib/ws-kernel.coffee
@@ -6,6 +6,7 @@ uuid = require 'uuid'
 services = require('./jupyter-js-services-shim')
 
 Kernel = require './kernel'
+InputView = require './input-view'
 
 module.exports =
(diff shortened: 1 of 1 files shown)


lib/ws-kernel.coffee
```diff
@@ -50,6 +51,20 @@ class WSKernel extends Kernel
                 stream: 'status'
             onResults?(result)
 
+        future.onStdin = (message) =>
+            if message.header.msg_type is not 'input_request'
```
lgeiger: I would prefer 
```coffeescript
unless message.header.msg_type is 'input_request'
```
since it is used more often in our code base.

nikitakit: Right. No wonder I felt that I hadn't ever seen `if ... is not` in hydrogen before.

PR #411, opened 2016-08-17
Remote kernels: notification if no gateways specified in settings

nikitakit:
This is yet another (independent) PR addressing part of #383

If no gateways have been set up, this presents a nicer error message when "Connect to remote kernel" is run.

### DIFF
--- a/lib/ws-kernel-picker.coffee
+++ b/lib/ws-kernel-picker.coffee
@@ -42,17 +42,20 @@ class WSKernelPicker
         @_onChosen = onChosen
 
     toggle: (grammar) ->
+        gateways = Config.getJson('gateways', [])
+        if _.isEmpty gateways
+            atom.notifications.addError 'No remote kernel gateways available',
+                description: 'Use the Hydrogen package settings to specify the list
+                of remote servers. Hydrogen can use remote kernels on either a
+                Jupyter Kernel Gateway or Jupyter notebook server.'
+            return
+
         @_grammar = grammar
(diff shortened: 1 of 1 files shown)


lib/ws-kernel-picker.coffee
```diff
@@ -42,17 +42,21 @@ class WSKernelPicker
         @_onChosen = onChosen
 
     toggle: (grammar) ->
+        gateways = Config.getJson('gateways', [])
+        if _.isEmpty gateways
+            atom.notifications.addError 'No remote kernel gateways available',
+                dismissable: true
```
lgeiger: I think `dismissable: false` should be the way to go. Then the notification gets dismissed after a few second automatically. This is the default behavior so removing this line will work.

lgeiger:
The line wrapping isn't working for me:
bildschirmfoto 2016-08-18 um 12 37 57
This may be a problem with atom. Are you experiencing the same behavior?

nikitakit:
The truncated display is not what I was seeing before:

hydrogen-no-gateway

However, even the character-level wrapping above is still undesirable.

Can you try the new version I just pushed? It uses `description` (rendered as text and auto-wrapped) instead of `detail` (rendered as monospace). However the docs for this API are very sparse, so I can't tell if this is supported on older versions of atom. I'm currently on atom 1.8.0

lgeiger:
Thanks investigating! This works for me on `v1.9.8`.

It looks like this API was available for some time now but the docs where missing (See https://github.com/atom/atom/pull/11942).

Issue #403, opened 2016-08-11
Connect jupyter console to existing kernel


neurotronix:
I usually execute some python 3 code via hydrogen and then connect to the existing kernel via jupyter console. However, the directory where I would usually find the kernel config files must have changed since I am not able anymore to connect to an existing kernel. I would really appreciate a comment on that issue. Many thanks in advance!

n-riesco:
We're now using https://github.com/nteract/spawnteract to launch a kernel.

I think the easiest way to locate the connection file is to use the dev console. After starting a kernel, press `shift-ctrl-i` to open the dev console and look for `ZMQKernel` in the console log. You should see something like this:

![screenshot from 2016-08-11 14-55-21](https://cloud.githubusercontent.com/assets/6199391/17591053/40f4ba62-5fd4-11e6-995a-3f1a6e1644be.png)

Alternatively, you could run `jupyter --runtime-dir` to get the runtime folder.

neurotronix:
Perfect, thank you!

lgeiger:
@neurotronix In version `v0.13.0` we added a command to copy the path to the connection file. In order to allow connecting to the same kernel via `jupyter console --existing <path to connectionfile>`.

Issue #365, opened 2016-07-19
Add console support


spencerlyon2:
It would be great to have a hydrogen console. 

I often work where I leave semi-permanent code in the editor and turn to the console or REPL for experimental code. Being able to do this all inside with hydrogen connected to the same kernel would be _awesome_!

The [atom-ink](https://github.com/JunoLab/atom-ink) project provides the machinery for such a console. 

Is it within the scope of hydrogen to build something like this? If so, it is feasible to hook up atom-ink?

n-riesco:
There is a similar request in #166  .

Note that during an Atom session, all the kernels for a given language share the same kernel. So you could simply open a new panel and use it as an REPL.

We also have some functionality to reuse kernel connections. See the section for docker containers [here](https://github.com/nteract/hydrogen#custom-kernel-connection-inside-docker).

spencerlyon2:
What I had in mind seems different from both opening a separate buffer to act as a "repl" or spawning a qt-console. 

The atom-ink package provides a fairly standard REPL experience with partial readline support, command history, etc. Using ink might not be the right solution here, but I'm referring to it because I think it is a good example of what I had in mind.

n-riesco:
I'm going to label this issue as feature enhancement.

I've also noticed this feature was already listed in our [mostly-ignored-by-us roadmap wiki page](https://github.com/nteract/hydrogen/wiki/Roadmap) :blush: .

lgeiger:
@spencerlyon2 I created a new package called [`hydrogen-launcher`](https://atom.io/packages/hydrogen-launcher) which allows you to spawn a new terminal with a `jupyter console` or `qtconsole` connected to the running Hydrogen kernel.

spencerlyon2:
That's great, thanks!

Issue #166, opened 2016-02-17
Command to open respective QT-console


schlichtanders:
Feature-Request

it would be great to have a command which starts a qt-console/jupyter-process on the kernel used by hydrogen.
Extra Feature: If the qt-console/jupyter-process can optionally be embedded into atom directly, this would be nice, of course.

rgbkrk:
Hey @schlichtanders!

Design wise, I'd prefer that we emit an event that another package can listen to for this. That Atom package can then launch the qt-console with:

```bash
jupyter qtconsole --existing /path/to/Jupyter/runtime/kernel-16113.json
```

> If the qt-console/jupyter-process can optionally be embedded into atom directly, this would be nice, of course.

As another pane within Atom with a qt-console like experience? I agree! Probably again as a different Atom package, and should definitely exist.

schlichtanders:
Dear Kyle Kelley,

how would one distinguish the different languages?
E.g. open a qt-console for a running python-kernel and not a R-kernel,
which may run in parallel?

(note, that I lack how events work in atom - maybe my question sounds
rather senseless when having such a background).

rgbkrk:
Generally for Jupyter spec conforming projects, a connection file is created somewhere like `~/Library/Jupyter/runtime`:

```
$ ls ~/Library/Jupyter/runtime/
kernel-20275.json      notebook_cookie_secret
```

Within that file is the information for connecting to that kernel:

```
$ cat ~/Library/Jupyter/runtime/kernel-20275.json
{
  "stdin_port": 51222,
  "ip": "127.0.0.1",
  "control_port": 51223,
  "hb_port": 51224,
  "signature_scheme": "hmac-sha256",
  "key": "17f8a27f-612c-4c63-be5f-95a718243cf6",
  "shell_port": 51220,
  "transport": "tcp",
  "iopub_port": 51221
}
```

Any frontend, including `jupyter console`, `jupyter qtconsole`, [`sidecar`](https://github.com/nteract/sidecar), [`nteract`](https://github.com/nteract/nteract), and others can use this to connect to the backing runtime/language. The rest that flows from there is the [jupyter message spec](http://jupyter-client.readthedocs.org/en/latest/messaging.html).

schlichtanders:
thanks for explaining

rgbkrk:
At the command line, open up `jupyter console` then run `%connect_info`:

```
$ jupyter console
Jupyter Console 4.1.0.dev

[...]

PR #416, opened 2016-08-22
Add command to copy path to connection file

lgeiger:
This allows the user to easily connect to a ZMQKernel via a external console.

Related to #166, #365 and #403

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -70,6 +70,8 @@ module.exports = Hydrogen =
                 @handleKernelCommand command: 'interrupt-kernel'
             'hydrogen:restart-kernel': =>
                 @handleKernelCommand command: 'restart-kernel'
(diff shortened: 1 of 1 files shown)


rgbkrk:
Neat!

lib/main.coffee
```diff
@@ -358,3 +360,25 @@ module.exports = Hydrogen =
...
 
         @wsKernelPicker.toggle @editor.getGrammar()
+
+
+    copyPathToConnectionFile: ->
+        grammar = @editor.getGrammar()
+        language = @kernelManager.getLanguageFor grammar
+
+        unless @kernel?
+            message = "No running kernel for language `#{language}` found"
```
n-riesco: I think the message should be `No running kernel associated with this file` to account for the case  a kernel hasn't been launched yet.

This would make lines 366 and 367 unnecessary.

PR #421, opened 2016-08-29
Remote kernels: add command to disconnect from WSKernel

nikitakit:
This is another request brought up in #383 

This command is accessible via the status bar whenever a WSKernel is active. I didn't add it to the global list of atom commands because calling this command when a ZMQKernel is active makes no sense.


lgeiger:
Works great 👍

PR #422, opened 2016-08-29
Remote kernels: filter sessions and kernels by language

nikitakit:
When connecting to a remote gateway, only the existing sessions that match the current grammar are presented as options. If creating a new session, only the kernelspecs matching the current grammar are made available.

This makes remote kernels more closely match local kernels, as requested in #383 

(I have additionally renamed some variables in this PR, because the overuse of the word "spec" made the old code extremely hard to understand)

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -373,7 +373,11 @@ module.exports = Hydrogen =
                 @kernelManager.setRunningKernelFor grammar, kernel
                 @onKernelChanged kernel
 
-        @wsKernelPicker.toggle @editor.getGrammar()
+        grammar = @editor.getGrammar()
+        language = @kernelManager.getLanguageFor grammar
+
+        @wsKernelPicker.toggle grammar, (kernelSpec) =>
+            @kernelManager.kernelSpecProvidesLanguage(kernelSpec, language)
 
 
     copyPathToConnectionFile: ->
--- a/lib/ws-kernel-picker.coffee
+++ b/lib/ws-kernel-picker.coffee
@@ -41,7 +41,7 @@ class WSKernelPicker
     constructor: (onChosen) ->
         @_onChosen = onChosen
 
-    toggle: (grammar) ->
+    toggle: (@_grammar, @_kernelSpecFilter) ->
         gateways = Config.getJson('gateways', [])
         if _.isEmpty gateways
             atom.notifications.addError 'No remote kernel gateways available',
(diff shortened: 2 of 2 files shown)


lib/main.coffee
```diff
@@ -368,7 +368,11 @@ module.exports = Hydrogen =
                 @kernelManager.setRunningKernelFor grammar, kernel
                 @onKernelChanged kernel
 
-        @wsKernelPicker.toggle @editor.getGrammar()
+        grammar = @editor.getGrammar()
+        language = @kernelManager.getLanguageFor grammar
+
+        @wsKernelPicker.toggle @editor.getGrammar(), (kernelSpec) =>
```
lgeiger:
```coffeescript
@wsKernelPicker.toggle grammar, (kernelSpec) =>
```

lib/ws-kernel-picker.coffee
```diff
@@ -51,71 +51,85 @@ class WSKernelPicker
             return
 
         @_grammar = grammar
+        @_kernelSpecFilter = kernelSpecFilter
```
lgeiger: You can do this assignment in the method definition like this:
```coffeescript
toggle: (@_grammar, @_kernelSpecFilter) ->
```

lib/ws-kernel-picker.coffee
```diff
@@ -51,71 +51,85 @@ class WSKernelPicker
...
-            services.getKernelSpecs(spec.options)
-            .then (specmodels) =>
-                items = _.map specmodels.kernelspecs, (specmodel) =>
-                    options = Object.assign({}, spec.options)
-                    options.kernelName = specmodel.name
-                    options.path = @_path
+    onGateway: (gatewayInfo) ->
+        services.getKernelSpecs(gatewayInfo.options)
+        .then (specmodels) =>
+            kernelSpecs = _.filter specmodels.kernelspecs, (specmodel) =>
```
lgeiger: I'm not a native speaker but I think using camel casing for `specModels` and `specModel` would be better. At least it would be consistent with `sessionModels`. What do you think?

lib/ws-kernel-picker.coffee
```diff
@@ -51,71 +51,85 @@ class WSKernelPicker
...
+            kernelListing.setItems items
+            unless sessionInfo.name?
+                kernelListing.setError 'This gateway does not support listing sessions'
         else
-            services.connectToSession(spec.model.id, spec.options).then(@onSessionChosen.bind this)
+            services.connectToSession(sessionInfo.model.id, sessionInfo.options).then(@onSessionChosen.bind this)
 
-    startSession: (spec) ->
-        services.startNewSession(spec.options).then(@onSessionChosen.bind this)
+    startSession: (info) ->
```
lgeiger: I would prefer `sessionInfo` as above.

lib/ws-kernel-picker.coffee
nikitakit: That makes sense

nikitakit:
@lgeiger Thanks for reviewing!

Just pushed a new version addressing your comments

PR #424, opened 2016-08-29
WSKernelPicker: Tildify session path

lgeiger:
PR:
bildschirmfoto 2016-08-29 um 23 24 36
Master:
bildschirmfoto 2016-08-29 um 23 27 01

I still don't like the `uuid` at the end. Is there a better possibility of uniquely identifying a session? Maybe the time of creation?

### DIFF
--- a/lib/ws-kernel-picker.coffee
+++ b/lib/ws-kernel-picker.coffee
@@ -1,5 +1,6 @@
 {SelectListView} = require 'atom-space-pen-views'
 _ = require 'lodash'
+tildify = require 'tildify'
 
 Config = require './config'
 services = require './jupyter-js-services-shim'
@@ -79,7 +80,7 @@ class WSKernelPicker
                         return true
                 items = sessionModels.map (model) ->
                     if model.notebook?.path?
-                        name = model.notebook.path
+                        name = tildify model.notebook.path
                     else
                         name = "Session #{model.id}"
                     return {
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "lodash": "^4.14.0",
     "requirejs": "^2.2.0",
     "spawnteract": "^2.1.1",
+    "tildify": "^1.2.0",
     "transformime": "^3.1.0",
     "transformime-marked": "0.0.1",
     "uuid": "^2.0.1",


nikitakit:
Still need to think about this some more (and also about uuids). Looks reasonable at first glance, but I'm curious: how does tildify work for remote machines where the home folder is in a different location?

lgeiger:
> how does tildify work for remote machines where the home folder is in a different location?

Good point!

It determines the home path like [this](https://github.com/sindresorhus/os-homedir/blob/master/index.js). This should work on remote machines as well.
Do you have the possibility to test with a remote machine?

nikitakit:
Oh right, even though the kernels are remote the filenames are still coming from the local filesystem.

This PR looks good to me.

Regarding session names, perhaps the time of creation would make for cleaner names. The other thing is that it would be nice to have a session renaming option in the signal list view. I didn't do this before because I didn't know how to prompt for input, but now that I thought about it the prompt could just reuse the InputView class.

nikitakit:
Oh, I see that I now have push access. Thanks, @lgeiger !

As I said above, this PR lgtm.

But before I start hitting merge buttons for code I've looked at, could you tell me a bit about any workflow policies/expectations in place here? I also want to make sure that me merging code doesn't come as an unpleasant surprise to any of the other contributors, including those who I haven't interacted with lately. Thanks!

lgeiger:
> Oh, I see that I now have push access. Thanks, @lgeiger !

I forgot to do that when we invited you to nteract. So welcome to the party 🎉 

We follow the [C4.1 process](http://rfc.zeromq.org/spec:22/C4/), the essence of it is described [here](https://github.com/nteract/nteract/blob/master/CONTRIBUTING.md#what-does-c41-mean-for-me).

It's great to have you on board and we are allways happy to welcome new contributors!

cc @rgbkrk @n-riesco

rgbkrk:
Yup! Welcome!

Main thing to keep in mind is that we want to keep moving the code together. We tend to merge sooner and try to iterate. The only thing that should hold us back is breaking userspace in some way. Since we're largely UI focused, we don't have to worry about APIs as much.

PR #425, opened 2016-08-30
Remote kernels: add rename command

nikitakit:
This adds a command for renaming remote sessions.

The names were originally intended to be paths to notebook files, but in the case of Hydrogen they can really be anything. The only consequence of picking a non-path name is that the notebook web UI can't be used to talk to that kernel.

I had to add a new text input view, because I wanted it to be cancellable by pressing esc. The InputView doesn't do this, which is probably a good thing in the case of input.

### DIFF
--- a/lib/signal-list-view.coffee
+++ b/lib/signal-list-view.coffee
@@ -29,6 +29,11 @@ class SignalListView extends SelectListView
 
         @wsKernelCommands = [
             {
+                name: 'Rename'
+                value: 'rename-kernel'
+                language: null
+            },
+            {
                 name: 'Disconnect from'
                 value: 'disconnect-kernel'
                 language: null
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -206,6 +206,9 @@ module.exports = Hydrogen =
             @kernelManager.startKernel kernelSpec, grammar, (kernel) =>
                 @onKernelChanged kernel
 
+        else if command is 'rename-kernel'
+            kernel.promptRename?()
+
         else if command is 'disconnect-kernel'
             @clearResultBubbles()
             @kernelManager.destroyRunningKernelFor grammar
--- a/lib/rename-view.coffee
+++ b/lib/rename-view.coffee
@@ -1,0 +1,40 @@
+{$, TextEditorView, View} = require 'atom-space-pen-views'
+
+module.exports =
+class RenameView extends View
+    @content: (@prompt) ->
+        @div =>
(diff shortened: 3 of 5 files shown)


lib/ws-kernel-picker.coffee
```diff
@@ -124,7 +124,9 @@ class WSKernelPicker
             unless sessionInfo.name?
                 kernelListing.setError 'This gateway does not support listing sessions'
         else
-            services.connectToSession(sessionInfo.model.id, sessionInfo.options).then(@onSessionChosen.bind this)
+            services.connectToSession(sessionInfo.model.id, sessionInfo.options).then (session) =>
+                session.update?(sessionInfo.model) # Work around https://github.com/jupyter/jupyter-js-services/issues/214
```
lgeiger: Great this issue is fixed in `v0.18.1` of `jupyter-js-services`

lgeiger:
There is a lot of duplicated code in `RenameView` and `InputView`.
I'm a bit unsure if introducing the new class or modifying `InputView` to work for both use cases is better. @rgbkrk @n-riesco @nikitakit What do you think?

nikitakit:
I'll defer stylistic questions to someone who has worked more in the atom/js ecosystem.

Though personally, I should say that I think it's possible to reduce the number of view-related classes even further. For example the SignalListView and the CustomListView I wrote for the WSKernelPicker aren't really that different. It might even be possible to use the parent SelectListView class without subclassing it, by creating instances and on-the-spot overriding a few methods that need to contain application logic.

rgbkrk:
If you're unsure if modifying both to work is the best, I'd aim for creating the new one until a refactor is obvious later.

nikitakit:
I've pushed a rebase that bumps the `jupyter-js-services` dependency, so the workaround for the path bug is no longer included.

Let me know if you'd like me to refactor the view code in any way.

lgeiger:
Great I'll test now.
I'm good with doing the refactoring in in the future.

lib/signal-list-view.coffee
```diff
@@ -29,6 +29,11 @@ class SignalListView extends SelectListView
 
         @wsKernelCommands = [
             {
+                name: 'Rename'
```
lgeiger: For me `Rename session for` or something similar would be clearer. I find `Rename Python 3 kernel` a bit misleading.

Other than that everything works beautifully 👍

lgeiger: I have started cleaning up `SignalListView` some time ago. I'll address this in this PR.

Issue #331, opened 2016-07-08
[Feature request] Show execution count in results bubble


nikitakit:
The Jupyter notebook numbers any return values from cells, e.g. `Out [3]`. This number is called the `execution_count` in the Jupyter messaging spec.

I often find myself wanting to refer to the output of a previous computation, but the reference is currently not exposed through hydrogen. I think it would be useful to render the number somewhere or add a copy-execution-count button to the right sidebar.

nikitakit:
@lgeiger @rgbkrk Could I ask for your feedback about possible UI designs for this?

This feature would have a big impact on my workflow, so I'm trying to take a shot at implementing it.

The tricky thing is that hydrogen uses different display styles for one-line outputs (inline display) and multi-line (it embeds a little frame with the output and some buttons). This feature would need to fit in with both styles. What do you think about displaying the execution count on mouseover? (Add it to the little bubble that says "Copy to clipboard").

rgbkrk:
Now that we have the inline decorators, I've wondered why we can't always show it below.

As for the mouse over, that sounds like a good idea.

The only other thing that comes to mind is showing Out[num]. I'll be thinking on this.

lgeiger:
Closed by #472

PR #428, opened 2016-09-02
Add new button to insert result into document

fasiha:
Hello wonderful Hydrogen developers!

This pull request currently works beautifully for me (working currently with JavaScript and ijs), but I’d love to get feedback on how to generalize it and get it into Hydrogen proper, if you think it’s worth it.

Here’s what it does:

animation_with_pause

In words: the results box has an additional button that grabs the results and inserts them below the selection. It currently wraps the results in `/* result: <results> */`, which is very JavaScript-specific, I know, and which is currently hardcoded into the button’s callback.

I’m surprised how frequently I want to inject the results of evaluation back into my document—it’s enabling me to be much more literate in both my code (à la literate programming) and my Markdown files (now that I have the ability to tell Hydrogen to use ijs or jupyter while editing Markdown).

If you think this is more widely useful, I’m more than happy to think about how to generalize this to work for all Hydrogen users, by customizing the output to be language-aware (i.e., use `#` for Python, `/**/` for JS, etc.), etc.

If you’re only lukewarm on the entire idea, let me know and I can write a blog post about how I use this functionality with examples.

In any case, feedback is much-appreciated!

### DIFF
--- a/lib/result-view.coffee
+++ b/lib/result-view.coffee
@@ -70,7 +70,9 @@ class ResultView
             bubbleText = @getAllText()
             editor = atom.workspace.getActiveTextEditor()
             editor.insertNewlineBelow()
-            editor.insertText('/* result: \n' + bubbleText + '\n*/')
+            editor.insertText('result: \n' + bubbleText + '\n',
+              select: true)
+            editor.toggleLineCommentsInSelection()
         actionPanel.appendChild(insertButton)
 
         @setMultiline false
--- a/lib/result-view.coffee
+++ b/lib/result-view.coffee
@@ -63,6 +63,16 @@ class ResultView
                 editor.insertText(bubbleText)
         actionPanel.appendChild(openButton)
 
+        insertButton = document.createElement('div')
+        insertButton.classList.add 'action-button',
+            'insert-button', 'icon', 'icon-triangle-down'
+        insertButton.onclick = =>
+            bubbleText = @getAllText()
+            editor = atom.workspace.getActiveTextEditor()
+            editor.insertNewlineBelow()
+            editor.insertText('/* result: \n' + bubbleText + '\n*/')
+        actionPanel.appendChild(insertButton)
+
         @setMultiline false
 
         @tooltips = new CompositeDisposable()
@@ -70,6 +80,8 @@ class ResultView
             title: 'Copy to clipboard'
         @tooltips.add atom.tooltips.add openButton,
             title: 'Open in new editor'
+        @tooltips.add atom.tooltips.add insertButton,
+            title: 'Insert below'
 
         @_hasResult = false
 


lgeiger:
Yeah 🎉

This could be a great feature for a lot of users. It would be awesome to generalize this so we can include it in Hydrogen.

To get the comment strings for the current grammar take a look [these lines](https://github.com/nteract/hydrogen/blob/master/lib/code-manager.coffee#L143). Maybe you can also use [this function](https://github.com/atom/atom/blob/master/src/language-mode.coffee#L26) of atom or find inspiration there.

A blog post about hydrogen would be awesome ❤️. If you decide to write one hit us up on twitter @nteractio so we can spread the word.

If you have any questions feel free to ask here or on slack!

lgeiger:
> To get the comment strings for the current grammar take a look [these lines](https://github.com/nteract/hydrogen/blob/master/lib/code-manager.coffee#L143).

Shoot! This was removed in [`Atom v1.11.0`](https://github.com/atom/atom/commit/b97f317cd1cdc42a95c37274ed2818f665fd921f). Thanks for bringing this up!

fasiha:
@lgeiger thanks for your kind words 🙇, and helpful advice! It will now insert the result text, select it, and calls Atom’s built-in `toggleLineCommentsInSelection`—modularity FTW!

lgeiger:
Neat! 🎉

lib/result-view.coffee
```diff
@@ -63,13 +63,27 @@ class ResultView
...
         actionPanel.appendChild(openButton)
 
+        insertButton = document.createElement('div')
+        insertButton.classList.add 'action-button',
+            'insert-button', 'icon', 'icon-triangle-down'
+        insertButton.onclick = =>
+            bubbleText = @getAllText()
+            editor = atom.workspace.getActiveTextEditor()
+            editor.insertNewlineBelow()
+            editor.insertText('result: \n' + bubbleText + '\n',
```
lgeiger: For me a capitalized `Result:` would be a bit nicer

lib/result-view.coffee
```diff
@@ -63,13 +63,27 @@ class ResultView
...
+              select: true)
+            editor.toggleLineCommentsInSelection()
+        actionPanel.appendChild(insertButton)
```
lgeiger: This functionality doesn't really make sense inside the watch panel.
```coffeescript
if @marker?
    actionPanel.appendChild(insertButton)
```
This should (haven't tested this yet) hide the button if in the watch sidebar.

nikitakit:
I like this idea!

I'm wondering if it makes sense to use `Out[...]` instead of `result` as the title, to match the notebook. (See also #331). On the other hand, this numbering scheme might be too cryptic if the ultimate use-case is to generate a document for the public.

Also (and I don't want this idea to delay merging the current version of this feature), but perhaps there's a way to make this more generally applicable. One use case is an in-atom iBash terminal, where the output doesn't need to be escaped at all (this encroaches a bit on what [xiki](http://xiki.org/) does). Another might be that in markdown documents, where the code gets escaped in code-blocks, but the output may need to be inserted plain. Though I'm not sure if there's a general way to do this without adding a bunch of complicated options.

fasiha:
@nikitakit thanks for your feedback! I’ve looked at [org-mode Babel](http://orgmode.org/worg/org-contrib/babel/), and the wealth of options and customizability offered by that emacs package assures me that this rabbit hole goes very, verrrry deep 😱…

Thinking about this, and about @lgeiger’s comment on the header (“result” vs “Result” vs something customizable vs …), made me really wish Atom had a proper macro system, letting one compose a sequence of editor actions, which one names, then executes whenever. I wound up installing [atom-macros](https://atom.io/packages/atom-macros) and discovered that I could make a tidy little macro to invoke after *copying* the Hydrogen result, and which accomplishes exactly the same thing as this pull request—
```coffee
# in ~/atom/macros.coffee after installing atom-macros
@pasteAsComment = ->
  # Copy Hydrogen results first, then invoke this macro
  editor.insertNewlineBelow()
  editor.insertText('result: \n' + atom.clipboard.read() + '\n',
    select: true)
  editor.toggleLineCommentsInSelection()
```
This required no changes to Hydrogen, and offers the user complete customizability!—to comment or not, to have a header or not, etc.

Having said all this, I’m sure Hydrogen doesn’t want a dependency on atom-macros nor does it want to reimplement its functionality, not just to give users the ability to customize how to insert results into their document, so I’m not sure where that leaves this pull request. I still think it’s a very nice feature, I just don’t know if, without adding a couple of user-customizable options (to control the header and comment-or-not), it can be made into something suitable for all its potential users?

lgeiger:
@fasiha Thanks for investigating and sorry about the late reply. I feel the same about the user customizability. I'm not sure if this is something we can achieve without introducing a lot of overhead code.

That said I've been experimenting with Atoms service API for interacting with Hydrogen and have a plugin working which has access to the kernel connecting file provided by Hydrogen.

(Disclamer: I have no idea if this will work)
We could allow plugins to override the "open in a new text editor" button of the result view or introduce some kind of a magic button for them to use. The plugins should then provide a function to be executed if the button is clicked.

If you are interested in developing a package which extends hydrogen with this functionality, I'm more than happy to investigate if this is really possible.

fasiha:
Thanks @lgeiger! Your comment motivated me to finally read about the Services API, and I have something working—it’s extremely janky but it demonstrates the concept I think:

Hydrogen provides a service called “most recent results”. When you run a command, Hydrogen updates some internal state which the service provider grabs and returns whenever it’s invoked. A consumer plugin can call that provider and get an array of Jupyter responses to the most recent “Run”. This accomplishes my task!

I’d like some feedback about that approach but I don’t think I should push those changes into this pull request’s branch? Should I just open another pull request?

rgbkrk:
Another PR sounds good. I'd love to see this services approach.

lgeiger:
@fasiha Thanks a lot for your work!

I'm going to close this in favor of #441 since it provides more flexibility.

rgbkrk:
@fasiha I've also added you on as a collaborator, even though this wasn't merged because both PRs and general contributions have been incredibly thoughtful. Thanks for building Hydrogen with us. Welcome to the nteract party! :tada:

PR #433, opened 2016-09-03
Don't move the cursor down if there is no following code to execute

lgeiger:
This prevents the cursor skipping to the last line. Helpful if there are a lot of comments at the end of a file.

Brought up by @matthew-brett on [slack](https://nteract.slack.com/archives/general/p1472939771000014).

### DIFF
--- a/lib/code-manager.coffee
+++ b/lib/code-manager.coffee
@@ -182,14 +182,12 @@ class CodeManager
     moveDown: (row) ->
         lastRow = @editor.getLastBufferRow()
 
-        if row >= lastRow
-            @editor.moveToBottom()
-            @editor.insertNewline()
-            return
-
         while row < lastRow
             row++
             break if not @isBlank(row)
+
+        unless row < lastRow
+            return
 
         @editor.setCursorBufferPosition
             row: row


n-riesco:
I would like to revert this PR. It breaks the use case where a user is adding code at the end of a file and running it.

n-riesco:
Or alternatively, we could make sure that `move down`  doesn't skip comments.

lgeiger:
Thanks for going through the PRs, I really appreciate your feedback 👍 

> It breaks the use case where a user is adding code at the end of a file and running it.

The idea behind this PR was, if you have a large amount of comments at the end of a file it the command wouldn't scroll down to the very bottom of the page. 

> Or alternatively, we could make sure that move down doesn't skip comments.

That would probably be the best option as long as we ensure that executing each line of the file without manually moving the cursor down is still possible.
~~We could move the logic for skipping the comments into the function which determines what code to execute.~~ Oh no this isn't possible. Do you have an idea on how to achieve this?

n-riesco:
@lgeiger I won't look at it now. I've opened an issue so that we don't forget.

PR #438, opened 2016-09-08
Plugin API

lgeiger:
This will provide the connection file for plugins to use. I'll start to document everything next week.

I have no idea on how to correctly design the API. Your feedback is greatly appreciated.

I used this to write a quick [plugin](https://github.com/lgeiger/hydrogen-launcher) to launch jupyter consoles (~~⚠️  WIP~~):
hydrogen

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -73,8 +73,6 @@ module.exports = Hydrogen =
                 @handleKernelCommand command: 'restart-kernel'
             'hydrogen:shutdown-kernel': =>
                 @handleKernelCommand command: 'shutdown-kernel'
-            'hydrogen:copy-path-to-connection-file': =>
-                @copyPathToConnectionFile()
 
         @subscriptions.add atom.commands.add 'atom-workspace',
             'hydrogen:clear-results': => @clearResultBubbles()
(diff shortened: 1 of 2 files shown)


nikitakit:
Really happy to see a plugin interface start to develop!

In light of my experience integrating my `xdbg` package with hydrogen, I'd like to make a suggestion regarding the look of the API.

One idea that I think would work well is to expose kernel objects to the plugins, along with methods for getting access to these kernels (e.g. `getRunningKernel`, or `onKernelStart`). If we don't want to commit to maintaining a stable interface in the `kernel.coffee` file, it possible instead to create special kernel wrappers for the public API, that expose only the functionality we feel comfortable supporting long-term.

I mention this here because if we want to go with such a kernel-oriented interface, it might be more natural to make the get-connection-file an attribute of these kernel objects (instead of a toplevel function). This makes for a more consistent API, and also allows plugins that read connection files of kernels as they are spawned, and immediately act on that knowledge.

nikitakit:
@lgeiger I've opened #446 as a proof-of-concept of what I said in the previous comment.

lgeiger:
Closing this in favor of #446

PR #446, opened 2016-09-18
Wrapper-based plugin API

nikitakit:
Here is my take on a possible way to implement a plugin API for hydrogen.

Thanks to @lgeiger and @fasiha for looking into plugin solutions beforehand! 

This PR is influenced by my experience extending hydrogen (where extend=fork because at the time I didn't know about how providers/consumers work in atom). The idea is to have wrapper classes that add a layer of indirection for plugins. This leads to slightly more complex code on hydrogen's end, but it formalizes the API commitments that Hydrogen is willing to make to plugin developers.

The code here is a sketch of the general idea I have, and I certainly don't want to suggest that this be merged without prior discussion of the direction the project wants to take with respect to plugins.

I've also added a few `emitter` objects throughout hydrogen, which to the best of my knowledge are the canonical way to implement events in atom. The idea is to demonstrate how the API can expand to allow plugins to subscribe to events. (For example, one of my own use-cases involves being able to detect when some code is about to be executed in a kernel, and react to that event)

### DIFF
--- a/lib/kernel.coffee
+++ b/lib/kernel.coffee
@@ -5,9 +5,12 @@ _ = require 'lodash'
 jmp = require 'jmp'
 uuid = require 'uuid'
 zmq = jmp.zmq
+{Emitter} = require 'event-kit'
 
 StatusView = require './status-view'
 WatchSidebar = require './watch-sidebar'
+
+HydrogenKernel = require './plugin-api/hydrogen-kernel'
 
 module.exports =
 class Kernel
@@ -17,6 +20,15 @@ class Kernel
         @watchSidebar = new WatchSidebar this
         @statusView = new StatusView @kernelSpec.display_name
 
+        @emitter = new Emitter()
+
+        @pluginWrapper = null
+
+    getPluginWrapper: ->
+        unless @pluginWrapper?
+            @pluginWrapper = HydrogenKernel(this)
+
+        return @pluginWrapper
 
     addWatchCallback: (watchCallback) ->
         @watchCallbacks.push(watchCallback)
@@ -199,3 +211,7 @@ class Kernel
 
     destroy: ->
         console.log 'Kernel: Destroying base kernel'
+        if @pluginWrapper
+            @pluginWrapper.destroyed = true
+        @emitter.emit 'did-destroy'
+        @emitter.dispose()
--- a/lib/kernel.coffee
+++ b/lib/kernel.coffee
@@ -1,3 +1,5 @@
+{Emitter} = require 'atom'
+
 child_process = require 'child_process'
 path = require 'path'
 
@@ -5,11 +7,9 @@ _ = require 'lodash'
 jmp = require 'jmp'
 uuid = require 'uuid'
 zmq = jmp.zmq
-{Emitter} = require 'event-kit'
 
 StatusView = require './status-view'
 WatchSidebar = require './watch-sidebar'
-
 HydrogenKernel = require './plugin-api/hydrogen-kernel'
 
 module.exports =
--- a/lib/kernel.coffee
+++ b/lib/kernel.coffee
@@ -26,7 +26,7 @@ class Kernel
 
     getPluginWrapper: ->
         unless @pluginWrapper?
-            @pluginWrapper = HydrogenKernel(this)
+            @pluginWrapper = new HydrogenKernel(this)
 
         return @pluginWrapper
 
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -1,5 +1,6 @@
 # coffeelint: disable = missing_fat_arrows
 {CompositeDisposable} = require 'atom'
+{Emitter} = require 'event-kit'
 
 _ = require 'lodash'
 
@@ -13,6 +14,8 @@ Config = require './config'
 KernelManager = require './kernel-manager'
 Inspector = require './inspector'
 AutocompleteProvider = require './autocomplete-provider'
+
+HydrogenProvider = require './plugin-api/hydrogen-provider'
 
 module.exports = Hydrogen =
     config: Config.schema
@@ -32,6 +35,7 @@ module.exports = Hydrogen =
     watchSidebarIsVisible: false
 
     activate: (state) ->
+        @emitter = new Emitter()
         @kernelManager = new KernelManager()
         @inspector = new Inspector @kernelManager
         @codeManager = new CodeManager()
@@ -83,11 +87,19 @@ module.exports = Hydrogen =
             if item and item is atom.workspace.getActiveTextEditor()
                 @onEditorChanged item
 
+        @hydrogenProvider = null
+
 
     deactivate: ->
         @subscriptions.dispose()
         @kernelManager.destroy()
         @statusBarTile.destroy()
+
+    provideHydrogen: ->
+        unless @hydrogenProvider?
+            @hydrogenProvider = new HydrogenProvider(this)
+
+        return @hydrogenProvider
 
 
     consumeStatusBar: (statusBar) ->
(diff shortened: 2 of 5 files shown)


lgeiger:
This is interesting! I have thought about doing something similar.

@nikitaki, @fasiha, @n-riesco, @rgbkrk There are now a few proof of concept PR showing what can be done with a plugin API. I think we should discuss this in the next ntetact dev meeting in detail. Maybe we can even come up with a general spec to be used by `nteract` as well. The next meeting will probably on Monday the 26. at 9am pacific central time (I hope this is correct).

rgbkrk:
Sure, we can chat about it at the dev meeting. Next meeting is Monday the 26 at 9 am central. nteract/nteract has the beginnings of a plugin system, though it will likely form on top of our redux layer and be fairly similar to hyperterm.

nikitakit:
I can make it to the meeting next Monday. (Though you'll have to let me know what the meeting format is / how do I join)

rgbkrk:
People what they're going to do in the next 2 weeks as well as what they did in the past 2 weeks, if they want to. Discussions spawn naturally from there as a bit of cross-pollination amongst the various projects.

nikitakit:
@lgeiger Nice catch about requiring emitter from atom. I recall seeing it in the docs, but I wasn't sure. I think this also means that event-kit can be removed from the package.json

fasiha:
From reading over the code (as an Atom/CoffeeScript noob), I understand that this is a far more general and flexible approach than the one in nteract/hydrogen/pull/441, super-awesome!

I can’t quite tell what a consumer plugin would look like—would it (wild guess) use the `onDidChangeKernel` in `hydrogen-provider.coffee` to somehow set up its own listener?

Or does the consumer basically have complete access to the kernel object and do whatever it wants to it 😮?

lgeiger:
Oh wow pushing to forks from a pull request is strange but amazing 😂 .

I'm looking forward seeing you in the meeting. There will be an general announcement in our slack channel. They are held via appear.in in the [nteract](https://appear.in/nteract) room. You can take a look at the [notes from last time](https://github.com/nteract/meeting-minutes).

nikitakit:
@fasiha From the consumer side, it depends on whether the plugin responds to direct user commands, or events from hydrogen.

For the first case (e.g. reading connection files to spawn a connected console), a plugin can look up the current kernel, get its connection file, and then work with it.

For the second case, the plugin can listen to `onDidChangeKernel` and then register events for that kernel (e.g. use something like `onDidGetResult`, which does not exist yet, to read the results and paste them inline into the document). I was also considering adding a way to register event listeners for all kernels (present and yet-to-be-spawned), but that part is not fleshed out yet.

The idea is that the consumer has complete access to the kernel wrapper object (called `HydrogenKernel` here), which only exposes APIs we consider stable. So the range of modifications a consumer can make is very large, but we still make a distinction between internal and external interfaces.

lgeiger:
I tried using this PR but Hydrogen provides only
bildschirmfoto 2016-09-20 um 11 05 30
Any idea why?

lgeiger:
I just had a look at the atom `status-bar` package. They only provide the relevant methods of the [ `status-bar-view` class](https://github.com/atom/status-bar/blob/master/lib/status-bar-view.coffee) like [this](https://github.com/atom/status-bar/blob/master/lib/main.coffee#L64).
This would mean we could get rid of the additional wrapper while still only exposing the methods we are comfortable sharing.

nikitakit:
@lgeiger The image in your first comment is as expected, because HydrogenProvider only has two properties. All of the methods are hidden under `__proto__`, because that's how classes work in javascript. Calling methods on the object should still work.

nikitakit:
Regarding the way the status-bar package exposes a subset of methods, it's certainly possible to do so as well. I don't see declaring a wrapper inline as being that different from making a class. It would be pretty straightforward to do so for the main hydrogen class, but less so if we want to wrap kernels.

One thing to consider, if you look at the `HydrogenKernel` wrapper proposal, is that sometimes plugin calls need to perform additional validation that is not used internally. In particular, hydrogen can guarantee that once a kernel is not in use, all references to the `Kernel` object are removed. On the other hand, the plugin API might need to perform additional validation to make sure that plugins can't use already-destroyed kernel objects to corrupt hydrogen's internal state. Depending on how much extra validation we want to do, the wrapper class approach may be cleaner than doing inline wrapper creation with `bind`.

We can discuss tradeoffs at the meeting.

lgeiger:
> @lgeiger The image in your first comment is as expected, because HydrogenProvider only has two properties. All of the methods are hidden under __proto__, because that's how classes work in javascript. Calling methods on the object should still work.

Sorry for the dumb question. This exposed my lack of knowledge regarding javascript 😂 

But the issue still exists. I'm trying to use it [here](https://github.com/lgeiger/hydrogen-launcher/blob/wrapper-plugin-api/lib/main.coffee#L46) and [here](https://github.com/lgeiger/hydrogen-launcher/blob/wrapper-plugin-api/lib/main.coffee#L77) but `getActiveKernel` always returns undefined. I think I'm using it wrong but couldn't get it to work.

nikitakit:
@lgeiger Don't worry about it!

I actually have no idea why you're having this issue. I tested a version of your launcher plugin on my machine, and it worked fine. The code is at https://github.com/nikitakit/hydrogen-launcher/tree/wrapper-api (note that I had some issues with term-launcher, so I just tested the other parts of the plugin)

My best guess would be that you either didn't spawn a kernel prior to calling the plugin, or that you're somehow still using your own version of the plugin api code. (Atom also does this weird thing that dev-mode plugins are gone if you forgot to start atom in dev mode)

What is the issue precisely?
* `getActiveKernel` is undefined (i.e. you can't call it at all)
* `getActiveKernel` returns undefined (should never happen)
* `getActiveKernel` returns null (should happen whenever there is no active kernel yet)

lgeiger:
Thanks for testing!

I used your code and get the following error: `Uncaught TypeError: Cannot read property 'getConnectionFile' of undefined`

I also checked to make sure both packages are installed correctly. To reproduce this issue I used the following steps on my machine:
- Restart atom
- Optional: Run `Hydrogen Launcher: Launch Jupyter Console` and see the error notification because no kernel is running.
- Run some code in Hydrogen to start a kernel
- Run `Hydrogen Launcher: Launch Jupyter Console`

lgeiger:
I figured it out. I'll push a quick fix.

lgeiger:
It's strange that the missing `new` operator worked on your machine but not for me.

It would be great if you could open a issue at the `term-launcher` repo and briefly describe your problem. I did just a rough test on linux and windows also you're probably the first on using it :smile:.

nikitakit:
@lgeiger Thanks for fixing this!

(Now it's my turn to be embarrassed by poor knowledge of javascript, because this is definitely not the first time I've forgotten `new`)

Not sure why it worked for me before. Maybe I refactored after testing, and that ended up breaking things. I'll get back to you on term-launcher, since I found in retrospect that I messed up the package.json, which could have caused the problem.

rgbkrk:
While totally a meta on the missing `new` (I don't recommend doing this), you can return a properly constructed `new`ed object as part of the function constructor:

```js
function X() { 
  if (! (this instanceof X)) {
    return new X(arguments);
  }
} 
```

Which results in a proper `new X()` to be setup. :wink:

nikitakit:
So this plugin situation has been at a standstill for quite a while...

It really seems that this is a "perfect is the enemy of the good" situation. Any chance we can just merge one of the proposals to move things forward? It's also possible to admit that the project lacks manpower to make any further progress on plugins, and close these issues.

nikitakit:
Could someone else hit the merge button on this one?

rgbkrk:
Here we go!

PR #448, opened 2016-09-18
add info on setting Visual Studio version in apm to troubleshooting

mmagnuski:
Let me know if this is ok.
I can also add a few lines about setting apm python version to conda env (I am using python 3 normally, so I created a conda environment with python 2 for hydrogen installation) if you think it would be useful.
As mentioned in the issue, I have documented my adventures with installing hydrogen [here](https://github.com/mmagnuski/mozg-rozpajton/blob/rey/install_hydrogen.md) - they're targeted at complete beginners so may be too explicit in some places (but I've had people typing pip commands in python REPL before so I want to be overly explicit).

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -13,6 +13,26 @@ There are a number of possible causes and solutions:
   [here](https://github.com/nodejs/node-gyp#installation) for installation
   instructions.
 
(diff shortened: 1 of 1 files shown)


lgeiger:
Thanks for taking the time! You can add the link to your beginner tutorial to the troubleshooting guide as well. This is really helpful for beginners.

PR #470, opened 2016-10-12
Fix cell separation 

### LINKED ISSUE #469: Cell separation on line 1 of file breaks things
Having a cell separator on the first line of a file breaks the cell-running functionality of hydrogen.

Consider the following file:
```python
# %%
print('cell 1')
# %%
print('cell 2')
# %%
print('cell 3')
```

Instead of being able to run the three cells individually, only the third cell is run no matter where the cursor is placed in the file.

(I'm currently using master branch, with Atom 1.10.2 on Mac)

lgeiger:
Fixes #469 

/cc @nikitakit

### DIFF
--- a/lib/code-manager.coffee
+++ b/lib/code-manager.coffee
@@ -126,9 +126,9 @@ class CodeManager
             return [start, end]
 
         regex = new RegExp regexString
-        cursor = @editor.getLastCursor().getBufferPosition()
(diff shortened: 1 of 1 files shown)


rgbkrk:
I'd merge because it looks like the right fix, but we can wait for feedback. Let me know.

lgeiger:
This can be merged. `@isComment` takes only one argument.

PR #472, opened 2016-10-12
Display execution counts for result bubbles

nikitakit:
This addresses #331

I'm still looking for UI suggestions, because the way the execution count is displayed here is super clunky.

### DIFF
--- a/lib/kernel.coffee
+++ b/lib/kernel.coffee
@@ -75,6 +75,9 @@ class Kernel
         unless result?
             result = @_parseStreamIOMessage message
 
+        unless result?
+            result = @_parseExecuteInputIOMessage message
+
         return result
 
 
@@ -208,6 +211,14 @@ class Kernel
 
         return result
 
+    _parseExecuteInputIOMessage: (message) ->
+        if message.header.msg_type is 'execute_input'
+            result =
+                data: message.content.execution_count
+                type: 'number'
+                stream: 'execution_count'
+
+        return result
 
     destroy: ->
         console.log 'Kernel: Destroying base kernel'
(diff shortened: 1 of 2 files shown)


lgeiger:
What about displaying the execution count somewhere in the status bar? Though that would only display the latest execution count.

nikitakit:
Status bar might be useful, though I'm not sure offhand what kinds of workflows it would enable. I have no idea to what extent others use execution counts, but for my own usecase they probably aren't enough.

For reference, I should explain my workflow:
* As context, do a lot of numeric work. If a cell of code runs without exceptions, that says nothing about whether it actually did the operation I want
* I use the pretty-printer for cell outputs to check whether things work as expected (sometimes running the cell many times after slight tweaks)
* Now I have output that's not assigned to a named variable that I might want to use later in a pipeline. I use the `_###` numbered variables provided by IPython extensively
* I often need to reference previous cells many executions back, and many of the quantities there don't have named variables. In the notebook I just consult the execution count for the cell. (This is where the status-bar approach breaks down)
* Once I have a pipeline working end-to-end, I delete any extraneous cells and spend some effort to come up with meaningful variable names

nikitakit:
Does anyone have additional thoughts on this change? (In particular, I'm not sure that "link" is the best icon, nor about icon ordering in the sidebar)

If no one is opposed, I'd like to propose this for merging as-is. This feature is very important for my usage, and I'd like to go back to using the vanilla master branch.

rgbkrk:
The only part I find a bit invasive is the automatic write to clipboard. Other than that, I think this is good to go.

lgeiger:
@nikitakit What do you think about the idea I outlined above, would it work for you?

I really want to get this into 1.0.0 🚀 
I'm happy to implement it, if you're busy at the moment.

nikitakit:
@lgeiger I was just updating this PR as you commented! (It took me a while to verify that tooltips accept a function for the title option)

I really like your suggestion; the extra button was definitely awkward.

Issue #461, opened 2016-09-30
"node-gyp rebuild fails" on my system


Nodragem:
I received that error after trying to install Hydrogen:
```
zmq@2.15.3 install /tmp/apm-install-dir-116830-27598-e9hmcd/node_modules/Hydrogen/node_modules/jmp/node_modules/zmq
> node-gyp rebuild

make: Entering directory `/tmp/apm-install-dir-116830-27598-e9hmcd/node_modules/Hydrogen/node_modules/jmp/node_modules/zmq/build'
  CXX(target) Release/obj.target/zmq/binding.o
make: Leaving directory `/tmp/apm-install-dir-116830-27598-e9hmcd/node_modules/Hydrogen/node_modules/jmp/node_modules/zmq/build'

../binding.cc:28:17: fatal error: zmq.h: No such file or directory
 #include <zmq.h>
                 ^
compilation terminated.
make: *** [Release/obj.target/zmq/binding.o] Error 1
gyp ERR! build error 
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/home/c1248317/MyApps/atom-1.9.8-amd64/resources/app/apm/node_modules/npm/node_modules/node-gyp/lib/build.js:269:23)
gyp ERR! stack     at ChildProcess.emit (events.js:98:17)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (child_process.js:820:12)
gyp ERR! System Linux 3.10.0-327.3.1.el7.x86_64
gyp ERR! command "node" "/home/c1248317/MyApps/atom-1.9.8-amd64/resources/app/apm/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
gyp ERR! cwd /tmp/apm-install-dir-116830-27598-e9hmcd/node_modules/Hydrogen/node_modules/jmp/node_modules/zmq
gyp ERR! node -v v0.10.40
gyp ERR! node-gyp -v v2.0.2
gyp ERR! not ok 
npm ERR! Linux 3.10.0-327.3.1.el7.x86_64
[...]

Issue #454, opened 2016-09-25
Refresh installation instructions with zmq-prebuilt


willingc:
As in #417, `zmq-prebuilt` can simplify installation of Hydrogen on all operating systems. We can simplify the install steps for users.

Issue #445, opened 2016-09-16
Hydrogen fails to install on Atom Mac OS


Chibuikem709:
I am trying to download Hydrogen onto Atom but I keep getting this error:
>Installing “Hydrogen@0.14.0” failed.Hide output…

> zmq@2.15.3 install /private/var/folders/ny/jd4d9h_90hv7l27ykfmg141m0000gn/T/apm-install-dir-116816-56245-1miui9c/node_modules/Hydrogen/node_modules/zmq
> node-gyp rebuild

>/private/var/folders/ny/jd4d9h_90hv7l27ykfmg141m0000gn/T/apm-install-dir-116816-56245-1miui9c
`-- (empty)

>Package libzmq was not found in the pkg-config search path.
>Perhaps you should add the directory containing `libzmq.pc'
>to the PKG_CONFIG_PATH environment variable
>No package 'libzmq' found
>gyp: Call to 'pkg-config libzmq --libs' returned exit status 1 while in binding.gyp. while trying to >load binding.gyp
>gyp ERR! configure error 
>gyp ERR! stack Error: `gyp` failed with exit code: 1
>gyp ERR! stack     at ChildProcess.onCpExit >(/Users/ChibuikemNwizu/Downloads/Atom.app/Contents/Resources/app/apm/node_modules/node-gyp/lib/configure.js:305:16)
>gyp ERR! stack     at emitTwo (events.js:87:13)
>gyp ERR! stack     at ChildProcess.emit (events.js:172:7)
>gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:200:12)
>gyp ERR! System Darwin 15.6.0
>gyp ERR! command >"/Users/ChibuikemNwizu/Downloads/Atom.app/Contents/Resources/app/apm/bin/node" >"/Users/ChibuikemNwizu/Downloads/Atom.app/Contents/Resources/app/apm/node_modules/.bin/node-gyp" "rebuild"
gyp ERR! cwd /private/var/folders/ny/jd4d9h_90hv7l27ykfmg141m0000gn/T/apm-install-dir-116816-56245-1miui9c/node_modules/Hydrogen/node_modules/zmq
>gyp ERR! node -v v4.4.5
>gyp ERR! node-gyp -v v3.4.0
gyp ERR! not ok 
[...]

Issue #440, opened 2016-09-11
Update installation instructions for windows


lgeiger:
I just discovered that Microsoft now provides a standalone C++ compiler without the need for visual studio:
http://landinghub.visualstudio.com/visual-cpp-build-tools

I haven't tested this yet, but I'm pretty confident.

lgeiger:
https://github.com/Microsoft/nodejs-guidelines/blob/master/windows-environment.md#prerequisites

mmagnuski:
I can confirm that standalone C++ compiler works well when installing hydrogen on windows. :)

PR #476, opened 2016-10-15
Use zmq-prebuilt

lgeiger:
This will bump `jmp` and switch to `zmq-prebuilt`. Any help with testing is really appreciated. The prerequisites can be found [here](https://github.com/lgeiger/hydrogen/blob/15fd7836371a024299290a2778a0180c12ce0c5c/README.md#dependencies).

**TODO:**
- Tested on:
  - [x] OS X: Atom beta
  - [x] OS X: Atom stable
  - [x] Windows 10 (64-bit): Atom stable
  - [x] Ubuntu 16.04 (64-bit): Atom stable
- Docs:
  - [x] Update README
  - [x] Update TROUBLESHOOTING
- Travis:
  - [x] Remove unnecessary dependencies
  - [x] Move to Linux for faster builds

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,9 @@
       "hydrogen:run-cell-and-move-down"
     ]
   },
-  "scripts": {},
+  "scripts": {
+    "install": "npm rebuild --build-from-source"
+  },
   "repository": "https://github.com/nteract/hydrogen",
   "license": "MIT",
   "engines": {
@@ -39,7 +41,7 @@
     "atom-message-panel": "^1.2.4",
     "atom-space-pen-views": "^2.0.5",
     "escape-string-regexp": "^1.0.5",
-    "jmp": "0.4.x",
+    "jmp": "0.6.x",
     "jupyter-js-services": "^0.18.1",
     "lodash": "^4.14.0",
     "requirejs": "^2.2.0",
@@ -64,10 +66,10 @@
         "2.0.0": "provide"
       }
     },
-    "hydrogen.provider" : {
-        "versions": {
-            "0.0.2": "provideHydrogen"
-        }
+    "hydrogen.provider": {
+      "versions": {
+        "0.0.2": "provideHydrogen"
+      }
     }
   },
   "devDependencies": {
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,25 +6,10 @@ notifications:
 
 script: 'curl -s https://raw.githubusercontent.com/atom/ci/master/build-package.sh | sh'
 
-os:
-  - osx
-
 env:
-  global:
-    - APM_TEST_PACKAGES=""
-
   matrix:
-    - ATOM_CHANNEL=stable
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -29,7 +29,9 @@
       "hydrogen:run-cell-and-move-down"
     ]
   },
-  "scripts": {},
+  "scripts": {
+    "install": "npm rebuild --build-from-source"
```
lgeiger: This will skip the prebuilt goodness and ensure that `zmq-prebuilt` is build from source. This flag is used by `prebuild`.

I may make a PR to `apm` to set this by default.

lgeiger:
- Closes #37
- Closes #445
- Closes #440
- Closes #454
- Closes #459
- Closes #461

:point_up: :open_mouth: 🎉

package.json
rgbkrk: This still installs all the dependencies too when you override install?

lgeiger: Good point I think I only removed `node_modules/zmq-prebuilt/` during testing. I test this.

lgeiger: `rm -rf node_modules` and `apm install` works 👍 

I override the install option only because of the `--build-from-source` flag.

rgbkrk:
> ☝️ 😮 🎉

Excellent emoji choice.

rgbkrk:
@nikitakit can you give this a spin?

lgeiger:
I just realized all my test machines have Node.js installed. 

We use a node script to download ØMQ on Linux and Mac. It would be great if someone could test this without a Node.js setup and see if `apm` uses the bundled node from Atom.

lgeiger:
Yeah Atom is smart and uses it's bundled node to execute the script. So there shouldn't be any problems.

nikitakit:
@rgbkrk Sorry I've been away lately, just got around to looking at this.

What commands am I supposed to run to test this?

I tried just running "apm install" from terminal, and the resulting build of Hydrogen works. But it still requires me to specify python2 via an env var, and needs to run gyp / rebuild from source. My impression is that the whole point of testing this is to make sure the prebuilt code is OK.

I'm on Mac, with Atom 1.10.2.

rgbkrk:
Since Atom will be on later Electron releases that aren't supported by prebuild, we end up always having to do a rebuild. 😢  However, this setup is simpler in that it does link statically and requires a lot less dependencies on the end user. They are still stuck needing Python2 (for the build).

lgeiger:
@nikitakit Thanks a lot for testing.
Yeah we still need to build from source and therefor are stuck with Python 2 because of node-gyp.
Since prebuild doesn't support building against Electron headers and Electrons ABI doesn't match with Nodes, this is our only option for now.

package.json
lgeiger: I totally forgot about this. I made a PR: https://github.com/atom/apm/pull/638

PR #479, opened 2016-10-18
Revert "CodeManager: Don't move the cursor down"

### LINKED ISSUE #443: Fix running last line and moving down
We want to support two use cases:

1. user writes a program by typing lines and executing them straight away.

2. user wants to keep a large amount of comments at the end of the file (see #433 for more details).

### LINKED ISSUE #475: Run and Move Down: Add line and move if no executable code left
I'm really blown away by Hydrogen and what it can do. I am coming to Python from Matlab and being able to see my outputs more easily is such a great way to code.

But my workflow seems to get hampered by the "Run and Move Down" command. If I'm running code as I'm writing it down, then I find myself executing the last line quite often. If the last line is also the last line in the document, then using "Run and Move Down" runs my code (obviously), generates a plot, but then erases that plot as soon as I insert a new line. If there is a blank line below my code, then it runs, and I can arrow down to continue coding without the plot disappearing; but this is not as ideal as my cursor already being down there.

So if it makes sense, I would love for "Run and Move Down" to pop down to the next blank line if there is no further executable code. And if there is no blank line, it would be great if it could add one and pop me down there (without causing the output to disappear). If anyone doesn't want this capability, then they could always use the regular Run instead, and it won't move.

Thanks for reading!

lgeiger:
This reverts #433 for now
Closes #475
Closes #443 

/cc @n-riesco @mryumyum

### DIFF
--- a/lib/code-manager.coffee
+++ b/lib/code-manager.coffee
@@ -209,12 +209,14 @@ class CodeManager
     moveDown: (row) ->
         lastRow = @editor.getLastBufferRow()
 
+        if row >= lastRow
+            @editor.moveToBottom()
+            @editor.insertNewline()
+            return
+
         while row < lastRow
             row++
             break if not @isBlank(row)
-
-        unless row < lastRow
-            return
 
         @editor.setCursorBufferPosition
             row: row


lgeiger:
We had a few bugfixes and introduced a initial version of the plugin API.
Should we ship `v0.15.0` now (before we merge #476)?

rgbkrk:
Sounds great.

lgeiger:
Shipped

PR #488, opened 2016-10-31
Make keybindings consistent with nteract and jupyter notebook

### LINKED ISSUE #304: Hydrogen keybindings
Moving the discussion from [here](https://github.com/nteract/hydrogen/pull/227#issuecomment-202116496) and [here](https://github.com/nteract/hydrogen/pull/227#issuecomment-202126973).

---

Shall we add a setting to enable the keybindings that we would like to introduce in Hydrogen v1.0.0?

? `hydrogen:inspect`

C-enter `hydrogen:run`
S-enter `hydrogen:run-and-move-down`

A-C-enter `hydrogen:run-cell`
A-S-enter `hydrogen:run-cell-and-move-down`

? `hydrogen:add-break-point`
S-backspace `hydrogen:remove-latest-break-point`
S-A-backspace `hydrogen:remove-all-break-points`

? `hydrogen:clear-results`

A-S-C-enter `hydrogen:run-all`

---

Currently, we're shadowing

S-C-enter `hydrogen:run-and-move-down` `editor:newline-above`
C-backspace `hydrogen:clear-results` `editor:delete-to-beginning-of-word`

lgeiger:
Since the zeromq install issues are solved. We're getting closer to `1.0.0`.
This is the right time to make the keybinding consistent with `nteract` and `jupyter notebook`.

Please try it and let me know what you think.
Closes #304

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ It's easiest to see these interactions visually:
 
 **"Hydrogen: Run And Move Down"** will run the the code as described above and move the cursor to the next executable line.
 
-If your code starts getting cluttered up with results, run **"Hydrogen: Clear Results"** to remove them all at once. You can also run this command with ⌘-⌥-⌫.
+If your code starts getting cluttered up with results, run **"Hydrogen: Clear Results"** to remove them all at once.
 
 ### "Hydrogen: Run Cell"
 A "code cell" is a block of lines to be executed at once. You can define them using inline comments. Hydrogen supports a
@@ -169,7 +169,7 @@ After you've run some code with Hydrogen, you can use the **"Hydrogen: Toggle Wa
 **IMPORTANT:** Be careful what you put in your watch expressions. If you write code that mutates state in a watch expression, that code will get run after every execute command and likely result in some _extremely confusing_ bugs.
 
 
-You can re-run the watch expressions by using the normal run shortcut (⌘-⌥-↩ by default) inside a watch expression's edit field.
+You can re-run the watch expressions by using the normal run shortcut (⌘-↩ by default) inside a watch expression's edit field.
 
 If you have multiple kernels running, you can switch between their watch expressions with the **"Hydrogen: Select Watch Kernel"** command (or just click on the "Kernel: <language>" text).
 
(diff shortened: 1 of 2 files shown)


nikitakit:
I notice that this still hasn't been merged yet. I originally wasn't going to comment because I'm not sure how substantive this is, but looks like this has remained unmerged anyway.

I agree that compatibility is a major consideration for the default keybinding set. Changes look good in this regard. Only thing I can point out to this regard is that on my Jupyter Notebook running in Mac Firefox, running a cell is bound to ctrl-enter and not cmd-enter.

In terms of my own usage, chances are I won't keep these bindings. (But I don't think I'm representative in the way I configure things, e.g. I use a Programmer Dvorak keyboard layout.) I personally only use two hydrogen commands: run, and run-cell-and-move-down. Currently I have them bound to shift-enter and ctrl-enter, but I should really remap one to cmd-enter. I find that having to distinguish between move-down and dont-move-down variants of the same command is too much of a cognitive burden and that it's much easier to stick to one and then navigate my way to the correct location. Unlike the notebook, atom has solid non-modal text navigation, so the variants don't add as much value. On the other hand I use cells a lot, so a simpler combo for them makes sense.

So anyway, these are my thoughts. No real point/punchline here.

lgeiger:
Thanks for commenting!
You're right, Atom makes it really easy for anyone to change keybindings. So this PR primarily reduces the barrier of entry for new users coming from notebooks.

 > Only thing I can point out to this regard is that on my Jupyter Notebook running in Mac Firefox, running a cell is bound to ctrl-enter and not cmd-enter.

I personally like to stay with `cmd-enter` since it's the default function key on mac. For me it just feels more native so I don't mind the inconsistency with jupyter there. 😉

rgbkrk:
Yeah I wish the jupyter binding was cmd-enter too. If I had to imagine, that might have been a browser limitation? I'm not sure.

lgeiger:
Anyone for a merge, so we can release 1.0.0?

rgbkrk:
:shipit:

PR #490, opened 2016-11-02
Remove copyPathToConnectionFile

lgeiger:
This removes `@copyPathToConnectionFile()` because the functionality is now provided by [Hydrogen Launcher](https://github.com/lgeiger/hydrogen-launcher).

The error handling was moved to the plugin provider.

### DIFF
--- a/lib/plugin-api/hydrogen-kernel.coffee
+++ b/lib/plugin-api/hydrogen-kernel.coffee
@@ -20,7 +20,6 @@ class HydrogenKernel
         @_assertNotDestroyed()
         connectionFile = @_kernel.connectionFile
         unless connectionFile?
-            atom.notifications.addError "No connection file for
-                #{@_kernel.kernelSpec.display_name} kernel found"
+            throw new Error "No connection file for #{@_kernel.kernelSpec.display_name} kernel found"
             return null
         return connectionFile
--- a/lib/plugin-api/hydrogen-kernel.coffee
+++ b/lib/plugin-api/hydrogen-kernel.coffee
@@ -21,5 +21,5 @@ class HydrogenKernel
         connectionFile = @_kernel.connectionFile
         unless connectionFile?
             throw new Error "No connection file for #{@_kernel.kernelSpec.display_name} kernel found"
-            return null
+
         return connectionFile
--- a/lib/plugin-api/hydrogen-kernel.coffee
+++ b/lib/plugin-api/hydrogen-kernel.coffee
@@ -20,7 +20,7 @@ class HydrogenKernel
         @_assertNotDestroyed()
         connectionFile = @_kernel.connectionFile
         unless connectionFile?
-            # Standardize on returning null if there is no connection file
-            # (e.g. for WSKernel)
+            atom.notifications.addError "No connection file for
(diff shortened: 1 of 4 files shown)


nikitakit:
:+1: on removing duplicate functionality.

I'm personally not sure about folding in the error handling. If a plugin has fallback behavior capable of handling those cases, it may want to suppress the error messages or replace them with messages of its own. (I've been bitten by a similar issue with `jupyter-js-services` in hydrogen, which still streams access denied errors to the atom dev console during the course of normal usage. Though thankfully it's just the dev console, not the notifications.)

Some alternatives to consider:
* Delegate error handling to plugins
* Methods that display error notifications accept something like a `suppressErrors`/`quiet` flag
* Don't display the notifications, but still provide some way of returning precise error text to plugins (e.g. throw an error that has description fields designed for use in notifications)

Any thoughts?

lgeiger:
@nikitakit Thanks for the feedback!
You're right, I think the easiest thing is to throw the error and catch it inside the plugin.

By the way, the plugin wrapper is super nice to use 👍 👍

lib/plugin-api/hydrogen-kernel.coffee
```diff
@@ -20,7 +20,6 @@ class HydrogenKernel
         @_assertNotDestroyed()
         connectionFile = @_kernel.connectionFile
         unless connectionFile?
-            # Standardize on returning null if there is no connection file
-            # (e.g. for WSKernel)
+            throw new Error "No connection file for #{@_kernel.kernelSpec.display_name} kernel found"
             return null
```
nikitakit: No need for return; it's misleading to put it right after a throw

lib/plugin-api/hydrogen-provider.coffee
```diff
@@ -11,7 +11,10 @@ class HydrogenProvider
                 callback null
 
     getActiveKernel: ->
-        unless @_hydrogen.kernel
+        unless @_hydrogen.kernel?
+            grammar = @_hydrogen.editor.getGrammar()
+            language = @_hydrogen.kernelManager.getLanguageFor grammar
+            throw new Error "No running kernel for language `#{language}` found"
             return null
```
nikitakit: Same

lib/plugin-api/hydrogen-kernel.coffee
lgeiger: Good catch! Fixed 👍

rgbkrk:
Thanks @nikitakit, your approval is a merge decree for me.

lgeiger:
@nikitakit Oops, you probably missed the [discussion about 2FA](https://github.com/nteract/nteract/issues/760).

I'd be really happy if you set up 2FA so I can add to the org again.
Are there any reasons preventing you from using 2FA?

rgbkrk:
@nikitakit didn't miss the discussion, we talked out-of-band in email about it. He's very much still welcome here.

Amusingly, one of the reasons we switched to 2FA was because of our automated posting of binaries for ZMQ and now we're on an org (zeromq) that doesn't have 2FA across the board. 
¯\\\_(ツ)\_/¯

PR #503, opened 2016-11-15
Move to ES6

### LINKED ISSUE #494: Move to ES6
I would like to (re)evaluate if we should port our code base from CoffeeScript to ES6. Atom is slowly moving away from CoffeeScript as well.

My main motivation is to move Hydrogen closer to nteract and be able to use all ES6 features.

I just made a quick [example plugin for Hydrogen](https://github.com/lgeiger/hydrogen-example-plugin) in ES6 and was really surprised how easy it was.
We don't need a build step, we only need to add `'use babel';` on top of the file and Atom takes care of the rest.

I don't feel strongly against CoffeeScript but if think we should have this discussion.
@rgbkrk @n-riesco @nikitakit What do you think?

lgeiger:
Fixes #494

This will port all coffeescript code to ES6. There will be probably a few bugs since I didn't test all functionality.


lib/autocomplete-provider.js
```diff
@@ -1,66 +1,83 @@
...
+        // The default provider has a priority of 0.
+        inclusionPriority: 1,
+        excludeLowerPriority: false,
+
+        // Required: Return a promise, an array of suggestions, or null.
+        getSuggestions({editor, bufferPosition, scopeDescriptor, prefix}) {
+            let grammar = editor.getGrammar();
+            let language = kernelManager.getLanguageFor(grammar);
+            let kernel = kernelManager.getRunningKernelFor(language);
+            if (kernel == null) {
```
n-riesco: `if(!kernel)`

lib/code-manager.js
```diff
@@ -1,223 +1,279 @@
...
+        let cursor = this.editor.getLastCursor();
+
+        let row = cursor.getBufferRow();
+        console.log('findCodeBlock:', row);
+
+        let indentLevel = cursor.getIndentLevel();
+
+        let foldable = this.editor.isFoldableAtBufferRow(row);
+        let foldRange = this.editor.languageMode.rowRangeForCodeFoldAtBufferRow(row);
+        if ((foldRange == null) || (foldRange[0] == null) || (foldRange[1] == null)) {
```
n-riesco: I guess this is how decaffeinate deals with `?`. :disappointed:

lib/code-manager.js
```diff
@@ -1,223 +1,279 @@
...
+    getSelectedText() {
+        return this.normalizeString(this.editor.getSelectedText());
+    }
 
 
-    normalizeString: (code) ->
-        if code?
-            return code.replace /\r\n|\r/g, '\n'
+    normalizeString(code) {
+        if (code != null) {
```
n-riesco: ditto

lib/code-manager.js
```diff
@@ -1,223 +1,279 @@
...
 
-    getFoldRange: (row) ->
-        range = @editor.languageMode.rowRangeForCodeFoldAtBufferRow(row)
-        if @getRow(range[1] + 1)?.trim() is 'end'
-            range[1] = range[1] + 1
-        console.log 'getFoldRange:', range
-        return range
+    getFoldRange(row) {
+        let range = this.editor.languageMode.rowRangeForCodeFoldAtBufferRow(row);
+        if (__guard__(this.getRow(range[1] + 1), x => x.trim()) === 'end') {
```
n-riesco: Oh! So decaffeinate sometimes uses a `__guard__`...

lib/code-manager.js
```diff
@@ -1,223 +1,279 @@
...
-        regexString = @getRegexString @editor
+    getCurrentCell() {
+        let buffer = this.editor.getBuffer();
+        let start = buffer.getFirstPosition();
+        let end = buffer.getEndPosition();
+        let regexString = this.getRegexString(this.editor);
 
-        unless regexString?
-            return [start, end]
+        if (regexString == null) {
```
n-riesco: ditto (I'm not going to stop marking these bugs)

lib/code-manager.js
```diff
@@ -1,223 +1,279 @@
...
+}
+function __range__(left, right, inclusive) {
+  let range = [];
+  let ascending = left < right;
+  let end = !inclusive ? right : ascending ? right + 1 : right - 1;
+  for (let i = left; ascending ? i < end : i > end; ascending ? i++ : i--) {
+    range.push(i);
+  }
+  return range;
+}
```
n-riesco: missing EOL

lib/config.js
```diff
@@ -1,95 +1,54 @@
-_ = require 'lodash'
+let Config;
```
n-riesco: I reckon this line can go.

lib/kernel-manager.js
```diff
@@ -1,262 +1,326 @@
...
+import KernelPicker from './kernel-picker';
 
-module.exports =
-class KernelManager
-    constructor: ->
-        @_runningKernels = {}
-        @_kernelSpecs = @getKernelSpecsFromSettings()
+export default class KernelManager {
+    constructor() {
+        this.getAllKernelSpecs = this.getAllKernelSpecs.bind(this);
```
n-riesco: Why is this necessary?

lib/kernel-manager.js
```diff
@@ -1,262 +1,326 @@
...
+        this.getKernelSpecsFromJupyter = this.getKernelSpecsFromJupyter.bind(this);
```
n-riesco: and this?

lib/main.js
```diff
@@ -1,392 +1,519 @@
...
-
-
-    insertResultBubble: (row) ->
-        buffer = @editor.getBuffer()
-        lineLength = buffer.lineLengthForRow(row)
-
-        marker = @editor.markBufferPosition
-            row: row
+// coffeelint: disable = missing_fat_arrows
+let Hydrogen;
```
n-riesco: remove?

lib/main.js
```diff
@@ -1,392 +1,519 @@
...
+        this.statusBarElement.classList.add('hydrogen');
+        this.statusBarElement.classList.add('status-container');
+        this.statusBarElement.onclick = this.showKernelCommands.bind(this);
+
+        this.onEditorChanged(atom.workspace.getActiveTextEditor());
+
+        this.subscriptions = new CompositeDisposable;
+
+        this.subscriptions.add(atom.commands.add('atom-text-editor', {
+            ['hydrogen:run']: () => this.run(),
```
n-riesco: are the square brackets really necessary?

lib/plugin-api/hydrogen-provider.js
```diff
@@ -11,31 +11,38 @@
...
+    */
+    getActiveKernel() {
+        if (this._hydrogen.kernel == null) {
+            let grammar = this._hydrogen.editor.getGrammar();
+            let language = this._hydrogen.kernelManager.getLanguageFor(grammar);
+            throw new Error(`No running kernel for language \`${language}\` found`);
+        }
 
-        return @_hydrogen.kernel.getPluginWrapper()
+        return this._hydrogen.kernel.getPluginWrapper();
```
n-riesco: Is building a wrapper so expensive that a getter makes a difference?

lib/rename-view.js
```diff
@@ -1,40 +1,74 @@
...
+import { $, TextEditorView, View } from 'atom-space-pen-views';
+
+export default class RenameView extends View {
+    constructor(...args) {
+        super(...args);
+        this.confirm = this.confirm.bind(this);
+        this.cancel = this.cancel.bind(this);
+    }
+
+    static content(prompt) {
```
n-riesco: It doesn't make sense that this method is static. Static methods don't use `this`.

lib/rename-view.js
```diff
@@ -1,74 +1,72 @@
...
 export default class RenameView extends View {
-    constructor(...args) {
-        super(...args);
-        this.confirm = this.confirm.bind(this);
-        this.cancel = this.cancel.bind(this);
-    }
+  constructor(...args) {
+    super(...args);
+    this.confirm = this.confirm.bind(this);
+    this.cancel = this.cancel.bind(this);
```
n-riesco: and this?

lib/input-view.js
```diff
@@ -1,67 +1,66 @@
...
 
 export default class InputView extends View {
-    constructor(...args) {
-        super(...args);
-        this.confirm = this.confirm.bind(this);
-        this.close = this.close.bind(this);
-    }
+  constructor(...args) {
+    super(...args);
+    this.confirm = this.confirm.bind(this);
```
n-riesco: why is this necessary?

lib/input-view.js
```diff
@@ -1,67 +1,66 @@
...
+    this.close = this.close.bind(this);
```
n-riesco: and this?

lib/watch-sidebar.js
```diff
@@ -6,148 +6,142 @@ import WatchView from './watch-view';
...
-        }
-        );
-
-        this.watchViews = [];
-        this.addWatch();
-
-        this.hide();
-        atom.workspace.addRightPanel({item: this.element});
+  constructor(kernel) {
+    this.resizeStarted = this.resizeStarted.bind(this);
```
n-riesco: why is this necessary?

lib/watch-sidebar.js
```diff
@@ -6,148 +6,142 @@ import WatchView from './watch-view';
...
+    this.resizeStopped = this.resizeStopped.bind(this);
```
n-riesco: and this?

lib/watch-sidebar.js
```diff
@@ -6,148 +6,142 @@ import WatchView from './watch-view';
...
+    this.resizeSidebar = this.resizeSidebar.bind(this);
```
n-riesco: and this?

n-riesco:
I haven't been able to go through all the commit. Here are a few more comments:

-  Use `const` where appropriate.

- Please, don't use 2-space indentation. Besides the readability issue, It encourages sloppy coding (with high cyclomatic complexity).

- special care is needed with the translation of `?`, because it's incorrect at many places.

.eslintrc.yml
```diff
@@ -0,0 +1,20 @@
...
+
+globals:
+  atom: true
+
+settings:
+  import/core-modules: [ atom ]
+
+rules:
+  no-console: [0]
+  no-use-before-define: [0]
```
n-riesco: why?

.eslintrc.yml
```diff
@@ -0,0 +1,20 @@
...
+  no-param-reassign: [0]
+  no-underscore-dangle: [0]
```
n-riesco: are we using underscore dangle?

.eslintrc.yml
```diff
@@ -0,0 +1,20 @@
...
+  no-useless-escape: [0]
```
n-riesco: ???

.eslintrc.yml
```diff
@@ -0,0 +1,20 @@
+extends: airbnb-base
```
n-riesco: I couldn't find what rules `airbnb-base` is using.

lib/kernel-manager.js
```diff
@@ -193,19 +193,20 @@ export default class KernelManager {
 
   getKernelSpecFor(language, callback) {
     if (language == null) {
-      return null;
+      return;
```
n-riesco: Unless all the `if(xxx == null)` are fixed, this change will break users of `getKernelSpecFor`

lib/main.js
```diff
@@ -391,9 +391,9 @@ const Hydrogen = {
 
   _runAll(kernel) {
     const breakpoints = this.codeManager.getBreakpoints();
-    _.forEach(breakpoints, (start, i) => {
-      if (i === breakpoints.length) return;
-      const end = breakpoints[i + 1];
+    _.forEach(breakpoints, (end, i) => {
+      if (i === 0) return;
+      const start = breakpoints[i - 1];
```
n-riesco: I guess there is something else going on (not just a lint error)

spec/code-manager-spec.js
```diff
@@ -0,0 +1,64 @@
+'use babel';
+
+import CodeManager from '../lib/code-manager';
+
+describe('CodeManager', () => {
+  let CM = null;
```
n-riesco: `let CM;` should be enough

lib/kernel-manager.js
```diff
@@ -40,30 +40,29 @@ export default class KernelManager {
     const language = this.getLanguageFor(grammar);
     const kernel = this._runningKernels[language];
     delete this._runningKernels[language];
-    return __guard__(kernel, x => x.destroy());
+    __guard__(kernel, x => x.destroy());
   }
 
 
-  restartRunningKernelFor(grammar, onRestarted) {
+  restartRunningKernelFor(grammar, onRestarted = () => {}) {
```
n-riesco: this will create a `() => {}`

lib/plugin-api/hydrogen-kernel.js
```diff
@@ -1,36 +1,43 @@
...
+*/
 
-module.exports =
-class HydrogenKernel
-    constructor: (@_kernel) ->
-        @destroyed = false
+export default class HydrogenKernel {
+    constructor(_kernel) {
+        this._kernel = _kernel;
+        this.destroyed = false;
```
nikitakit: Note that `destroyed` is intentionally a part of the plugin API (i.e. plugins are expected and encouraged to use it). It can be switched to a property that proxies to `_kernel`, but removing it entirely is a breaking change. For a plugin API, stability is more important than perfection -- i.e. I would encourage not getting into the habit of making breaking changes.

lib/plugin-api/hydrogen-provider.js
nikitakit: This is so that a plugin always gets the same wrapper object for the same kernel (same reference, not just equal-by-value). This should be true no matter the source API call, e.g. `onDidChangeKernel` vs. `getActiveKernel`.

Plugins need to be able to store metadata per-kernel. I decided to go with the approach of letting them attach custom fields to the wrapper object, and expecting the fields to be there later on. (It's possible to make a breaking change to some uuid-based thing, but that's separate from ES6 migration. Especially since hydrogen has no native notion of kernel ids.)

lib/result-view.js
```diff
@@ -1,238 +1,273 @@
...
+        return this;
+    }
+
+    addCopyTooltip(element) {
+        return this.tooltips.add(atom.tooltips.add(element, {
+            title: () => {
+                if (this._executionCount === null) {
+                    return 'Copy to clipboard';
+                } else {
+                    return `Copy to clipboard (Out[${this._executionCount}])`;
```
nikitakit: Sorry, could you explain the race condition? I don't see any. Note that `title()` gets called during the mouseover event, not at construction-time.

lib/plugin-api/hydrogen-provider.js
n-riesco: > This is so that a plugin always gets the same wrapper object for the same kernel (same reference, not just equal-by-value).

Wouldn't you get the same wrapper if the wrapper is created by Kernel's constructor?

> Plugins need to be able to store metadata per-kernel.

That's a bad pattern, because API users don't have any guarantee to avoid name clashes.
Composition would be simple and safer pattern:
```js
let myH2Kernel = {
    kernel: kernelWrapper,
    myMetadata1: myMetadata1, 
    myMetadata2: myMetadata2, 
};

Comment (nikitakit) [inline lib/plugin-api/hydrogen-provider.js]: Name clashes are definitely a real issue, but the pattern you suggest is not an ideal solution either. If you listen to `onDidChangeKernel` events, you get wrapper objects and not any custom struct. So each plug in will need to end up rolling their own dictionary lookup to fix that. I think adding some methods to encourage better namespacing (e.g. `getData (namespace)`), or figuring out how to return a different set of wrapper objects to each consumer, are more promising approaches.

File: lib/result-view.js (hunk 29 above)
Comment (n-riesco) [inline lib/result-view.js]: My misunderstanding! I missed that `addResult` is only invoked with results for a given request (and thus a unique execution count).

File: lib/plugin-api/hydrogen-provider.js (hunk 12 above)
Comment (n-riesco) [inline lib/plugin-api/hydrogen-provider.js]: Oh! I see were you're coming from. Equality should work in this case; e.g.:
```
> var a = {}; (function(paramA) { return paramA === a; })(a);
true
```

**Comment by lgeiger** [comment]
@n-riesco Thanks for taking a look! I'm going through your comments now.

Most of them should be fixed already by the follow up commits after. Sorry for the bad commit history.

> Use const where appropriate.

This is fixed and is enforced by eslint.

> Please, don't use 2-space indentation. Besides the readability issue, it doesn't discourage sloppy coding (with high cyclomatic complexity).

I disagree. I think 4-spaces are really helpful for languages like coffeescript or python which rely heavily on indentation. But the standard for JS is 2-spaces, isn't it?

> special care is needed with the translation of ?, because it's incorrect at many places.

👍 👍

**Comment by rgbkrk** [comment]
> Please, don't use 2-space indentation. Besides the readability issue, it doesn't discourage sloppy coding (with high cyclomatic complexity).

The 2 space indentation comes from the airbnb eslint configuration. standard js also uses 2 space. Most developers I work with use 2-space indents.

I realize it's arbitrary - this is more about aligning with standards others use that we want to collaborate with.

File: lib/code-manager.js (hunk 5 above)
Comment (lgeiger) [inline lib/code-manager.js]: I'll cleanup the rest. 👍

File: lib/kernel-manager.js (hunk 8 above)
Comment (lgeiger) [inline lib/kernel-manager.js]: Hmm I actually have no idea.

**Comment by n-riesco** [comment]
On 15/11/16 19:01, Lukas Geiger wrote:
> @n-riesco <https://github.com/n-riesco> Thanks for taking a look! I'm going through your comments now.
> 
>     Please, don't use 2-space indentation. Besides the readability issue, it doesn't discourage sloppy coding (with high cyclomatic complexity).
> 
> I disagree. I think 4-spaces are really helpful for languages like coffeescript or python which rely heavily on indentation. But the standard for JS is 2-spaces, isn't it?

No, it isn't. Airbnb's standard is far from being the standard in Javascript.

What are the benefits of using 2-space indent?

**Comment by n-riesco** [comment]
On 15/11/16 18:58, Kyle Kelley wrote:
> *@rgbkrk* commented on this pull request.
> 
> ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
> 
> In .eslintrc.yml <https://github.com/nteract/hydrogen/pull/503>:
> 
>> @@ -0,0 +1,20 @@
> +extends: airbnb-base
> 
> https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb-base

Perhaps it's obvious, but I couldn't find the list of rules in that link.

**Comment by n-riesco** [comment]
On 15/11/16 19:01, Kyle Kelley wrote:
>     Please, don't use 2-space indentation. Besides the readability issue, it doesn't discourage sloppy coding (with high cyclomatic complexity).
> 
> The 2 space indentation comes from the airbnb eslint configuration. standard js also uses 2 space. Most developers I work with use 2-space indents.
> 
> I realize it's arbitrary - this is more about aligning with standards others use that we want to collaborate with.

I don't think it's arbitrary.
I've given a reason why 2-space indent is bad.
It doesn't penalise deep levels of indentation.

File: lib/main.js (hunk 11 above)
Comment (lgeiger) [inline lib/main.js]: They are just an artifact from decaffeinate.

File: .eslintrc.yml [hunk 30]
```diff
@@ -0,0 +1,20 @@
...
+  import/core-modules: [ atom ]
+
+rules:
+  no-console: [0]
+  no-use-before-define: [0]
+  no-param-reassign: [0]
+  no-underscore-dangle: [0]
+  no-useless-escape: [0]
+  prefer-rest-params: [0]
+  class-methods-use-this: [0]
```
lgeiger: I'm going to remove this again and use `static` methods as you mentioned above.

rgbkrk:
> Perhaps it's obvious, but I couldn't find the list of rules in that link.

From https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/.eslintrc it says it extends from `index.js` which loads each of the rules in `rules/`.

.eslintrc.yml
lgeiger: > I'm going to remove this again and use static methods as you mentioned above.

My bad. This won't work.

I'm using this rule mainly because it would arror in files like [kernel.js] since a lot of the methods don't use `this`.

.eslintrc.yml
lgeiger: Yep we have methods like `_execute` etc.

.eslintrc.yml
lgeiger: Otherwise it would complain about [this line](https://github.com/lgeiger/hydrogen/blob/43e8ae80732fab70a8c37669626d0bdb8009518e/lib/code-manager.js#L208)

rgbkrk:
> I've given a reason why 2-space indent is bad. It doesn't penalise deep levels of indentation.

Noted. I agree that 2-space indent does not penalize deep levels of indentation.

I'm still in favor of 2-space indent which is used by

* [airbnb](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/style.js#L64)
* [standard](http://standardjs.com/#the-rules)

However, I will refrain from going back and forth on this. It's up to you how you wish to carry this out with collaborators.

rgbkrk:
If I seem strongly opinionated on this, I'm not really. I'm pointing out the community effects. I can adapt to whatever convention the linter tells me to do.

lib/main.js
lgeiger: I remember fixing this the same way in the past but I can't remember what the exact problem was.

lib/kernel-manager.js
lgeiger: We could check for existence [here](https://github.com/lgeiger/hydrogen/blob/43e8ae80732fab70a8c37669626d0bdb8009518e/lib/kernel-manager.js#L51) If you prefere.

n-riesco:
```js
if (kernel instanceof WSKernel) {
    const future = kernel.restart();
    if (onRestarted) future.then(() => onRestarted(kernel));
}
```
and no need to define a closure below

```js
this.startKernel(kernelSpec, grammar, onRestarted);
```

lib/input-view.js
```diff
@@ -23,7 +23,7 @@ export default class InputView extends View {
   initialize(prompt, onConfirmed) {
     this.prompt = prompt;
     this.onConfirmed = onConfirmed;
-    atom.commands.add(this.element, { 'core:confirm': this.confirm });
+    atom.commands.add(this.element, { 'core:confirm': () => this.confirm() });
```
lgeiger: I'm going to refactor the input views soon. Since I noticed this quick fix half way through refactoring all of them.

rgbkrk: Does this get called without a bound `this` then?

You could also do:

```js
atom.commands.add(this.element, { 'core:confirm': this.confirm.bind(this) });
```

Depending on what `this` context you want.

lgeiger:
I could live with 4-space indents but I prefer 2-spaces since it makes it consistent within the nteract org.
Since the main reason we move to JS is to improve the experience for new contributors, I think using 4-spaces would be a step back on this goal.

That said, I tested every command in Hydrogen and fixed the bugs I initially introduced here. There are a few things left to do (probably some unnecessary `.bind(this)`, etc.). But I like to leave them for future PRs to keep the diff minimal.

lib/rename-view.js
lgeiger: I refactored it in #508

lib/watch-sidebar.js
lgeiger: They are actually necessary. I think because they are used [here](https://github.com/nteract/hydrogen/pull/503/files#diff-2809de1cb9bae13793c2862fe73ff693R118).

lgeiger:
Rebased

Issue #252, opened 2016-04-07
[DISCUSSION] Hydrogen CSS namespace


rgbkrk:
For the sake of readability in a user's config, we should start using a CSS (well, LESS) namespace for our outputs (and watch panel). Right now, as suggested in https://github.com/nteract/hydrogen/issues/160 to adjust the font-size you have to override `bubble-output-container`. Nested within other Atom stylesheets, it will begin to be unclear what `bubble` is when a user comes back to their stylesheet.

I'd love to see `hydrogen` in our class names and re-use existing styles where possible. We would still need to maintain the old classes for backwards compatibility, while moving ahead with new classes.

Thoughts?

rgbkrk:
[Related Twitter conversation](https://twitter.com/treycausey/status/718131309474746368)

n-riesco:
`hydrogen.less` could do with a clean up (not only font sizes, colours as well).

In the long term, I'd like to make possible the creation of plugins and themes.

n-riesco:
Perhaps, we could add a setting to configure this font size.

lgeiger:
I agree. A cleanup of `hydrogen.less` is much needed!

BenRussert:
I have been experimenting w/ hydrogen.less in @lgeiger's fork #772, I was thinking of starting a README on namespace/styling hydrogen elements. This way we could have a reference for development and a guide for users who want to customize on their own styles,less.

lgeiger:
@BenRussert That's a excellent idea!

I would also love to have support for different themes, though I haven't looked at how we can do this in Atom.

PR #516, opened 2016-11-24
Full width result view

### LINKED ISSUE #512: Limit results box width
Right now, it doesn't appear that the result bubble size is suitably limited in the horizontal dimension. The limit is determined by the longest line of the file (assuming soft linewrap is disabled), not by the width of the containing window.

If the lines printed to stdout are long, the box expands to the point where the controls (close/copy/open in new tab) are no longer visible. This problem is especially noticeable if using multiple panes.

I would propose to limit the result box width to be smaller than the current pane width. Some extra padding on the right would be useful to not interrupt scrolling. This is because results boxes can interfere with scrolling through the file: partway through scrolling the mouse runs over a result box and starts scrolling the stdout contents instead. Having margins around the results box helps prevent this. (It's technically possible to avoid this by either using the minimap to scroll, or scrolling by placing the mouse over the line number indicators. But at least for me my muscle memory doesn't have me do that.)

In addition to providing a margin specified in terms of pixels, another option is to use the "preferred line length" setting for the current file and cut of the results box at that many characters.

Screenshot of the problem:
screen shot 2016-11-20 at 8 46 36 pm

lgeiger:
Here is my initial proposal for a more robust multiline output area. The "single line" and status outputs are left as is.

I'm not a web designer and I'm very open to suggestions and contributions on this design. In any way this PR should really simplify the output system which makes it possible to add support for themes in the future.
bildschirmfoto 2016-11-24 um 00 51 13

~~The one thing that doesn't work yet is scrolling inside the output cells. I would really appreciate your help on this.~~

Closes #512

### DIFF
--- a/lib/result-view.js
+++ b/lib/result-view.js
@@ -131,30 +131,16 @@ export default class ResultView {
         htmlElement = webview;
       }
 
-      console.log('ResultView: Rendering as MIME ', mimeType);
-      console.log('ResultView: Rendering as ', htmlElement);
-      // this.getAllText must be called after appending the htmlElement
-      // in order to obtain innerText
-      this.outputContainer.appendChild(htmlElement);
-
-      if (mimeType === 'image/svg+xml') {
-        this.element.classList.add('svg');
-      }
-
-      if (mimeType === 'text/markdown') {
-        this.element.classList.add('markdown');
-      }
-
-      if (mimeType === 'text/latex') {
-        this.element.classList.add('latex');
-      }
-
       if ((result.stream === 'stderr' || result.stream === 'error') &&
       this.outputContainer.getElementsByTagName('span').length === 0) {
         this.outputContainer.classList.add('plain-error');
       } else {
         this.outputContainer.classList.remove('plain-error');
       }
+
+      console.log('ResultView: Rendering as MIME ', mimeType);
+      console.log('ResultView: Rendering as ', htmlElement);
+      this.outputContainer.appendChild(htmlElement);
     };
 
     const onError = error => console.error('ResultView: Rendering error:', error);
--- a/lib/result-view.js
+++ b/lib/result-view.js
@@ -11,67 +11,49 @@ export default class ResultView {
 
   constructor(marker) {
     this.marker = marker;
+
     this.element = document.createElement('div');
-    this.element.classList.add('hydrogen', 'output-bubble', 'empty');
+    this.element.classList.add('hydrogen', 'result-view', 'hidden');
 
-    const outputContainer = document.createElement('div');
-    outputContainer.classList.add('bubble-output-container');
-    outputContainer.onmousewheel = e => e.stopPropagation();
-    this.element.appendChild(outputContainer);
-
-    this.resultContainer = document.createElement('div');
-    this.resultContainer.classList.add('bubble-result-container');
-    outputContainer.appendChild(this.resultContainer);
-
-    this.errorContainer = document.createElement('div');
-    this.errorContainer.classList.add('bubble-error-container');
-    outputContainer.appendChild(this.errorContainer);
+    this.outputContainer = document.createElement('div');
+    this.outputContainer.classList.add('output-container', 'hidden');
 
     this.statusContainer = document.createElement('div');
-    this.statusContainer.classList.add('bubble-status-container');
-    this.spinner = this.buildSpinner();
-    this.statusContainer.appendChild(this.spinner);
-    outputContainer.appendChild(this.statusContainer);
+    this.statusContainer.classList.add('status-container');
 
-    const richCloseButton = document.createElement('div');
-    richCloseButton.classList.add('rich-close-button', 'icon', 'icon-x');
-    richCloseButton.onclick = () => this.destroy();
-    this.element.appendChild(richCloseButton);
+    this.spinner = this.buildSpinner();
 
     const actionPanel = document.createElement('div');
-    actionPanel.classList.add('bubble-action-panel');
-    this.element.appendChild(actionPanel);
-
-    const closeButton = document.createElement('div');
-    closeButton.classList.add('action-button',
-      'close-button', 'icon', 'icon-x');
-    closeButton.onclick = () => this.destroy();
-    actionPanel.appendChild(closeButton);
-
-
-    const padding = document.createElement('div');
-    padding.classList.add('padding');
-    actionPanel.appendChild(padding);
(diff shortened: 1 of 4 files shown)


lib/result-view.js
```diff
@@ -54,6 +53,17 @@ export default class ResultView {
...
 
+    this.outputContainer.onmousewheel = (event) => {
+      const clientHeight = this.outputContainer.clientHeight;
+      const scrollHeight = this.outputContainer.scrollHeight;
+      const scrollTop = this.outputContainer.scrollTop;
+      const atTop = (scrollTop !== 0 && event.deltaY < 0);
+      const atBottom = (scrollTop !== scrollHeight - clientHeight && event.deltaY > 0);
+
+      if (clientHeight < scrollHeight && (atTop || atBottom)) {
+        event.stopPropagation();
```
lgeiger: There's probably a better way of doing this, but this will work for now.

n-riesco:
> The one thing that doesn't work yet is scrolling inside the output cells.

Wasn't this handled in CSS before? `style='overflow:auto'`? E.g.:

```html
<div style='overflow:auto;width:200;height:200;border:solid'>
<div style='width:300;height:300;background-color:pink'>300x300</div>
</div>

<div style='overflow:auto;width:200;height:200;border:solid'>
<div style='width:100;height:100;background-color:pink'>100x100</div>
</div>
```

lgeiger:
Whoops I forgot to cross out my comment about the scrolling. I figured it out in the latest commit.

We used `outputContainer.onmousewheel = e => e.stopPropagation();` before to enable scrolling in the result bubble. But this blocks the scrolling of the text editor.
This behavior is inconvenient if the result bubbles are fullscreen so I needed to add some logic for when to scroll inside the bubble and when not.

@n-riesco Thanks for the help!

n-riesco:
@lgeiger :+1: I should've read the code!

hastebrot:
It would also be nice if the limit in `result-view.js` for the multiline result view could be configured in the settings view. `50` is currently hard-coded in `if (previousText === '' && text.length < 50`.

lgeiger:
@hastebrot Thanks for chiming in. You are right, the hard-coded 50 characters are not great.

I think the best way would be to check if the inline output fits in the current view instead of making it user configurable. Though I'm not sure how this could be implemented.

Would you like to tackle this one? I'm happy to merge a PR 😄

lgeiger:
At the moment I'm not using Hydrogen for my daily work. Could someone try out this PR and see how it feels?

n-riesco:
Some feedback after taking this PR for a spin:

- the output box taking almost all the available width looks ugly for small outputs (see screenshot)

- action panel overlaps with large outputs (see screenshot)

- svg output is centered unlike all the other outputs (see screenshot)

- (this may be hard to fix, without breaking other functionality; e.g. autocomplete) but I find very distracting that the margin goes over the output boxes.

- I manage to get a `Uncaught TypeError: Cannot read property 'screenRow' of undefined`when scrolling up

- It'd be nice to have a way to distinguish output, stdout and stderr

screenshot from 2016-11-29 18-53-07

screenshot from 2016-11-29 18-54-27

screenshot from 2016-11-29 18-56-55

lgeiger:
Thanks for taking a look!

> the output box taking almost all the available width looks ugly for small outputs

I did this on purpose to make the result bubbles feel more robust and visually consistent. I think the variable sizes can make the screen very cluttered. I wanted to make the multiline bubbles feel more like cells so we can support rich interactive output like plotly in the future.

> action panel overlaps with large outputs

Do you have a suggestion how we could fix it? The only option I see is to go back to a horizontal menu (which I tried to avoid).

> svg output is centered unlike all the other outputs 

I adopted the behavior of nteract here. We could align everything left if you like.

> I find very distracting that the margin goes over the output boxes.

Could you elaborate?

> I manage to get a Uncaught TypeError: Cannot read property 'screenRow' of undefinedwhen scrolling up

This is fixed in the latest Atom release: https://github.com/atom/atom/pull/13321

> It'd be nice to have a way to distinguish output, stdout and stderr

I think we would need to introduce multiple output containers to make this work. How does nteract and jupyter handle this?

rgbkrk:
> > svg output is centered unlike all the other outputs
> I adopted the behavior of nteract here.

I feel like our centering is a bad idea for images inside of nteract/nteract. We should use the default image layout that comes from the browser

n-riesco:
>  I wanted to make the multiline bubbles feel more like cells so we can support rich interactive output like plotly in the future.

I don't follow the reasoning there (i.e. I don't see what wouldn't allow the support of an interactive output). Having said this, please, take my comment just as one user's comment. If you don't see anything ugly, your opinion is more important.

> Do you have a suggestion how we could fix it?

padding?

> I adopted the behaviour of nteract here. We could align everything left if you like.

Again, take my comment as one user's comment.

> Could you elaborate?

margin


> How does nteract and jupyter handle this?

screenshot from 2016-11-29 19-58-32



---

One question: is `Hydrogen` planning to replicate `nteract`'s functionality?

lgeiger:
I should have expressed my self a bit more carefully. Please don't see this PR as "my vision for Hydrogen". I originally only wanted to refactor the result view code to address #252.
My refactor completely broke the styling so I just started from scratch 😆.

> I don't follow the reasoning there

The main thing I don't like about the current output bubbles are their behavior with large output. For example multiple plots in one output bubble don't look good.
This is why I initially opted for full width bubbles. I definitely see your point with the full width output bubbles now (I just tried it with a external widescreen monitor 😄 ).

I'll try to revise the design next week to address your feedback.

> I find very distracting that the margin goes over the output boxes.

Oh the wrap-guide, yes that's distracting. But I don't know if we can do anything about that.

> It'd be nice to have a way to distinguish output, stdout and stderr

I think we should leave this for a later PR.

> One question: is Hydrogen planning to replicate nteract's functionality?

Oh no, I just used it for inspiration here. Maybe a bit too much 😄

nikitakit:
I haven't had a chance to test the code yet, but I just want to quickly comment regarding the margin box. Have you tried limiting the width of the results box to the margin? This would help with scrolling, in that you could now use the area right of the margin to scroll the document instead of the results view.

It should also be the case that if you're actually following coding style and wrapping at the margin, you may want to open up multiple panes to make the most of your horizontal screen space. Any thoughts on this?

lgeiger:
Thanks for all of your comments! 👍 

I'm going to close this PR because I tried to do too much at once.

That said, I learned a lot during the process and will split the improvements into smaller PRs.

PR #520, opened 2016-11-29
Bump uuid and only import uuid/v4

lgeiger:
After https://github.com/n-riesco/jmp/pull/25 is shipped Hydrogen and jmp will use the same uuid package.


rgbkrk:
Hoorah for unity.

PR #521, opened 2016-11-29
Use commonmark instead of marked

lgeiger:
[`marked`](https://github.com/chjj/marked) isn't actively maintained anymore and has a few bugs.

This PR will make it consistent with nteract and use [`commonmark`](https://github.com/jgm/commonmark.js) as a markdown renderer. If anyone feels strongly against `commonmark` we can also consider using [`markdown-it`](https://github.com/markdown-it/markdown-it).

As an added bonus this reduces our dependencies.

### DIFF
--- a/lib/result-view.js
+++ b/lib/result-view.js
@@ -2,10 +2,9 @@
 
 import { CompositeDisposable } from 'atom';
 
-import * as transformime from 'transformime';
-import MarkdownTransform from 'transformime-marked';
+import { createTransform } from 'transformime';
 
-const transform = transformime.createTransform([MarkdownTransform]);
+const transform = createTransform();
 
 export default class ResultView {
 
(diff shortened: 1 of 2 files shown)


lib/result-view.js
```diff
@@ -2,10 +2,9 @@
 
 import { CompositeDisposable } from 'atom';
 
-import * as transformime from 'transformime';
-import MarkdownTransform from 'transformime-marked';
+import { createTransform } from 'transformime';
 
-const transform = transformime.createTransform([MarkdownTransform]);
+const transform = createTransform();
```
rgbkrk: Is there a default transform that uses commonmark in transformime then?

lgeiger: Yes: https://github.com/nteract/transformime/blob/master/src/commonmark.transform.js

rgbkrk:
Uh oh, I merged the other PR -- you'll need to rebase this one.

lgeiger:
No problem. Already rebased

PR #524, opened 2016-12-01
Update README.md with Kernel Gateway instructions for Docker

tiangolo:
# Update README.md with Kernel Gateway instructions for Docker

Add examples and tips of usage. 

This would help Docker users and should be simpler than using custom kernel connection files. 

Also, this would work without problems in Windows and Mac. (Actually, I wrote and tested it in a Windows machine with Docker Toolbox).

---

Note: I wrote the functionality to use custom kernel connection files (at least the first version) a while ago.

But now that you added support for kernel gateways, I think it should be the preferred way. I find it a lot cleaner and simpler to use. Additionally, with Kernel Gateways it works right away in non-Linux machines (Windows and Mac). :tada:

BTW, awesome work with all the improvements! With Kernel Gateway support, `zmq-prebuilt` and all! :tada:  Have some :taco: :taco: :taco: and :cake: !

### DIFF
--- a/README.md
+++ b/README.md
@@ -376,7 +376,7 @@ jupyter kernelgateway --ip=0.0.0.0 --port=8888
 - In the section "List of kernel gateways to use" add settings for the container your created
 - Use a `name` that you can remind when running Hydrogen
 - In the `baseUrl` section use the host or IP that you use to access your Docker containers:
-  - If you are using Docker Toolbox in Windows or Mac (or Docker for Windows, Docker for Mac), as it will be running in a virtual machine, the IP (host) would probably be like: `192.168.99.100`
+  - If you are using Docker Toolbox in Windows or Mac (or Docker for Windows, Docker for Mac), as it will be running in a virtual machine, the IP (host) would probably be like: `192.168.99.100`, you can read about it and check the `docker-machine ip default` command [in the official Docker docs](https://docs.docker.com/machine/get-started/#/run-containers-and-experiment-with-machine-commands)
   - If you are using Docker in a Linux machine and you are running Atom in that same machine you can just use `localhost` as the host of your `baseUrl`
 - For example, a possible configuration for Docker Toolbox in Windows or Mac could be:
 
--- a/README.md
+++ b/README.md
@@ -247,11 +247,223 @@ Unlike with local kernels, Hydrogen does not kill remote kernels when it disconn
 
 **Note:** Unlike a notebook server, the jupyter kernel gateway by default disables listing already-running kernels. This means that once disconnected from a kernel, you will not be able to reconnect to it. You can set `c.KernelGatewayApp.list_kernels = True` in your kernel gateway configuration to change this behavior.
 
+### Example remote kernel gateway
+
+To create a remote kernel gateway you could
+
+- Install Jupyter Kernel Gateway:
+
+```bash
+pip install jupyter_kernel_gateway
+```
+
+- Run the kernel gateway with parameters specifying the IP address and port to listen to, e.g.:
+
+```bash
+jupyter kernelgateway --ip=0.0.0.0 --port=8888
+```
+
+**Note**: "`0.0.0.0`" means "listen to all IPs, including public IPs".
+
+## Docker execution via kernel gateways
+
+You can use the same technique to create a kernel gateway in a Docker container. That would allow you to develop from Atom but with all the dependencies, autocompletion, environment, etc. of a Docker container.
+
+But, due to the way that the kernel gateway creates sub-processes for each kernel, you have to use it in a special way, you can't run the `jupyter kernelgateway` directly in your `Dockerfile` `CMD` section. You need to call it with an init manager such as [tini](https://github.com/krallin/tini) or run it from an interactive console.
+
+Here's an example of how to setup a Docker execution environment with Hydrogen running the kernel gateway from an interactive console:
+
+### Example Docker kernel gateway
+
+#### Dockerfile
+
+- Create a `Dockerfile`
+- Install `jupyter_kernel_gateway` in your `Dockerfile`
+- Expose the gateway port, in this example it will be `8888`:
+
+```Dockerfile
+FROM python:2.7
+
+# Remove in production
+RUN pip install jupyter_kernel_gateway
+EXPOSE 8888
+```
+
+#### Run Docker Container with Docker commands
+
+**Note**: alternatively, see below for `docker-compose` instructions.
+
+- Build your container:
+
+```bash
+docker build -t hydro .
+```
+
+- Run your container mapping the port of the gateway
+- Give your container a name
+- Make it run an infinite loop that just keeps the container alive:
+
+```bash
+docker run -d -p 8888:8888 --name hydro  hydro bash -c "while true; do sleep 10; done"
+```
+
+**Note**: you will only be able to run one container using that port mapping. So, if you had another container using that port, you will have to stop that one first. Or alternatively, you can create a mapping to a new port and add that configuration in the Hydrogen settings (see below).
+
+- Execute an interactive bash session in your running container:
+
+```bash
+docker exec -it hydro bash
+```
+
+- From that interactive session, start the gateway
+- Specify the IP `0.0.0.0` to make your container listen to public connections
+- Specify the port that you exposed in your `Dockerfile`:
+
+```bash
+jupyter kernelgateway --ip=0.0.0.0 --port=8888
+```
+
+#### Run Docker Container with Docker Compose
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -247,11 +247,223 @@ Unlike with local kernels, Hydrogen does not kill remote kernels when it disconn
...
+```bash
+docker run -d -p 8888:8888 --name hydro  hydro bash -c "while true; do sleep 10; done"
+```
+
+**Note**: you will only be able to run one container using that port mapping. So, if you had another container using that port, you will have to stop that one first. Or alternatively, you can create a mapping to a new port and add that configuration in the Hydrogen settings (see below).
+
+- Execute an interactive bash session in your running container:
+
+```bash
+docker exec -it hydro bash
```
MaximilianR: Why this approach rather than `tini`?

README.md
```diff
@@ -247,11 +247,223 @@ Unlike with local kernels, Hydrogen does not kill remote kernels when it disconn
...
+    build: .
+    ports:
+      - "8888:8888"
+    command: bash -c "while true; do sleep 10; done"
+```
+
+- The `docker-compose.yml` file has a port mapping using the port exposed in the `Dockerfile` and used in the `jupyter kernelgateway` command
+- The `command` overrides the default `CMD` in the `Dockerfile` (if there was one) and executes an infinite loop that would just keep the container alive
+
+**Note**: you will only be able to run one container using that port mapping. So, if you had another container using that port, you will have to stop that one first. Or alternatively, you can create a mapping to a new port and add that configuration in the Hydrogen settings (see below).
```
MaximilianR: FWIW this is repeated

README.md
```diff
@@ -247,11 +247,223 @@ Unlike with local kernels, Hydrogen does not kill remote kernels when it disconn
...
+
+#### Connect Atom
+
+- Go to the settings in Atom with: `ctrl-shift-p` and type `Settings View: Open`
+- Go to the "Packages" section
+- Type `Hydrogen` and go to package settings
+- In the section "List of kernel gateways to use" add settings for the container your created
+- Use a `name` that you can remind when running Hydrogen
+- In the `baseUrl` section use the host or IP that you use to access your Docker containers:
+  - If you are using Docker Toolbox in Windows or Mac (or Docker for Windows, Docker for Mac), as it will be running in a virtual machine, the IP (host) would probably be like: `192.168.99.100`
```
MaximilianR: How would someone find this?

Despite the port mappings in the d-c file, `localhost:8888` doesn't work?

README.md
```diff
@@ -247,11 +247,223 @@ Unlike with local kernels, Hydrogen does not kill remote kernels when it disconn
...
+```python
+import markdown [✓]
+markdown.version ['2.6.6']
+```
+
+#### Terminate the connection and container
+
+- To terminate a running kernel gateway you can "kill" it as any Linux process with `ctrl-c`
+
+But because of the way Jupyter Kernel Gateway creates sub-processes and due to the fact that you are running in a Docker container, the actual kernel process will still be running.
```
MaximilianR: Is this still true with `tini`?

MaximilianR:
Thanks a lot for doing this! Really helpful

README.md
```diff
@@ -247,11 +247,223 @@ Unlike with local kernels, Hydrogen does not kill remote kernels when it disconn
...
+#### Dockerfile
+
+- Create a `Dockerfile`
+- Install `jupyter_kernel_gateway` in your `Dockerfile`
+- Expose the gateway port, in this example it will be `8888`:
+
+```Dockerfile
+FROM python:2.7
+
+# Remove in production
```
MaximilianR: What to remove?

README.md
tiangolo: Because this would allow adding one single package and command during development that can then be removed for production. 

If for example the production image uses another system (e.g. Supervisor) it would become too cumbersome to add and configure `tini` and the Kernel Gateway during development but then remove it and use Supervisor again for production. That would happen, for example, in the case of a web application with `Flask`. I would gladly use interactively the models and everything in the Python environment during development and then remove the Kernel Gateway section for production.

If the only wants to have a Python environment in a remote Kernel just for, let's say, exploratory analysis, he could simply use one of the [Docker Jupyter Stacks](https://github.com/jupyter/docker-stacks) as his base image. But that would be a very specific case.

I wanted to document the very simplest approach that could be extended as needed by each user.

README.md
tiangolo: You refer to the "**Note**", right? Yep, I did it deliberately. 

I would expect users to skip one of the two sections (`docker` or `docker-compose`) and follow only the section of interest, but the warning applies for both. 

I wouldn't expect someone following the `docker-compose` approach to read carefully the note in the previous (skipped) section.

README.md
tiangolo: If they use anything different than Docker in Linux it will be running on a virtual machine (Docker for Windows, Docker for Mac or Docker Toolbox). So it would be a local virtual network IP. And if everything goes normally, it would be precisely that IP: `192.168.99.100`. But that could vary in each installation.

In any case, if someone got to the point in where he is using Docker in a non-Linux system, he most probably knows that. Also it is [officially documented here](https://docs.docker.com/machine/get-started/#/run-containers-and-experiment-with-machine-commands).

But if it's required, I could add a note stating that one possible way to get the IP of a locally running Docker in a non-Linux system could be: `docker-machine ip default`. Although this won't necessarily work "all" the time, for example if someone changed the name of the Docker Machine from `default` to something different. But most of the time (in a local environment) it would work.

And in another case, if the user is running Docker in a host different than the one running Atom, maybe a different machine in the same network or a (*not recommended*) public VPS on the cloud, the IP would be that of that other machine. And that would depend on each user.

README.md
tiangolo: That is just a comment that could be omitted, but it might remind someone that those 2 lines are only for development.

The block (of 2 lines) in the `Dockerfile` below the comment are the ones to be removed or commented out in production. For development it would look like:

```Dockerfile
# Remove in production
RUN pip install jupyter_kernel_gateway
EXPOSE 8888
```

And for production it could be:

```Dockerfile
# Remove in production
# RUN pip install jupyter_kernel_gateway
# EXPOSE 8888
```

Although, as commented above, the user could test his "production" environment without commenting out those lines in his `Dockerfile`, just by running the container normally, without overriding the default command with the loop and without having to re-build the image.

README.md
lgeiger: I think it would be enough to add the link to the official document 👍

tiangolo: Cool! I updated it in the newest commit.

tiangolo:
Thanks @MaximilianR for the review!

tiangolo:
Thanks @lgeiger for checking and approving! You guys are some of the coolest and most cheerful Open Source repo maintainers.

lgeiger:
@tiangolo We love merging your PRs keep up the great work!

MaximilianR:
It's merged and undoubtedly better than the last one. 

That said - I do think this would be far far simpler if running this was encompassed in the dockerfile, rather than 1) the dockerfile, 2) the run command, 3) the exec command and 4) the kernelgateway command in bash. 

But @tiangolo appreciate you doing this, and only wish it had been here earlier!

tiangolo:
I agree with you @MaximilianR in that I wish it could be simpler, for example, keeping a Docker container alive requires that loop that doesn't seem very simple at first.

But for the long term I think it can have it's advantages, for example:

* It should work for a very broad range of use cases, being quite "generalizable"
* Not having to re-build the image to change the local environment (that could mean even minutes for the building process to download all the dependencies, at every switch of environment... that would kind of "kill" the feature to me)
* By already having a dummy live container would allow doing things like, in the same container run the Gateway Kernel and at the same time run a Flask debugging server (via another exec command). Then develop using the Kernel and by every change saved the Flask debugging server would re-load (using a volume mounting the app code)
* If by accident someone sent the `Dockerfile` for production without commenting the development part, the image would still work as normally (somewhat bigger, by downloading the packages, but still would work)
* If someone needed to try live installing a new package (before adding it to the `Dockerfile` and re-compiling) he could just terminate the Kernel Gateway and install the package from the same container exec console and re-start the Kernel Gateway right away

---

I think it could be nice to have an "alternative" option too in the docs, for example showing how to use [Jupyter Docker Stacks](https://github.com/jupyter/docker-stacks) as in the [official Kernel Gateway docs](http://jupyter-kernel-gateway.readthedocs.io/en/latest/getting-started.html#using-a-docker-stacks-image). It probably might be just a section linking to those docs as an alternative to "Docker" and "Docker Compose". Or showing how to do it from scratch with `tini`. 

Maybe for the next time... (or you might consider creating a PR adding that :smile:).

PR #519, opened 2016-11-29
Remove unnecessary postinstall step

lgeiger:
`prebuild` 4.5.0 is here 🚀 

https://github.com/mafintosh/prebuild/pull/135 makes the postinstall step unnecessary which means `apm rebuild` will work now.
We should ship 1.1.0 after this.

Fixes #518 and may help with #507

PR #526, opened 2016-12-01
Bump jmp@0.7.4

lgeiger:
Necessary for #519


n-riesco:
Tested `apm rebuild`, but Atom failed to start because of `uuid/v4`. After running `apm install`, everything works.

lgeiger:
@n-riesco Thanks for testing! I think we should ship a new minor version soon.

PR #535, opened 2016-12-09
Limit result view width

### LINKED ISSUE #512: Limit results box width

lgeiger:
Fixes #512 

/cc @nikitakit

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -302,11 +302,14 @@ const Hydrogen = {
     const { element } = view;
 
     const lineHeight = this.editor.getLineHeightInPixels();
-    view.spinner.setAttribute('style',
-      `width: ${lineHeight + 2}px; height: ${lineHeight - 4}px;`);
+    view.spinner.setAttribute('style', `
+      width: ${lineHeight + 2}px;
+      height: ${lineHeight - 4}px;`);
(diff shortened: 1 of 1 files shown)


n-riesco:
WFM

screenshot from 2016-12-09 15-56-30

PR #536, opened 2016-12-10
Don't block editor scrolling in result views

lgeiger:
scroll

### DIFF
--- a/lib/result-view.js
+++ b/lib/result-view.js
@@ -15,7 +15,17 @@ export default class ResultView {
 
     const outputContainer = document.createElement('div');
     outputContainer.classList.add('bubble-output-container');
-    outputContainer.onmousewheel = e => e.stopPropagation();
+    outputContainer.onmousewheel = (event) => {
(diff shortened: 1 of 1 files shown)


lib/result-view.js
```diff
@@ -15,7 +15,17 @@ export default class ResultView {
...
+      const clientHeight = outputContainer.clientHeight;
+      const scrollHeight = outputContainer.scrollHeight;
+      const scrollTop = outputContainer.scrollTop;
+      const atTop = (scrollTop !== 0 && event.deltaY < 0);
+      const atBottom = (scrollTop !== scrollHeight - clientHeight && event.deltaY > 0);
+
+      if (clientHeight < scrollHeight && (atTop || atBottom)) {
+        event.stopPropagation();
+      }
+    };
```
rgbkrk: Gosh this looks familiar

lgeiger: Really? I tried to look for something like this in nteract but could find it 😄

PR #537, opened 2016-12-10
Fix: execution_count message interfers with spinner

lgeiger:
Bring back the smooth spinner


lib/result-view.js
```diff
@@ -113,6 +113,8 @@ export default class ResultView {
       return;
     }
 
+    console.log('ResultView: Add result', result);
+
```
rgbkrk: Do you mean to leave this console log?

lgeiger: I only moved it a few lines down in order to remove the logs generated from execution_count and status events.

nikitakit:
Thanks for fixing this! I started running into this bug after making the execution count changes, but couldn't figure out exactly what was going on.

PR #542, opened 2016-12-11
Upgrade to @jupyterlab/services 0.34.2

nikitakit:
This updates to the latest APIs, and changes all mentions of `jupyter-js-services` to `@jupyterlab/services`.

This allows token-based authentication for notebooks, starting with notebook version 4.3. However there's something funny going on with gateways (see https://github.com/jupyter/kernel_gateway/issues/204), so we might have to hold off on merging this until the cause of the gateway issue is determined.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ To connect to a gateway server, you must first add the connection information to
 }]
 ```
 
-Each entry in the gateways list needs at minimum a `name` (for displaying in the UI), and a value for `options.baseUrl`. The `options` are passed directly to the `jupyter-js-services` npm package, which includes documentation for additional fields.
+Each entry in the gateways list needs at minimum a `name` (for displaying in the UI), and a value for `options.baseUrl`. The `options` are passed directly to the [`@jupyterlab/services`](https://github.com/jupyterlab/services) npm package, which includes documentation for additional fields.
 
 After gateways have been configured, you can use the **"Hydrogen: Connect to Remote Kernel"** command. You will be prompted to select a gateway, and then given the choice to either create a new session or connect to an existing one.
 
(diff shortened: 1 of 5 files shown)


nikitakit:
(Pushed new version that removes debug code I accidentally left in. Sorry about that)

nikitakit:
Based on the comments at jupyter/kernel_gateway#204 (as well as personal testing), the issue is with the combination of kernel gateway v1.1.1 and. notebook v4.3.0. This combination is broken regardless of which version of `@jupyterlab/services` is used to access it. Therefore this gateway issue shouldn't block merging this PR.

Of course, any regression testing for known-working server versions would be appreciated.

lgeiger:
@nikitakit Thanks for investigating.
I have the mentioned versions installed and also experience this (expected) problem. I'm going to merge it since the problem isn't caused by this PR!

lgeiger:
I noticed some differences in the error handling between connecting to `jupyter notebook` or `jupyter kernelgateway`:

The connection error get's handled correctly in the case of connecting to a notebook. Using the kernelgateway the connection fails silently (only get's logged in the dev console).

We might have discussed this already in the past and it was due to technical limitations but I can't remember it for sure and can't find the relevant issue.

nikitakit:
Could you provide some more details on the difference you're encountering? Connections failing silently is a bug that should be fixed.

The only difference between notebooks and gateways is the API call `listRunning`, which fails for gateways with the default config. The error gets logged to the dev console, while the code proceeds to create a new session. But for all other API calls (such as creating/connecting a session or listing kernelspecs), we should still strive to surface errors to the user. We should have the full capabilities to handle errors such as an invalid config, server-not-running, or request unauthorized in a reasonable manner.

lgeiger:
Here's a GIF using `jupyter kernelgateway`:
gateway

And the console log:
bildschirmfoto 2016-12-12 um 23 17 04

The error message in the terminal is the same as you posted in https://github.com/jupyter/kernel_gateway/issues/204

PR #550, opened 2016-12-16
Better scoping of keybindings

### LINKED ISSUE #548: keybinding shift-enter overrides other keybindings
I'm finding that `hydrogen:run-and-move-down` is overriding the key binding `shift+enter` anywhere in atom.

This includes when the find/replace box is active (here shift+enter should move the cursor to the previous match) and when in the stage files modal of the git-plus packages (here the keybinding should confirm staging of selected files).

Is there any way we can get better scoping for the hydrogen command so it isn't just `.atom-text-editor`? Maybe we can restrict it to only be active when the `source` scope is active?

lgeiger:
Only respect keybindings in `atom-text-editor:not([mini])`.

Fixes #548

### DIFF
--- a/keymaps/hydrogen.cson
+++ b/keymaps/hydrogen.cson
@@ -9,7 +9,7 @@
 # https://atom.io/docs/latest/behind-atom-keymaps-in-depth
 
 # ---------- Editor --------------
-'atom-text-editor':
+'atom-text-editor:not([mini])':
     'shift-enter': 'hydrogen:run-and-move-down'
     'shift-alt-enter': 'hydrogen:run-cell-and-move-down'
     'alt-i': 'hydrogen:toggle-inspector'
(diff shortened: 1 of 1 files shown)


keymaps/hydrogen.cson
```diff
@@ -9,17 +9,14 @@
...
     'shift-enter': 'hydrogen:run-and-move-down'
     'shift-alt-enter': 'hydrogen:run-cell-and-move-down'
     'alt-i': 'hydrogen:toggle-inspector'
 
 # Override shift-enter and cmd-enter
 '.platform-darwin atom-text-editor:not([mini])':
     'shift-enter': 'hydrogen:run-and-move-down'
-    'cmd-enter': 'hydrogen:run'
-
-'.platform-darwin atom-text-editor':
```
rgbkrk: Took me a sec to realize CMD enter was still available here

lgeiger:
Lets 🚢 1.4.0

PR #552, opened 2016-12-16
Improve documentation for remote kernels

nikitakit:
The first commit adds a few words describing the `token` option.

It the second commit, I make the recommendation that users install Jupyter Notebook (and not the kernel gateway), and document how to set it up locally and on the public internet.

I feel obliged to mention why I would personally recommend the notebook over the gateway server. When I first wrote the remote kernel feature, the kernel gateway server was a newly-added toplevel Jupyter project. It seemed very appealing to have the kernel management capabilities of the notebook without the web UI -- something that is still true today. Without doing much testing I assumed that the two would be roughly equivalent, and wrote all the docs to mention gateways.

As it turns out, however, the gateway server actually depends on the notebook and imports large chunks of its code (so it offers no savings in terms of install size). Moreover, the two servers have diverging sets of features. Notebook 4.3 now has a token implementation that enables security and interacts well with Hydrogen. Kernel Gateway + Notebook 4.2 also has tokens, but its implementation differs slightly in a way that breaks hydrogen (it can't use `?token=<>` URL parameters for websocket connections, meaning that you can't actually send code to be run). This could conceivably be fixed in a Kernel Gateway + Notebook 4.3 combination, but that one remains broken (I reported this bug a few days ago). In fact, I suspect that if you `pip install` the kernel gateway today on a clean system, you will get the broken combination that will fail 100% of the time.

In conclusion, these problems will probably take a few dev cycles to be resolved upstream. In the meantime, I would like to advise Hydrogen users to use a configuration that is known to be working and that encourages good security practices. (Ideally we would update the Docker docs as well, but I haven't touched those because I can't test anything docker-related)

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,12 +185,13 @@ To connect to a gateway server, you must first add the connection information to
 [{
     "name": "Remote notebook",
     "options": {
-            "baseUrl": "http://example.com:8888"
+            "baseUrl": "http://example.com:8888",
+            "token": "my_secret_token"
     }
 }]
 ```
 
-Each entry in the gateways list needs at minimum a `name` (for displaying in the UI), and a value for `options.baseUrl`. The `options` are passed directly to the [`@jupyterlab/services`](https://github.com/jupyterlab/services) npm package, which includes documentation for additional fields.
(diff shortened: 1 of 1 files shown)


rgbkrk:
Thanks @nikitakit for taking the time to improve the documentation _and_ to outline the reasoning and history so well.

lgeiger:
That's awesome, thanks 🎉

PR #559, opened 2016-12-22
fix(keymap): Don't override ctrl-backspace

### LINKED ISSUE #555: Don't override ctrl-backspace
Please don't override the `ctrl-backspace` shortcut.

It is a standard shortcut for "delete a word before the cursor". Overriding it confuses those who use it.

lgeiger:
Fixes #555

### DIFF
--- a/keymaps/hydrogen.cson
+++ b/keymaps/hydrogen.cson
@@ -21,10 +21,10 @@
     'cmd-enter': 'hydrogen:run'
     'cmd-ctrl-enter': 'hydrogen:run-all'
 
-# Override ctrl-enter and ctrl-backspace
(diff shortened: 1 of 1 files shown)


lgeiger:
Should we ship a patch release?

rgbkrk:
This does change user land behavior, would it really be major or was it never able to work right before with this prior key binding?

lgeiger:
I'm not sure if it ever worked correctly. To be save and if we want to adhere strictly to semver it would be a major release. But I don't think we're quite ready for `v2.0.0`.

rgbkrk:
I think doing a patch release is A-OK.

PR #566, opened 2016-12-28
Support multiple kernels per file (markdown, &c.)

fasiha:
This PR adds one module to Hydrogen, in `robust-grammar.js`, which seeks to generalize [`TextEditor.getGrammar()`](https://atom.io/docs/api/v1.12.7/TextEditor#instance-getGrammar). If the current editor is AsciiDoc or Github-Flavored Markdown, we check the cursor’s [scope](https://atom.io/docs/api/v1.12.7/TextEditor#instance-scopeDescriptorForBufferPosition) for an alternative grammar. This happens when editing fenced code blocks in GFM, e.g.

With this robust alternative to `getGrammar()`, Hydrogen just has to pick the correct kernel given the cursor position.

Hydrogen continues to work with stand-alone (non-Markdown) source files, as illustrated below (where `Hydrogen: Run` is invoked on the fenced code block around the cursor by a keyboard macro):

gfm-support

The extra call to the robust grammar checker each time `createResultBubble()` is run *should* be negligible. If this PR is accepted, the doors will be open for multilingual literate programming documentation!

(Caveat—as in the screencap above, this depends on `language-gfm` to understand a particular fenced code block, and it doesn’t (yet) support Octave. This issue is tracked at atom/language-gfm/issues/170.)

### DIFF
--- a/lib/autocomplete-provider.js
+++ b/lib/autocomplete-provider.js
@@ -2,7 +2,7 @@
 
 import _ from 'lodash';
 
-import RobustGrammar from './robust-grammar';
+import getRobustGrammar from './robust-grammar';
 
 const iconHTML = `<img src='${__dirname}/../static/logo.svg' style='width: 100%;'>`;
 
@@ -31,7 +31,7 @@ export default function (kernelManager) {
 
     // Required: Return a promise, an array of suggestions, or null.
     getSuggestions({ editor, bufferPosition, prefix }) {
-      const grammar = RobustGrammar(editor);
+      const grammar = getRobustGrammar(editor);
       const language = kernelManager.getLanguageFor(grammar);
       const kernel = kernelManager.getRunningKernelFor(language);
       if (!kernel) {
--- a/lib/autocomplete-provider.js
+++ b/lib/autocomplete-provider.js
@@ -2,6 +2,8 @@
 
 import _ from 'lodash';
 
+import RobustGrammar from './robust-grammar';
+
 const iconHTML = `<img src='${__dirname}/../static/logo.svg' style='width: 100%;'>`;
 
 const regexes = {
@@ -29,7 +31,7 @@ export default function (kernelManager) {
 
     // Required: Return a promise, an array of suggestions, or null.
     getSuggestions({ editor, bufferPosition, prefix }) {
-      const grammar = editor.getGrammar();
+      const grammar = RobustGrammar(editor);
       const language = kernelManager.getLanguageFor(grammar);
       const kernel = kernelManager.getRunningKernelFor(language);
       if (!kernel) {
--- a/lib/inspector.js
+++ b/lib/inspector.js
@@ -3,6 +3,8 @@
 import { MessagePanelView, PlainMessageView } from 'atom-message-panel';
 import * as transformime from 'transformime';
 
+import RobustGrammar from './robust-grammar';
+
 const transform = transformime.createTransform();
 
 export default class Inspector {
@@ -13,8 +15,8 @@ export default class Inspector {
   }
 
   toggle() {
-    const editor = atom.workspace.getActiveTextEditor();
-    const grammar = editor.getGrammar();
+    const editor = atom.workspace.getActiveTextEditor()
+    const grammar = RobustGrammar(editor);
     const language = this.kernelManager.getLanguageFor(grammar);
     const kernel = this.kernelManager.getRunningKernelFor(language);
     if (!kernel) {
--- a/lib/inspector.js
+++ b/lib/inspector.js
@@ -15,7 +15,7 @@ export default class Inspector {
   }
 
   toggle() {
-    const editor = atom.workspace.getActiveTextEditor()
+    const editor = atom.workspace.getActiveTextEditor();
     const grammar = RobustGrammar(editor);
     const language = this.kernelManager.getLanguageFor(grammar);
     const kernel = this.kernelManager.getRunningKernelFor(language);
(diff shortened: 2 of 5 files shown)


rgbkrk:
You did it!

lib/autocomplete-provider.js
```diff
@@ -2,6 +2,8 @@
 
 import _ from 'lodash';
 
+import RobustGrammar from './robust-grammar';
```
rgbkrk: Since this isn't a class and is invoked as a function, use a leading lowercase like `robustGrammar`. In fact, I would probably call it `getRobustGrammar` or `getCurrentGrammar`.

rgbkrk:
I'm certainly in favor of this, my one nit (which will require a search-replace refactor) is mostly for code clarity.

fasiha:
Thank you for your kind words! I changed it to `getRobustGrammar` 🙇

lib/autocomplete-provider.js
fasiha: Done!

rgbkrk:
How does kernel cleanup happen with multiple kernels?

fasiha:
Ah. Right.

It appears that restarting and shutting down a kernel will do this to *all* kernels open in the document. That’s not right. I’ll investigate and see how to make only the kernel tied to the grammar local to the cursor (i.e., in a fenced code block) is affected by these commands.

lgeiger:
Woo that's great 🎉 

I'm looking forward to test and review this PR tomorrow.

fasiha:
@lgeiger thanks, I humbly request your help in understanding why, after two kernels are started (in the example above, JS and Python), and I go to restart just one of them, it appears that both have restarted. That is, after just restarting one kernel, `kernel-manager.js`’s `_runningKernels` member still shows the other kernel as active, but it appears that none of the variables I created in *either* kernel survived the restart. Scratching head…

lgeiger:
I found two (related problems) wich may be the cause for the shutdown weirdness.

1. The status bar doesn't update together with the current grammar, which leads to potential exceptions:
err1

2. Hydrogen always starts a new kernel for even if a kernel with the same grammar is already running:
err2

fasiha:
Sorry for the delay! These holidays are brutal on my computer time…

@lgeiger thank you for pointing this out, those do look like they’re related! But, super-sorry, I’m having trouble understanding what sequence of operations trigger that—could you write how to achieve those two problems? Thank you thank you 🙇🙏!

lgeiger:
Unfortunately I don't really know how to fix this either.

Part of the problem is that [`onKernelCanged`](https://github.com/nteract/hydrogen/blob/master/lib/main.js#L142)  is only triggered [`onEditorChanged`](https://github.com/nteract/hydrogen/blob/master/lib/main.js#L126). But we would need it to be also triggered on a grammar change caused by the cursor moving to an other grammar block. I don't know if Atom has a event for that.

rgbkrk:
> we would need it to be also triggered on a grammar change caused by the cursor moving to an other grammar block. 

What we're wanting and it doesn't seem like we have access to easily is when the scope changes so we can run `atom.grammars.grammarForScopeName`. The only way I'm seeing that now is if we watch all the buffer changes (get the scope, do the rest of the calculations in here).

jonwoodring:
I'm really looking forward to this merge. This is exactly what I was looking for out of Hydrogen (markdown and fences for literate programming).

Have you thought about a way to .md => code? That is, automatically change markdown to source code comments and remove fences?

Thanks so much for this.

rgbkrk:
Checkout #637 for this now.

fasiha:
Hi @jonwoodring hmm, it’s a bit off-topic for here, hopefully we don’t get chased off 😅. There are so many ways to do Markdown⟹code, but I’m ashamed to say I’ve mostly used this Hydrogen feature as a notebook (in the Jupyter/IPython sense), where code doesn’t ever leave the documentation.

In a [previous incarnation](https://github.com/nteract/hydrogen/pull/428) of this feature, I had another Atom plugin that would ask Hydrogen to evaluate the current fenced code block, then insert the result as comments at the bottom of the code block. That was sufficient to complete the Markdown documentation, while also leaving open the possibility of compiling the Markdown to HTML where fenced code blocks could become *live*, so for transpile-to-JS languages a visitor could change the code and see the results live ([Klipse](http://blog.klipse.tech/klipse/2016/09/09/klipse-languages.html) lets us do this very nicely—JS, ClojureScript, even Python and [C++!](http://blog.klipse.tech/cpp/2016/12/29/blog-cpp.html)).

On one [project](https://github.com/fasiha/mudderjs#code-to-detect-the-prefix-property), at the bottom of code blocks I had comments like `// < export TARGET.js` which then a simple external (non-Atom) script could locate, and “tangle” (in Knuth literate-programming terminology) into source code.

I only did this as a stop-gap, because without sourcemaps it’s hard to translate an error to original source. So in the future I will look into making sourcemaps (for JS), and failing that, see what it’d be like to copy the entire Markdown file to every code output file and just “invert” code–no-code sections, so each line of Markdown would get commented out and the only thing left uncommented in each file would be the code belong to that file. I’m not sure if that’d work but…

Sorry, no answers here yet 🙇!

fasiha:
@jonwoodring the recently-released Github-Flavored Markdown spec (see [example 111](https://github.github.com/gfm/#example-111) and preceding discussion) allows the info string of fenced code blocks to have arbitrary data, so we can have code blocks like, e.g.,
~~~markdown
```js file=src/index.js
// cool code here
```
~~~
Here’s a very early mover in this space: https://github.com/killercup/waltz

rgbkrk:
Thanks for noting that! I'm so happy they made a specification on top of commonmark.

Issue #528, opened 2016-12-02
Docker proposal


MaximilianR:
Follow up to https://github.com/nteract/hydrogen/pull/524

I think the simplest way is a dockerfile:

```dockerfile
FROM python:2
ARG TINI_VERSION=v0.10.0
ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
RUN chmod +x /tini
RUN pip install jupyter_kernel_gateway
ENV KG_LIST_KERNELS=True
```

A docker-compose:

```yaml
version: '2'
services:
  jupyter:
    context: .
    entrypoint: ["/tini", "--"]
    command: ["jupyter", "kernelgateway", "--ip=0.0.0.0", "--port=8888"]
```

And a kernel settings default:

```
[{"name": "docker", "options": {"baseUrl": "http://localhost:8888"}}]
```

I'd propose this replaces basically all the README from L250-580. But probably I'm missing something and there's useful stuff there?

Questions:

- Do we need a docker-compose file, or should we just the commands in the dockerfile?
- Should we map the current path by default? This would allow you to run the code you're working on
- What else are we missing?

lgeiger:
Thanks for opening the issue. Unfortunately I don't know much about Docker.

@tiangolo May be able to answer your questions.

MaximilianR:
Thanks for the response @tiangolo !

> ...But in that case, wouldn't it be easier to use a Jupyter Docker stack?

Yes, good idea. 

> Or the equivalent docker-compose.yml, probably both options.

Agreed

> The kernel settings wouldn't necessarily be like you specified above, as many users won't run that from a Linux machine but rather from Windows or Mac, and in those cases it will mostly be a Virtual Machine with a local IP (as 192.168.99.100).

I use Docker for Mac and `localhost` worked for me. What makes you think you need the IP?

> So it should probably be kept for backward compatibility.

Your call, but there is a lot there! Might be simpler as a link to a 'Further detail' / 'Previous implementation' page. For a first time user, there's a lot of detail that's not really needed, and makes it look arduous to set this up when it's surprisingly simple. I'm sure there are several good approaches for making it simpler, though. 

> I think I can easily add that to the README.md (it's almost already written in this issue 😆 ). Or do you want to take a stab at it and create a PR?

If you want to add it I would be honored! If you prefer me to, I'm happy to take a stab in the near future.

tiangolo:
Excellent @MaximilianR !

[...]

PR #569, opened 2016-12-30
Kernel Connection docs in README improvements

tiangolo:
As discussed in https://github.com/nteract/hydrogen/issues/528:

* Put the section for custom Kernel Connection files in an external file (not in the README).
* Add a section in the README with instructions to set up a Docker kernel gateway based on a Jupyter Docker Stack.
* Adjust other sections to account for the new section.

I tested all the instructions in a local "demo" project to make sure they worked as expected.

I guess @MaximilianR and @lgeiger can check it. :smiley: 

I'm sorry for the delay creating this PR, I got very busy with end-of-year project deliveries and I wanted to have the time to test all the instructions and make sure everything worked.

### DIFF
--- a/KERNEL_CONNECTION.md
+++ b/KERNEL_CONNECTION.md
@@ -1,0 +1,128 @@
+# Custom kernel connection example (inside Docker)
+
+**Note**: Hydrogen now supports using kernel gateways ([see the instructions in the main page](https://github.com/nteract/hydrogen/blob/master/README.md)). Using that option might be simpler and will allow you to use the functionality in Windows or Mac very easily.
+
+You can use a custom kernel connection file to connect to a previously created kernel.
+
+For example, you can run a kernel inside a Docker container and make Hydrogen connect to it automatically.
+
+Hydrogen will look for a kernel JSON connection file under `./hydrogen/connection.json` inside your project. If that file exists, Hydrogen will try to connect to the kernel specified by that connection file.
+
+Here's a simple recipe for doing and testing that with Python:
+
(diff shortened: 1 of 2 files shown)


KERNEL_CONNECTION.md
```diff
@@ -0,0 +1,128 @@
+# Custom kernel connection example (inside Docker)
+
+**Note**: Hydrogen now supports using kernel gateways ([see the instructions in the main page](https://github.com/nteract/hydrogen/blob/master/README.md)). Using that option might be simpler and will allow you to use the functionality in Windows or Mac very easily.
```
lgeiger: This sounds to me a bit like kernel gateways aren't supported on Linux which isn't the case.

tiangolo: Good point. Which is like totally the contrary...

Maybe like this:

---

**Note**: Hydrogen now supports using kernel gateways ([see the instructions in the main page](https://github.com/nteract/hydrogen/blob/master/README.md)). Using that option might be simpler in most of the use cases. For example, it will allow you to use the functionality from a Docker virtual machine in Windows or Mac as easily as if it was Linux.

---

What do you think @lgeiger ?

lgeiger: @tiangolo Sounds great! 👍

tiangolo: Done @lgeiger ! :tada:

lgeiger: Thanks 👍 ❤️

PR #573, opened 2017-01-04
fix excessive trim of output

### LINKED ISSUE #564: preview have a problem
Thank you for this package,It's very nice.
But I found a petty problem
code:
```
for x in range(1, 11):
    print('{0:2d} {1:3d} {2:4d}'.format(x, x*x, x*x*x))
```
 the first line is doesn't work under the preview:
```
1   1    1
 2   4    8
 3   9   27
 4  16   64
 5  25  125
 6  36  216
 7  49  343
 8  64  512
 9  81  729
10 100 1000
```

slavaGanzin:
fix #564 
Works on Python 3 and ijavascript
Should be tested on other kernels

### DIFF
--- a/lib/kernel.js
+++ b/lib/kernel.js
@@ -131,7 +131,7 @@ export default class Kernel {
         type: 'text',
         stream: 'pyout',
       };
-      result.data['text/plain'] = result.data['text/plain'].trim();
+      result.data['text/plain'] = result.data['text/plain'].replace(/^(\s+\n)+/, '')
     } else {
       result = {
         data: {},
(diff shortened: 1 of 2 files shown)


lib/kernel.js
```diff
@@ -131,7 +131,7 @@ export default class Kernel {
         type: 'text',
         stream: 'pyout',
       };
-      result.data['text/plain'] = result.data['text/plain'].trim();
+      result.data['text/plain'] = result.data['text/plain'].replace(/^(\s+\n)+/, '');
```
lgeiger: This Regex will only trim the beginning of the string and not the end.

rgbkrk:
@lgeiger why are plaintext outputs trimmed to begin with?

lgeiger:
@rgbkrk I don't know the full reason but if it isn't trimmed our single line output wouldn't work.

Maybe we should even solve it inside `transformime`.

rgbkrk:
Oh I see, that's the trick to make it similar to our LightTable like single-line-values whereas Jupyter frontends are generally assumed to be an overall output.

slavaGanzin:
As I remember: It was trimmed because julia kernel prepend answer with some newlines

PR #574, opened 2017-01-09
hydrogen projects

slavaGanzin:
This is work in progress PR and it's open for discussion.

### Purposes:
1. Make hydrogen friendlier for development purposes
2. Forget about static code analyzers (as a bonus)
3. Be compatible with https://github.com/nteract/hydrogen/pull/566

### Changes:
1. ```hydrogen.cson``` file can be added in root of any of your projects. It configure multiple kernels initialization, for example you can launch some code before returning kernel to user.
2. startupCode setting was removed in prior of using new init file
3. Kernel root is always project root (and not first launched file dir)

### Hydrogen.cson
```cson
defaultKernel: 'Python 3'
#Status: Not done
# Specifies what kernel to launch as a default in this project
# Value: display_name

init:
# Key - display_name of kernel
# value - params
  "Javascript (Node.js)":
     code: '"use strict"' #code that should be launched before running files
     files: [ #list of files that would be executed by kernel before returning to user
         './any.file.js' #Path, for now, should be relative to project root
     ]
  "Python 3":
     code: '''
%matplotlib inline
import nltk
import numpy
'''
    
```

### DIFF
--- a/hydrogen.cson
+++ b/hydrogen.cson
@@ -1,0 +1,14 @@
+javascript:
+  code: 'a=1'
+  
+python:
+  kernel: 'Python 3'
+  code: '''
+  qwe=123
+  '''
+  files: [
+    './py.py'
+    './py2.py'
+  ]
+    
+  
--- a/lib/config.js
+++ b/lib/config.js
@@ -43,7 +43,7 @@ const Config = {
       type: 'string',
       default: '{}',
     }
-  },
+  }
 };
 
 export default Config;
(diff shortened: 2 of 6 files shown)


n-riesco:
> 2. `startupCode` setting was removed in prior of using new init file

I'd like to keep `startupCode`, because it'd be very inconvenient to have to create a `hydrogen.cson` for every python project just to add `%matplotlib inline`.

In general, I think `hydrogen.cson` should be an opt-in feature and Hydrogen should be able to work if not present.

slavaGanzin:
I'll revert startupCode

PR #581, opened 2017-01-11
Fix autocomplete when using MagicPython

### LINKED ISSUE #465: Spurious autocomplete when using MagicPython
I recently installed the MagicPython package for atom, to get myself better syntax highlighting for python3-only features. However, this causes hydrogen to start autocompleting `%%!` after you type a `:`, which is extremely annoying.

Steps to reproduce:
- Install the MagicPython package, and disable the language-python package
- Set hydrogen's language mappings setting to include `{"python":"magicpython"}`
- Open a python file and make sure the grammar is set to MagicPython
- Start a hydrogen python kernel
- Type `while True:`, then press enter

Expected behavior: the cursor goes to a new line
Actual behavior: autocomplete menu suggesting `%%!` pops up right before you press enter, so you end up accidentally accepting the suggestion instead of moving to a new line.

I find that having to press ESC to close autocomplete any time I want to start an indented block is really irritating.

Does anyone have an idea about why this is happening, and if there's a way to address this?

lgeiger:
This will add support for non default grammars.

Fixes #465

### DIFF
--- a/lib/autocomplete-provider.js
+++ b/lib/autocomplete-provider.js
@@ -2,6 +2,7 @@
 
 import _ from 'lodash';
 import log from './log';
+import Config from './config';
 
 const iconHTML = `<img src='${__dirname}/../static/logo.svg' style='width: 100%;'>`;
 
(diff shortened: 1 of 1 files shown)


Joulious:
Unfortunately this bug still persists. I use Hydrogen with MagicPython. Autocomplete only works when I enable language-python. Could you please take a look into this again? Thank you.

rgbkrk:
Can you document the full behavior and show the current version of Atom, MagicPython, and Hydrogen that you're on?

PR #593, opened 2017-01-23
Disable kernel interrupt on windows

### LINKED ISSUE #592: Break/interrupt kernel and continue using envirnoment; Bug?
Hi, 
as far as I understand it is not possible to currently interrupt/break some running code and continue using the up until the then computed variables. Atleast if I use the 'interrupt-kernel' command I am unable to run new code before restarting the kernel (this could be a local issue, but I have also seen the same issue among friends). Meaning I have to run all the code again (which is kind of tedious when testing). Is this meant to work or is it not currently implemented?

Note: I have tried searching for similar issues without success. It could however be that I have used the wrong wording. If that is the case, my apologies.

Best,
Tobias

lgeiger:
Unfortunately windows doesn't support signals. 

[Node.js docs](https://nodejs.org/api/process.html#process_signal_events):

> Note: Windows does not support sending signals, but Node.js offers some emulation with process.kill(), and ChildProcess.kill(). Sending signal 0 can be used to test for the existence of a process. Sending SIGINT, SIGTERM, and SIGKILL cause the unconditional termination of the target process.

Closes #592

### DIFF
--- a/lib/zmq-kernel.js
+++ b/lib/zmq-kernel.js
@@ -109,12 +109,16 @@ export default class ZMQKernel extends Kernel {
 
 
   interrupt() {
-    if (this.kernelProcess) {
+    if (process.platform === 'win32') {
+      atom.notifications.addWarning('Cannot interrupt this kernel', {
+        detail: 'Kernel interruption is currently not supported in Windows.',
+      });
+    } else if (this.kernelProcess) {
       log('ZMQKernel: sending SIGINT');
       this.kernelProcess.kill('SIGINT');
     } else {
       log('ZMQKernel: cannot interrupt an existing kernel');
-      atom.notifications.addWarning('Cannot interrupt this kernel');
+      atom.notifications.addWarning('Cannot interrupt an existing kernel');
     }
   }
 


nikitakit:
This incompatibility with Windows is certainly a good catch!

Do you think it would be worthwhile to implement a workaround using the Windows event APIs, rather than simply declaring interruption to be unsupported?

IIRC the notebook server on Windows supports interrupting kernels -- and it looks like the code responsible can be found [here](https://github.com/jupyter/jupyter_client/blob/7ade9f3db2a08372fc1eb6e1bf3a74f20b83e42e/jupyter_client/manager.py#L394), [here](https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/win_interrupt.py), and [here](https://github.com/jupyter/jupyter_client/blob/439981a03fb0c986b104fc039a64f504902210e6/jupyter_client/launcher.py#L83).

Note that I don't use Windows myself, so I won't be able to evaluate/test this.

lgeiger:
Thanks for the links! Implementing a workaround would definitely be a better solution and we could use it in nteract as well.

That said, I'm not using Windows either and have no knowledge about Windows events so I won't implement this any time soon.

I guess disabling the feature is the best we can do so far.

nikitakit:
@lgeiger That's fair. It then makes sense to keep an issue filed in case someone with Windows wants to contribute a workaround.

lgeiger:
> It then makes sense to keep an issue filed in case someone with Windows wants to contribute a workaround.

Totally agree 👍 
Would you like to open one?

PR #607, opened 2017-02-06
#600: remove img style to prevent img downscaling

### LINKED ISSUE #250: Figure Size
Thanks for the awesome package! 

Would it be possible to make the figure size adjustable? I've tried to adjust it using matplotlib, but didn't seem work.

### LINKED ISSUE #600: How is the displayed figure size set in Atom/Hydrogen?
I'm not sure this is the right place to ask this question, but here goes.

I can't figure out how the displayed size of a pyplot figure in Hydrogen is determined. It seems like there is a maximum pixel area or something. Here are two example screenshots. Note that the widths of the two figures should be the same, but if I increase the height, the overall width is reduced (compare figure widths to the 80 character text width line). It's harder to tell, but the second figure is slightly taller than the first. What's going on? The overarching goal here is to get Atom/Hydrogen to display larger figures.

**figsize=(12,8)**
screen shot 2017-01-27 at 17 42 46

**figsize=(12,10)**
screen shot 2017-01-27 at 17 43 09

Earthson:
EDIT:
Fixes #250 
Fixes #600

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "Hydrogen",
   "main": "./lib/main",
-  "version": "1.5.4",
+  "version": "1.5.5",
   "description": "Run code and get results inline using Jupyter kernels like IPython, IJulia, and iTorch",
   "author": "Will Whitney <me@willwhitney.com>",
   "keywords": [


styles/hydrogen.less
```diff
@@ -398,9 +394,6 @@
             }
 
             video, img, svg {
```
lgeiger: Since it is empty now we can remove the whole `video, img, svg` selector.

lgeiger:
I really like the new behavior.
@n-riesco @nikitakit @rgbkrk What do you think?
scale

n-riesco:
I also prefer the new behaviour, but not in the watch view.

lgeiger:
> I also prefer the new behaviour, but not in the watch view.

Good point! The watch view should still scale to fit.

Earthson:
@lgeiger Is it possible to make bubble max-height configurable?

lgeiger:
Sorry closed it by accident. -> #609

PR #609, opened 2017-02-07
Remove img style to prevent img downscaling

### LINKED ISSUE #250: Figure Size

### LINKED ISSUE #600: How is the displayed figure size set in Atom/Hydrogen?

lgeiger:
Follow up of #607 

I also included a fix to address https://github.com/nteract/hydrogen/pull/607#issuecomment-277645086
Fixes #250 
Fixes #600

### DIFF
--- a/styles/hydrogen.less
+++ b/styles/hydrogen.less
@@ -137,10 +137,6 @@
 
         video, img, svg {
             display: block;
-            max-height: 500px !important;
-            max-width: 100% !important;
-            width: unset !important;
-            height: unset !important;
         }
 
         td, th {
@@ -398,9 +394,6 @@
             }
 
             video, img, svg {
-                height: 100%;
-                width: 100%;
-                max-height: 350px !important;
             }
         }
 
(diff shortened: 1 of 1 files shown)


lgeiger:
@Earthson Thank you for tackling this!

PR #621, opened 2017-02-27
Fix cell magic

### LINKED ISSUE #612: Multiple cell magics do not work with Run All
**Description:**
When I have multiple cell magics in a document, only the very first one works with Run All.

**Steps to Reproduce:**

1. `Run All` on this document: image

2. Observe syntax error (cell magic is not being picked up.)

**Versions:**

```
ProductName:	Mac OS X
ProductVersion:	10.10.5
Atom    : 1.13.1
Hydrogen: 1.5.5
```

**Logs:**

Please post any error logs and the output of the developer tools as described in our [Debugging Guide](https://github.com/nteract/hydrogen/blob/master/TROUBLESHOOTING.md).

lgeiger:
Fixes #612

### DIFF
--- a/lib/code-manager.js
+++ b/lib/code-manager.js
@@ -160,6 +160,8 @@ export default class CodeManager {
     if (cursor.row > 0) {
       buffer.backwardsScanInRange(regex, [start, cursor], ({ range }) => {
         start = range.start;
+        start.row += 1;
+        start.column = 0;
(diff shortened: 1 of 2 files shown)


lgeiger:
For cell magics it's important that they're on the first line of the cell. This wasn't the case before.

PR #622, opened 2017-02-27
Always handle stderr of kernel process

### LINKED ISSUE #619: Handle stderr of kernel process
I think we should raise an error notification if the kernel process writes to `stderr`. At the moment we only log them to the dev console.

This will help debugging issues like https://github.com/nteract/hydrogen/issues/618#issuecomment-280914369

lgeiger:
Fixes #619

### DIFF
--- a/lib/zmq-kernel.js
+++ b/lib/zmq-kernel.js
@@ -47,17 +47,10 @@ export default class ZMQKernel extends Kernel {
       });
 
       this.kernelProcess.stderr.on('data', (data) => {
-        data = data.toString();
-
-        log('ZMQKernel: stderr:', data);
-
-        const regexp = getKernelNotificationsRegExp();
-        if (regexp && regexp.test(data)) {
-          atom.notifications.addError(this.kernelSpec.display_name, {
-            description: data,
-            dismissable: true,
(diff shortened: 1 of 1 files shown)


lib/zmq-kernel.js
```diff
@@ -47,17 +47,10 @@ export default class ZMQKernel extends Kernel {
       });
 
       this.kernelProcess.stderr.on('data', (data) => {
-        data = data.toString();
-
-        log('ZMQKernel: stderr:', data);
-
-        const regexp = getKernelNotificationsRegExp();
-        if (regexp && regexp.test(data)) {
```
lgeiger: We still use this `RegExp` for `stdout`. I think we should make it easier for users and only provide a boolean option which defaults to `false`.

@n-riesco If I recall correctly you added the `RegExp`. What do you think?

n-riesco: LGTM

IPython and IJulia write to `stdout` on startup (so problem there).
If there are any kernels that write to `stderr` on startup, we can revisit this decision.
We'll also need to revisit this code, once Jupyter introduces the use of a kernel nanny.

PR #625, opened 2017-02-28
Replace kernel notification regexp with boolean

lgeiger:
This simplifies it for users to config notifications from the kernel process (`stdout`). Though this will make it less flexible.

Fixes https://github.com/nteract/hydrogen/pull/622#discussion_r103333484

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -21,9 +21,9 @@ const Config = {
     },
     kernelNotifications: {
       title: 'Enable Kernel Notifications',
-      description: 'By default, kernel notifications are only displayed in the developer console. This setting defines a RegExp to filter what kernel notifications will also be shown as Atom notification bubbles. Example: `error|warning`',
-      type: 'string',
-      default: '(?!)',
+      description: 'Notify if kernels writes to stdout. By default, kernel notifications are only displayed in the developer console.',
+      type: 'boolean',
+      default: false,
     },
     gateways: {
       title: 'List of kernel gateways to use',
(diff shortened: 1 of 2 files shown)


lib/config.js
```diff
@@ -21,9 +21,9 @@ const Config = {
     },
     kernelNotifications: {
       title: 'Enable Kernel Notifications',
-      description: 'By default, kernel notifications are only displayed in the developer console. This setting defines a RegExp to filter what kernel notifications will also be shown as Atom notification bubbles. Example: `error|warning`',
-      type: 'string',
-      default: '(?!)',
+      description: 'Notify if kernels writes to stdout. By default, kernel notifications are only displayed in the developer console.',
+      type: 'boolean',
+      default: false,
```
rgbkrk: Do we have to do any migration work since the type has changed here?

rgbkrk:
Wow I just accidentally clicked approve after "requesting changes". Oops.

lib/zmq-kernel.js
```diff
@@ -22,27 +22,16 @@ export default class ZMQKernel extends Kernel {
...
-      };
-
       this.kernelProcess.stdout.on('data', (data) => {
         data = data.toString();
 
-        log('ZMQKernel: stdout:', data);
-
-        const regexp = getKernelNotificationsRegExp();
-        if (regexp && regexp.test(data)) {
+        if (atom.config.get('Hydrogen.kernelNotifications')) {
```
lgeiger: > Do we have to do any migration work since the type has changed here?

We don't need to do any migration work since it's only used here.

PR #626, opened 2017-03-01
log: Add config option

n-riesco:
* Added debug option, so that it's possible to enable debug messages on
  a running session.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -17,12 +17,14 @@ const Config = {
     autocomplete: {
       title: 'Enable Autocomplete',
       includeTitle: false,
+      description: 'If enabled, use autocomplete options provided by the current kernel.',
       type: 'boolean',
       default: true,
     },
     debug: {
-      title: 'Enable debug messages',
+      title: 'Enable Debug Messages',
       includeTitle: false,
+      description: 'If enabled, log debug messages onto the dev console.',
       type: 'boolean',
       default: false,
     },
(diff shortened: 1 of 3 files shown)


lgeiger:
Excellent idea! That's a lot easier 🎉 

Would you like to update the [debugging guide](https://github.com/nteract/hydrogen/blob/master/TROUBLESHOOTING.md#debugging-guide) as well?

n-riesco:
@lgeiger Shall I remove all the info about the opening atom in dev mode?

lgeiger:
> @lgeiger Shall I remove all the info about the opening atom in dev mode?

Yeah, I would replace it with the new config option.

lib/config.js
```diff
@@ -16,35 +16,49 @@ const Config = {
   schema: {
     autocomplete: {
       title: 'Enable Autocomplete',
+      includeTitle: false,
```
lgeiger: What does `includeTitle: false` do? I couldn't find anything in the Atom docs

n-riesco: I had to [read the code](https://github.com/atom/settings-view/blob/197f962fec680e1b9b1c7c8f5fd162a99731d959/lib/settings-panel.coffee#L48-L69) to fix this issue.

n-riesco: I'll try to get a PR merged in Atom after I'm done publishing an update for IJavascript.

lgeiger: Interesting, on my machine I don't see a difference between `includeTitle: false,` and the default setting.

If it fixes a issue on your machine, let's go with it 👍

PR #627, opened 2017-03-01
Evaluate React for Hydrogen

### LINKED ISSUE #40: Multiple independent kernels
It's useful to have separate kernels in different files/projects if they are not related. It seems it's only possible to use one shared kernel of the same type currently.

Maybe give each running kernel a unique identifier (hash, PID, ...), this makes them distinguishable and would also give the user the opportunity to switch between them (if necessary).

I'd like to use this feature to run/debug several micro services for example, once #25 is implemented.

lgeiger:
In order to gain more experience with React and to evaluate its usage inside Atom I took a stab at implementing the status bar as a React component. If we start using React we could use`nteract`'s components in the future. With this in mind I'd like to share this experiment with you (it got a bit out of control but it actually works 😄):

I setup a global store managed by [MobX](https://mobx.js.org/) which stores all running kernels and a reference to the current kernel depending on the active editor.
I opted for MobX over Redux because it is simpler to introduce in our current codebase (especially for a me who doesn't know very much about React and Redux). Though this doesn't mean we can't switch to Redux in the future, MobX brings us already in the right direction.

The idea is to make it easier to manage the running kernels and automatically update the relevant views (for now only the status bar) which would make it easier to fix #40 in the future.

@rgbkrk @n-riesco I would love to hear your feedback.  Do you think we should move in that direction?

The biggest downside is the package size increase of 5.9 MB. Though we could experiment with [Preact](https://preactjs.com/) for reducing the package size.

If we decide to move in that direction, here would be my future steps:
- [x] Test this PR to make sure everything works
- [ ] get rid of `kernelManager` or simplify it
- [ ] Reduce size of `main.js` since the functions can now use values from `store`
- [x] Refactor Inspector as React component and get rid of `atom-message-panel` (2MB)
- [ ] Refactor List Views to use [`atom-select-list`](https://github.com/atom/atom-select-list)  instead of the deprecated `atom-space-pen-views` which relies on jQuery
- [ ] Refactor Watch Sidebar as React component
- [ ] Replace `transformime` with `@nteract/transforms`


rgbkrk:
Ooooh, happy to see you're trying out MobX for this. 😄

rgbkrk:
Are you able to do some summary statistics to find out which packages are taking the most space?

lgeiger:
> Are you able to do some summary statistics to find out which packages are taking the most space?

MobX is quite lightweight. React and ReactDOM take 5.2MB

rgbkrk:
I'm generally in favor of us being able to share components, with React currently being the leading contender for cross project collaboration (even beyond nteract).

lib/kernel.js
```diff
@@ -1,20 +1,27 @@
...
+import { grammarToLanguage } from './utils';
+import store from './store';
 
-import StatusView from './status-view';
 import WatchSidebar from './watch-sidebar';
 import HydrogenKernel from './plugin-api/hydrogen-kernel';
 import log from './log';
 
 export default class Kernel {
+  @observable executionState = 'loading'
```
lgeiger: This is actually everything that's necessary for MobX to do it's magic. 👀 
I really enjoy it for our use case. c5b27c41a4b3f689231a850aba62e3ec1358111b Simplified it a lot.

rgbkrk:
Just based on the commits this is looking pretty cool!

lgeiger:
I should have better called this PR "Lukas learning React and MobX" 😄

React is a bit tricky when using it only on parts of the application but I'm beginning to really enjoy MobX. It makes everything much more transparent and can even be used without any frontend library.

lgeiger:
Does anyone use Facebook's Nuclide or any other packages using React and is able to test if this Branch plays well with it?

rgbkrk:
I use nuclide and could test this out.

lgeiger:
Thanks!

lgeiger:
I tried out this branch with Nuclide and it worked great 🎉 
I was a bit afraid about using different versions of React at first but it doesn't seam to be a problem.

rgbkrk:
It's the weekend now so I can still take this for a spin later on.

rgbkrk:
Are you wanting to merge as much as you've done here for the moment?

lgeiger:
> It's the weekend now so I can still take this for a spin later on.

No problem! I wasn't aware that Nuclide is that easy to install ;)

> Are you wanting to merge as much as you've done here for the moment?

I think this PR is in a good state now. There's still a lot to simplify here but we can keep that for later to keep things moving forward 👍 

One comment on the inspector:
This PR removes the option to minify the Inspector. It's been broken for most languages anyway and  in my option doesn't add a lot of value.

rgbkrk:
> I think this PR is in a good state now.

Are you hoping to merge this as is or do you want to complete more on your checklist?

lgeiger:
> Are you hoping to merge this as is or do you want to complete more on your checklist?

I think it's better to complete the checklist in future PRs to keep the diff somewhat manageable.

rgbkrk:
💥

lgeiger:
Atom's Linter v2 now uses React as well: https://github.com/steelbrain/linter-ui-default
Even more resources for inspiration.

PR #637, opened 2017-03-15
Support multiple kernels per markdown file

lgeiger:
This PR adapts #566 to our current setup of handling state with MobX and fixes https://github.com/nteract/hydrogen/pull/566#issuecomment-269627957.

@fasiha Would you like to try out this branch?

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -340,6 +340,10 @@ const Hydrogen = {
 
 
   runAll() {
+    if (isMultilanguageGrammar(store.editor.getGrammar())) {
+      atom.notifications.addError('"Run All" is not supported for this file type!');
+      return;
+    }
     if (store.kernel) {
       this._runAll(store.kernel);
       return;
@@ -362,6 +366,11 @@ const Hydrogen = {
 
 
   runAllAbove() {
+    if (isMultilanguageGrammar(store.editor.getGrammar())) {
+      atom.notifications.addError('"Run All Above" is not supported for this file type!');
+      return;
+    }
+
     const cursor = store.editor.getLastCursor();
     const row = codeManager.escapeBlankRows(0, cursor.getBufferRow());
     const code = codeManager.getRows(0, row);
@@ -373,6 +382,10 @@ const Hydrogen = {
 
 
   runCell(moveDown = false) {
+    if (isMultilanguageGrammar(store.editor.getGrammar())) {
+      atom.notifications.addError('"Run Cell" is not supported for this file type!');
+      return;
+    }
     const { start, end } = codeManager.getCurrentCell();
     const code = codeManager.getTextInRange(start, end);
     const endRow = codeManager.escapeBlankRows(start.row, end.row);
--- a/lib/main.js
+++ b/lib/main.js
@@ -24,6 +24,7 @@ import AutocompleteProvider from './autocomplete-provider';
 import HydrogenProvider from './plugin-api/hydrogen-provider';
 import reactFactory from './utils/react';
 import log from './utils/log';
+import { isMultilanguageGrammar } from './utils';
 
 const Hydrogen = {
   config: Config.schema,
@@ -78,6 +79,23 @@ const Hydrogen = {
     this.subscriptions.add(atom.workspace.observeActivePaneItem((item) => {
       if (item && item === atom.workspace.getActiveTextEditor()) {
         store.updateEditor(item);
+        store.setGrammar(item);
+      }
+    }));
+
+    this.subscriptions.add(atom.workspace.observeTextEditors((editor) => {
+      if (isMultilanguageGrammar(editor.getGrammar())) {
+        const cursorSubscription = editor.onDidChangeCursorPosition(
+          _.debounce(() => { store.setGrammar(editor); }, 75),
+        );
+
+        const editorSubscription = editor.onDidDestroy(() => {
+          cursorSubscription.dispose();
+          editorSubscription.dispose();
+        });
+
+        this.subscriptions.add(cursorSubscription);
+        this.subscriptions.add(editorSubscription);
       }
     }));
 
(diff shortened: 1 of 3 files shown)


lgeiger:
One thing I noticed is that with this workflow the commands "Run All" and "Run Cell" won't work as expected.

lgeiger:
multilang

lib/main.js
```diff
@@ -78,6 +79,23 @@ const Hydrogen = {
...
       if (item && item === atom.workspace.getActiveTextEditor()) {
         store.updateEditor(item);
+        store.setGrammar(item);
+      }
+    }));
+
+    this.subscriptions.add(atom.workspace.observeTextEditors((editor) => {
+      if (isMultilanguageGrammar(editor.getGrammar())) {
+        const cursorSubscription = editor.onDidChangeCursorPosition(
+          _.debounce(() => { store.setGrammar(editor); }, 75),
```
lgeiger: We don't need  to switch grammar instantaneously. Therefor we debounce the event.

lgeiger:
I tested this branch and didn't found any regressions 🎉 
The only downside is that  "Run All" and "Run Cell" aren't supported in this workflow, since we would need to do a lot of language specific parsing to be able to do the right thing.

fasiha:
This is *so* **awesome**—thank you!!! I’m not sure what “Run Cell” means but could “Run All” be faked by scrolling to the top of the doc, jumping to each fenced code block (triple-ticks or triple-tildes), evaluating “Run” till the last fenced code block, before returning to the original cursor location…?

rgbkrk:
Wow, Hydrogen just keeps getting better and more stable. _drool_

lgeiger:
@fasiha Would you be willing to add a few lines about the usage of this feature to our README? That would be super cool! 🎉 

> Could “Run All” be faked 

You're right we could fake it but that would mean introducing (probably a lot of) language specific code which we try to avoid in Hydrogen since it makes maintaining quite hard.
Unless you are interested in implementing this I think for now the best way would be to disable "Run Cell" and "Run All"  for markdown files.
What do you think?

fasiha:
> @fasiha Would you be willing to add a few lines about the usage of this feature to our README? 

@lgeiger that’d be the least I could do, thanks for trusting me with it. I’m using the feature right now to write a doc for Real Work (¡!), let me see how compelling a use-case I can make.

Could of questions, because I’m dense—

(1) Is there anything Markdown-specific about this? From quickly reading the source, it appears that it should work with AsciiDoc or reST or in fact any format that Atom understands to have “code blocks”—am I right? (Not asking this to be ungrateful 😂! Just wondering how inclusive it can be. I personally don’t use anything except Markdown, though that’s just out of laziness than any proper investigation of alternatives.)

(2) If so, is that what you mean by “language-specific code”? I.e., Hydrogen would have to know, for each format (Markdown, asciiDoc, etc.), where the fenced code block around the cursor started and ended, in order to send that code snippet to Jupyter? I can see that as being really bothersome, yes, so I’m fine with having to manually-select the fenced code block (or have a custom macro, as I have).

nikitakit:
With this functionality, doesn't it make sense to have "Run Cell" run the current fenced block?

I don't use multi-language Markdown documents myself, but my understanding is that a natural way of splitting up code into sections is to create multiple code blocks (possibly with Markdown text in between)

lgeiger:
> That’d be the least I could do, thanks for trusting me with it. I’m using the feature right now to write a doc for Real Work (¡!), let me see how compelling a use-case I can make.

Excellent, thanks @fasiha! If you want to write a whole blog post about your use case we're always happy to add it to our [Medium publication](https://medium.com/nteract). 😄 

> (1) Is there anything Markdown-specific about this? 

You are right, it works for every markup language that supports code blocks with syntax highlighting in Atom (I only tested it with markdown so far). To support another language one only needs to add it to [this array](https://github.com/nteract/hydrogen/pull/637/files#diff-eae83c5838640e1c85538d8c9fbe35e3R7).

> (2) If so, is that what you mean by “language-specific code”?

Correct. We would need to write a different parser for every supported language.

> With this functionality, doesn't it make sense to have "Run Cell" run the current fenced block?

@nikitakit That's absolutely true, I just don't see a great way of implementing this right now and personally won't take a closer look at this problem. 😉

mcburton:
@lgeiger I know this is closed and merged, but I just want to say being able to "Run Cell" on Markdown fenced blocks is a _crucial_ feature. If Hydrogen could support this it would make possible Atom a plagform for writing generalized rmarkdown-ish files (whatever kernel jupyter supports). As @fasiha mentioned, this would also allow me to do "real work" (scholarly writing with citations, charts/graphs/, code blocks, SPELL CHECK). In short, it would be amazing (and a nice complement to the Notebook, which I also use heavily).

fasiha:
@mcburton I had a little macro for atom-macros (which no longer works,
argh) that selected fenced code blocks, invoked Hydrogen’s `Run` on the
selection, and returned the cursor to its initial location.

I plan to spin this off into its own plugin called select-GFM-fence, with a
customizable setting of a command to run after selection (which Hydrogen
users could set to `Hydrogen: Run`).

Would this address your need too?

mcburton:
@fasiha I'd be happy to try it out!

fasiha:
Note that while these Hydrogen changes help with interactive programming in
Markdown, there is not currently any way to “export” your Hydrogen/Jupyter
results (plots, tables, etc., which Hydrogen renders so nicely). So with
future `Run Cell` or `Run All Cells` support, you’d just see nice results
in your Atom editor, without any way of automatically exporting those
outputs to PNG files, HTML/GFM tables, etc. You’d have to manually save
figures and etc…

I guess that’d be the next step for Markdown+Hydrogen!

mcburton:
Yeah, this isn't a problem for me at the moment since I just want this for
previewing outputs in the writing process. The way Hydrogen is currently
set is almost perfect.

I could see a separate "publishing/export" step, akin to Knitr, that runs
the whole file and generates HTML/PDF/whatever and then connects into
Pandoc for further transformation.

Totally agree, great next step for markdown+hydrogen!

rgbkrk:
Building up a "log" of code run + outputs wouldn't be too bad. Call it "record", have it create a new cell (in-memory) on everything executed, save it to disk. That would be sweet.

lgeiger:
@fasiha @mcburton Great suggestions! I'm really happy to hear you working on a plugin for this!

In the past we had some discussions about introducing a [plugin API for such language specific features](https://github.com/nteract/hydrogen/blob/plugin-api/api.md#hydrogen-language-plugin-api). It would be awesome to expose more of Hydrogen's functionality to be able to interact with it in a more powerful way.

lgeiger:
I think we should ship a new version soon so people are able to try it out!

In the current state it's already pretty great for writing a README with powerful examples that are ensured to work 😄

PR #642, opened 2017-03-17
Add code of conduct

lgeiger:
It's never to late to add one 😄 

@rgbkrk Added your email to the contact section as well.


rgbkrk:
Thank you!

PR #650, opened 2017-03-21
Add initial Flow typings

lgeiger:
Almost forgot about this branch 😆 

This adds some initial flow typings in order to utlimately prevent us from introducing exception.


rgbkrk:
Excellent!

/cc @smashwilson -- looks like Lukas is making flow types for Atom -- have you seen anything of the sort?

lgeiger:
I used the flow types maintained by Facebook's Nuclide: https://github.com/facebook/nuclide/tree/master/flow-libs

`prettier-atom` also has a small set of flow types: https://github.com/jlongster/prettier-atom/blob/master/decls/index.js

tiangolo:
Although late...

Just in case and just so you know, I haven't used Flow, but I've been using TypeScript (I'm using it right now in an Electron project). And looking at the code they seem quite similar.

I don't really know the utility of my comment, but just FYI... :)

lgeiger:
What I like about flow is that we can introduce it gradually, which makes it very easy to add to an existing code base.

tiangolo:
Cool! :+1: 

The "best one" is the one you already know :smile:

lgeiger:
Feel free to add more flow types, PR's are very welcome 😄

Issue #560, opened 2016-12-26
Altair visualization library doesn't work


idmit:
Hello!

I've been working with your package lately and it's really great. Unfortunately, it fails to display the output of the [Altair](https://altair-viz.github.io/) library. I try running this code from their homepage:

```python
from altair import Chart, load_dataset

# load built-in dataset as a pandas DataFrame
cars = load_dataset('cars')

Chart(cars).mark_circle().encode(
    x='Horsepower',
    y='Miles_per_Gallon',
    color='Origin',
)
```

## Notebook

![screen shot 2016-12-27 at 01 35 51](https://cloud.githubusercontent.com/assets/2546728/21487211/dc56da64-cbd4-11e6-8a38-c4a5013dcba8.png)

## Hydrogen

![screen shot 2016-12-27 at 01 34 21](https://cloud.githubusercontent.com/assets/2546728/21487188/a17190ba-cbd4-11e6-9b20-b26223ef3846.png)

## Notebook output contents

```html
<div class="output_subarea output_html rendered_html"><div class="vega-embed" id="c21b9d22-57b7-4cc8-9edf-bac5abf1c864">
  <div class="vega" style="position: relative;">
    <canvas width="624" height="405" class="marks">
    </canvas>
  </div>
  <div class="vega-actions">
    <a href="#" target="_blank" download="vega.png">Export as PNG</a>
    <a href="#">View Source</a>
    <a href="#">Open in Vega Editor</a>
  </div>
</div>

<style>
.vega-embed svg, .vega-embed canvas {
  border: 1px dotted gray;
}

.vega-embed .vega-actions a {
  margin-right: 6px;
}
</style>
</div>
```

## Hydrogen output contents

```html
<div class="bubble-output-container">
  <div class="bubble-result-container">
    <div class="vega-embed" id="c8ff3624-14a1-4ad8-b422-bbaa30bdba13">
    </div>

    <style>
      .vega-embed svg, .vega-embed canvas {
        border: 1px dotted gray;
      }

      .vega-embed .vega-actions a {
        margin-right: 6px;
      }
    </style>
  </div>
  <div class="bubble-error-container plain-error">
  </div>
  <div class="bubble-status-container hidden" style="height: 21px">
    ...
  </div>
</div>
```

I am under the impression that the package that is used to embed Altair's graphics is https://github.com/vega/ipyvega, but I am not familiar with Altair's source enough to tell for sure.

[...]

Issue #557, opened 2016-12-21
Plotly and cufflinks code


erdemalpkaya:
Hey,

Does anyone know plotly  ' .iplot' code is not working in Atom/Hydrogen or I missed something?
Below my code is working different platform but not in the Atom!

<img width="778" alt="ex_atom" src="https://cloud.githubusercontent.com/assets/24259112/21386898/fa84d38c-c7d9-11e6-965c-8a0ad53ddfb8.png">
<img width="580" alt="ex_web" src="https://cloud.githubusercontent.com/assets/24259112/21386899/fad062ac-c7d9-11e6-8eb8-97f0150e12aa.png">

n-riesco:
@rgbkrk If I remember correctly, `nteract` handles this through `application/vnd.plotly.v1+json`. Is that how the notebook does it? Is there anything we could reuse in  Hydrogen?

n-riesco:
Re #368

rgbkrk:
We need to put pressure on the packages to use UMD bundles in their `text/html` payloads. It's failing in Hydrogen there because it's expecting `require` to be from requirejs when the `require` in Hydrogen/Atom is the node commonjs require.

rgbkrk:
It's either that or Hydrogen has to provide a requirejs shim within the scope of the output somehow.

hz:
Besides Plotly, could Bokeh be introduced with hydrogen remote gateway support?

rgbkrk:
I've been using Bokeh in nteract, I don't see why it wouldn't work in Hydrogen unless there are still content security policies dictated by Atom that we need to play well with.

erdemalpkaya:
Sorry, I wrote again but I would like work atom and still the same issue just I want to know how can we fix the problem?

bu2:
@rgbkrk Hi, I confirm that Bokeh does not work in Atom/Hydrogen because of CSP.

I test the following example:

```
import numpy as np
from six.moves import zip
from bokeh.plotting import figure, show, output_notebook

N = 4000
x = np.random.random(size=N) * 100
y = np.random.random(size=N) * 100
radii = np.random.random(size=N) * 1.5
colors = ["#%02x%02x%02x" % (int(r), int(g), 150) for r, g in zip(50+2*x, 30+2*y)]
TOOLS="hover,crosshair,pan,wheel_zoom,box_zoom,reset,tap,save,box_select,poly_select,lasso_select"


output_notebook()
p = figure(tools=TOOLS)
[...]

Issue #368, opened 2016-07-19
Plotly doesn't display because of Content Security Policy 


jmatsushita:
Hi there,

When running 

```
var plt = require('plotly-notebook-js')

var myPlot = plt.createPlot([{ x: [1,2,3], y: [3,4,5] }], { title: 'Plotly in Jupyter!' });

$$html$$ = myPlot.render();
```

I get the following error message in the dev console:

```
Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self'". Either the 'unsafe-inline' keyword, a hash ('sha256-RZDXV0IqxSd/acUJ7Xz1gAlZt9WVh3exdjHW6bPU0zI='), or a nonce ('nonce-...') is required to enable inline execution.
(anonymous function) @ result-view.coffee:138
```

It works fine in IJavascript.

Cheer!

Jun

n-riesco:
cc/ @lgeiger How do you reckon? Should we wrap all the `text/html` results in a webview?
And if so, how about `image/svg`, I think SVG can contain JS?

rgbkrk:
> how about image/svg, I think SVG can contain JS?

SVG can indeed contain JS.

lgeiger:
If we get it to work with webviews it would be great! We should do this directly in `transformime`. Maybe using webviews would also fix #369.

I don't think we need webviews for SVG though.

n-riesco:
@rgbkrk @lgeiger Are there any cons to using webviews for all `text/html` content?

What are the cons from SVGs?

lgeiger:
> Are there any cons to using webviews for all text/html content?

I wanted to ask the same question. I never used webviews.
I'm fine with using webviews for SVGs as well if JS in SVGs is a commonly used thing.

n-riesco:
The physicist in me would be very pleased with an SVG version of:


![](https://upload.wikimedia.org/wikipedia/commons/6/6d/Translational_motion.gif)

I don't know of any problems with webview in Chrome/Electron/Atom. But it isn't a standard HTML tag, and I think it's safer if `transformine` would let the user decide whether to use it or not.

lgeiger:
> The physicist in me would be very pleased with an SVG version of:

That's a good argument 🎉 

I did a quick test with webviews in hydrogen but javascript didn't work. I used this [transformer](https://github.com/nteract/hydrogen/blob/master/lib/result-view.coffee#L228).
```coffeescript
HTMLTransform = (mimetype, data, document) ->
[...]

Issue #27, opened 2015-05-24
support interactive HTML widgets like Gadfly and ipywidgets


willwhitney:
...or anything else that uses scripts (e.g. `itorch.Plot`)

rgbkrk:
Likely precursor upstream: https://github.com/ipython/ipywidgets/issues/13

RobSalz:
Yes please!!!!

n-riesco:
The CSP policy in Atom v1.6.0 doesn't allow the injection of scripts.
Will interactive HTML widgetis in Atom ever be possible?

@rgbkrk How about nteract? Are there any plans?

rgbkrk:
We certainly plan to have it in nteract, though @jdfreder and I have only sketched out ways we can handle it.

Seems like it might be worth seeing if Atom can let you create an iframe or a WebView.

aviatesk:
Do you guys think that this package may help ? https://github.com/atom/loophole
I'm not sure how to apply this to our React components though.

wadethestealth:
@aviatesk all we need is the services api. I have a pr using the plugin api already. However we could be more like autocomplete and consume it

aviatesk:
Do you mean if there is

wadethestealth:
@aviatesk the package would be similar to both plotly and mathjax which both had cst errors

kylebarron:
> Seems like it might be worth seeing if Atom can let you create an iframe or a WebView.

Any suggestions in this line of thinking? I might test this out

PR #655, opened 2017-03-21
Use @nteract/tranforms

lgeiger:
A initial stab at using [`@nteract/tranforms`](https://github.com/nteract/nteract/tree/master/packages/transforms) as a replacement for `transformime` in order to address #557, #560, #27 and #368 and support even more mimetypes.

## Moved to a later time:
- [x] Use `@nteract/tranforms` for inspector (the easy part 😄 )
- [ ] Refactor the following parts as React Components:
  - [ ] Result View probably powered by [`@nteract/display-area`](https://github.com/nteract/nteract/tree/master/packages/display-area)
    - [ ] Spinner component
    - [ ] Component for inline result
    - [ ] Component for multiline result
    - [ ] Component for rich output (I think we could even remove it and use the multiline output for simplicity and to have a more consistent appearance)
  - [ ] Watch Sidebar
    - [ ] Watch View component

There's quite a bit of work left to do and I don't know when I can get to it, so I'm parking it here 🚗 
It would be super cool if someone wants to help with this effort 😇

### DIFF
--- a/lib/kernel.js
+++ b/lib/kernel.js
@@ -2,6 +2,7 @@
 
 import { Emitter } from "atom";
 import { observable, action } from "mobx";
+import { is, Map as ImmutableMap } from "immutable";
 
 import { log } from "./utils";
 import store from "./store";
@@ -13,7 +14,7 @@ export default class Kernel {
   @observable executionState = "loading";
   @observable inspector = {
     visible: false,
-    HTML: "",
+    bundle: new ImmutableMap(),
     height: 240
   };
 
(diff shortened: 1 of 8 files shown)


lgeiger:
I aimed a bit high in this PR 😄 
 
In my last commit I took a different approach to the problem and used `@nteract/transforms` as a drop-in replacement of `transformime` without any refactoring.
This way we can iterate on this and can use all the awesome transforms right now:

transforms
 
Now we can easily add support for [more mime types](https://github.com/nteract/nteract/tree/master/packages/) or even use `@nteract/transforms-full` if `apm` handles the install nicely (one transform has `node-canvas` as a native but optional dependency which may cause install problems).

I propose we go with this as is and move the refactorings into the future.
It would be cool if someone could take this branch for a spin since there may some minor style glitches I didn't caught.

lib/kernel.js
```diff
@@ -141,65 +138,20 @@ export default class Kernel {
...
-        data: {},
-        type: mime,
-        stream: "pyout"
-      };
-      result.data[mime] = data[mime];
-    }
+    const result = {
+      data: data,
+      stream: "pyout"
+    };
```
rgbkrk: Classic message spec before v5. 😉

PR #661, opened 2017-03-24
Add asciidoc as markup grammar

lgeiger:
During testing #637 asciidoc didn't work for me. Turns out I used the wrong syntax to define code blocks:
asciidoc

/cc @fasiha


rgbkrk:
Well that's awesome.

PR #665, opened 2017-03-24
Support language-markdown package

lgeiger:
Aaaaand another PR for markup language support.
(Every time I push a PR I find a new language we can easily support, I should have done them in batches 😂)

Once https://github.com/burodepeper/language-markdown/pull/172 is merged we can support the [`language-markdown`](https://github.com/burodepeper/language-markdown) package (a popular package as a replacement for the included markdown language):

markdown

PR #667, opened 2017-03-25
Listen for grammar changes of the editor and update store

lgeiger:
Noticed while testing #665 and frequently switching grammars:
grammar-update


BenRussert:
Looks good. I had noticed that issue as well. Nice that you found a fix without adding complexity.

PR #670, opened 2017-03-26
Disable commands if in mini text editor

### LINKED ISSUE #669: Uncaught TypeError: Cannot read property 'getSelectedText' of null
[Enter steps to reproduce:]

1. ...
2. ...

**Atom**: 1.15.0 ia32
**Electron**: 1.3.13
**OS**: Microsoft Windows 8.1
**Thrown From**: [Hydrogen](https://github.com/nteract/hydrogen) package 1.7.0

### Stack Trace

Uncaught TypeError: Cannot read property 'getSelectedText' of null

```
At C:\Users\Vinicius\.atom\packages\Hydrogen\lib\code-manager.js:42

TypeError: Cannot read property 'getSelectedText' of null
    at getSelectedText (/packages/Hydrogen/lib/code-manager.js:42:38)
    at Object.findCodeBlock (/packages/Hydrogen/lib/code-manager.js:230:24)
    at Object.run (/packages/Hydrogen/lib/main.js:324:35)
    at /packages/Hydrogen/lib/main.js:49:48)
    at CommandRegistry.module.exports.CommandRegistry.handleCommandEvent (/app.asar/src/command-registry.js:259:29)
    at /app.asar/src/command-registry.js:3:59
    at KeymapManager.module.exports.KeymapManager.dispatchCommandEvent (/app.asar/node_modules/atom-keymap/lib/keymap-manager.js:599:16)
    at KeymapManager.module.exports.KeymapManager.handleKeyboardEvent (/app.asar/node_modules/atom-keymap/lib/keymap-manager.js:390:22)
    at WindowEventHandler.module.exports.WindowEventHandler.handleDocumentKeyEvent (/app.asar/src/window-event-handler.js:106:36)
    at /app.asar/src/window-event-handler.js:3:59)
```

### Commands

```
     -2:21.2.0 julia-client:select-working-folder (ul.list-inline.tab-bar.inset-panel)
 10x -1:54.8.0 core:move-down (div.panels)
     -1:20.4.0 julia-client:run-block (input.hidden-input)
     -1:20.4.0 autocomplete-plus:cancel (atom-text-editor.editor.is-focused)
     -0:57.9.0 julia-client:open-console (atom-workspace.workspace.scrollbars-visible-always.theme-one-dark-syntax.theme-one-dark-ui)
     -0:57.8.0 autocomplete-plus:cancel (atom-text-editor.editor)
  8x -0:35.1.0 core:backspace (input.hidden-input)
     -0:28.3.0 console:evaluate (input.hidden-input)
     -0:28.3.0 autocomplete-plus:cancel (atom-text-editor.editor.is-focused)
  2x -0:26.7.0 hydrogen:run-and-move-down (ink-console.)
     -0:19.2.0 autocomplete-plus:cancel (atom-text-editor.editor.is-focused)
```

### Non-Core Packages

```
atom-python-run 0.7.3 
autocomplete-python 1.8.62 
Hydrogen 1.7.0 
ink 0.6.5 
julia-client 0.5.12 
kite 0.18.0 
language-julia 0.11.0 
latex-completions 0.3.2 
tool-bar 1.1.0 
uber-juno 0.1.1 
```

lgeiger:
~~This will probably fix #669~~. For more infos see https://github.com/nteract/hydrogen/issues/669#issuecomment-289307440

I was able to reproduce the error of #669 inside the settings panel. Before one was able to execute a Hydrogen command inside a settings text box.

The watch sidebar still works. ~~I haven't tested it with `julia-client` though.~~

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -42,7 +42,7 @@ const Hydrogen = {
 
     this.markerBubbleMap = {};
 
-    store.subscriptions.add(atom.commands.add('atom-text-editor', {
+    store.subscriptions.add(atom.commands.add('atom-text-editor:not([mini])', {
       'hydrogen:run': () => this.run(),
       'hydrogen:run-all': () => this.runAll(),
       'hydrogen:run-all-above': () => this.runAllAbove(),


lgeiger:
This doesn't fix #669 but it's still a valuable PR.

We have the problem that packages like `julia-client` can inject text editor instances into their UI and therefor Hydrogen commands can be executed while inside their UI instead of being in a active text editor. In this case `store.editor` is null.

I think we should therefor guard against the case that `store.editor` is null unless there's a better solution.

lgeiger:
> I think we should therefor guard against the case that `store.editor` is `null`

Probably the most elegant solution is to refactor `code-manager.js` to be pure functions (don't rely on `store.editor`, instead pass in a `editor` as a function argument). This helps also with testing in the long run 😄

rgbkrk:
Yeah doing pure functions in the future would be really good.

Issue #35, opened 2015-05-25
Watch kernel and runtime directories for updates


rgbkrk:
We could use [`fs.watch`](https://nodejs.org/api/fs.html#fs_fs_watch_filename_options_listener) to track kernel installations (or other runtimes being launched, e.g. the notebook). To make this simpler, there are higher level watch libraries like https://github.com/mikeal/watch or https://github.com/bevry/watchr.

BenRussert:
@rgbkrk I wonder if (almost two years later) you have any more thoughts on this? Is there anything in the nteract repo that might be a good model?

rgbkrk:
[Sidecar](https://github.com/nteract/sidecar) detects new runtime directories, though the code has grown stale (no one hacking on it, lots of outdated libraries, no shared knowledge of what's going on in there). For Hydrogen, I'd write the kernel checking all over again or make it part of `kernelspecs`.

lgeiger:
👍 For using `kernelspecs`!

Though at the moment it doesn't detect the default python installation so users need to run `python -m ipykernel install --user` so it can be detected.

I'm wondering if the easiest thing would be, if `kernelspecs` ran `python -m ipykernel install --user` if it can't detect a python kernel and then rerun current detection method.

rgbkrk:
We could bring the default python detection to kernelspecs though, which is currently buried away in old PRs of mine on nteract/nteract that I've closed.

lgeiger:
> We could bring the default python detection to kernelspecs though, which is currently buried away in old PRs of mine on nteract/nteract that I've closed.

Ahh sure I forgot!

For reference: https://github.com/nteract/nteract/pull/1384

lgeiger:
Atom now has a [file watching API](https://atom.io/docs/api/v1.21.1/PathWatcher). Since we're now using `kernelspecs` we could easily watch the directories returned by [`jupyter-paths`](https://github.com/nteract/jupyter-paths) for changes and update the kernels.

I'd be very happy if someone is interested in implementing this.

wadethestealth:
Commenting as a reminder for potentially picking this up

PR #698, opened 2017-04-05
KernelManager: Fix language detection

### LINKED ISSUE #696: IJavascript not detected by Atom on Mac
**Description:**
IJavascript not detected by Atom on Mac

**Steps to Reproduce:**

1. Install Hyderogen
2. Install IJavaScript
3. OPen atom and do Hydrogen: update

**Versions:**

Mac - 10.12.3
Atom    : 1.15.0
Electron: 1.3.13
Chrome  : 52.0.2743.82
Node    : 6.5.0
Hydrogen: 1.10.0

**Logs:**

Please post any error logs and the output of the developer tools as described in our [Debugging Guide](https://github.com/nteract/hydrogen/blob/master/TROUBLESHOOTING.md).

n-riesco:
* Use lowercase in language name comparisons

* Ensured that `grammarToLanguage(grammar)` returns the mapped language.

Fixes #696

### DIFF
--- a/lib/kernel-manager.js
+++ b/lib/kernel-manager.js
@@ -166,7 +166,7 @@ class KernelManager {
   }
 
   kernelSpecProvidesLanguage(kernelSpec, grammarLanguage) {
-    return kernelSpec.language === grammarLanguage;
+    return kernelSpec.language.toLowerCase() === grammarLanguage.toLowerCase();
   }
 
   getKernelSpecsFromSettings() {


n-riesco:
Oh! I thought the plan was to use grammar language names instead of kernel language names.

lib/utils.js
```diff
@@ -36,7 +36,9 @@ export function grammarToLanguage(grammar: ?atom$Grammar) {
     l => l.toLowerCase() === grammarLanguage
   );
 
-  return kernelLanguage ? kernelLanguage.toLowerCase() : grammarLanguage;
```
lgeiger: [`_.findKey`](https://lodash.com/docs/4.17.4#findKey) already returns the mapped language.

lgeiger: So I think this line can remain unchanged

n-riesco: The way I understood `languageMappings` is that if it was set to `{ "babel": "javascript"}`, then a kernel that declares it can run `babel`, will be regarded by Hydrogen as a kernel that can run `javascript`.

n-riesco: I'm going to update the spec for `grammarToLanguage` so you can see what I mean.

n-riesco: Oh! I see my confusion. Since we are mapping by kernel names, we need to return the kernel name!

n-riesco: I'm chasing another bug; even with this PR `{ "babel": "javascript"}` doesn't work.

n-riesco:
@lgeiger I've noticed that I need to close and open Atom for change in `kernelSpec` to take effect.

Shall I open an issue?

lgeiger:
Sorry, I wasn't able to help earlier, glad you figured it out 👍 

> I'm chasing another bug; even with this PR `{ "babel": "javascript"}` doesn't work.

You're right the language mappings are confusing, but we can't change the behavior since it would break all existing configs and cause even more confusion 😄 
I tried to improve the [documentation](https://github.com/nteract/hydrogen/blob/master/lib/config.js#L55) in my last PR to make it clearer.

> I've noticed that I need to close and open Atom for changes in kernelSpec to take effect.

You should be able to run "Hydrogen: Update Kernels" to trigger a update. Is it broken?
If you like, you can chime in on #35 for automatically detecting kernel spec changes.

PR #704, opened 2017-04-10
Add animated logo

BenRussert:
I thought it would be cool to have a version of the hydrogen logo where the electron orbited the proton. Might be good for a loading icon or something.


rgbkrk:
Cool!

lgeiger:
Cool!

Do you want to add it to our readme, similar to the logo in https://github.com/nteract/nteract? 

We can even try and use it for our autocomplete results if it's not too distracting having multiple spinning icons 😉

BenRussert:
The reason I started experimenting with this was I thought it would be a cool "spinner" during cell loading. Or even if a logo popped up and did one orbit before output is displayed (then disappears). 

As a user I'd like to have a clear way of knowing: I just ran something with hydrogen and not (for example) the Julia console. Although, I also don't want clutter. 

But for now, I thought this looked cool enough to share!

PR #717, opened 2017-04-15
Fix: Unmount react if result bubble is destroyed

lgeiger:
I'm not 100% sure if this is really necessary.

But if it is this fixes a memory leak since it unmounts React if the result bubble isn't needed anymore.


rgbkrk:
Oh didn't realize this, makes sense.

Issue #684, opened 2017-04-01
Support multiple language mappings?


maximsch2:
I'm using Hydrogen to execute code in Julia files and also wanted to try out [Weave.jl](https://github.com/mpastell/Weave.jl). It has it's own Atom language file that suggests using language mappings to get Hydrogen to execute the code: https://github.com/mpastell/language-weave

The issue is, if I set up custom mapping to enable Julia code execution in "weave.jl markdown" files, then I lose that ability in regular "julia" files. It would be great if language mappings would allow to either specify a list of types for the mapping or just the mapping was reversed.

n-riesco:
I think the language mappings suggested in https://github.com/mpastell/language-weave are meant for a very old version of Hydrogen.

If I understand correctly, the grammar name is "weave.jl markdown" and the kernel name is "julia". If so, the correct language mapppings are these:

```json
{ "weave.jl markdown": "julia", "pweave markdown": "python" }
```

If this doesn't work, please, could you post the error message?

lgeiger:
@n-riesco I think the language mappings from weave are correct.
The problem is that if a language mapping is present, Hydrogen doesn't recognize the original language. For example it only works with `weave.jl markdow` files and not with `julia` files anymore.

For the specific case of weave, instead of using a language mapping the best way would be if they could integrate with our new [multi language](https://medium.com/nteract/hydrogen-introducing-rich-multi-language-documents-b5057ff34efc) feature (which should be very easy).

n-riesco:
> I think the language mappings from weave are correct.

Oh! I keep confusing that the `languageMappings` setting maps **kernel names** into **grammar names**.

I think a solution to this issue would be to replace [these lines](https://github.com/nteract/hydrogen/blob/af85dc465281d7c85c68eea19d598dd9633bcf8d/lib/kernel-manager.js#L176-L178) with 

```js
    if (mappedLanguage) {
      return mappedLanguage === language || kernelLanguage.toLowerCase() === language;
    }
```

In this way, Hydrogen would still consider the julia kernel for julia files despite the `languageMappings` setting.

@maximsch2 Alternatively, you could create a custom kernel spec by adding something like this to the `KernelSpec` setting:

```json
      {
          "kernelspecs": {
              "weave.jl markdown": {
                  "spec": {
                      "argv": [
                          "julia",
                          "-i",
                          "-F",
                          "/home/user/.julia/v0.5/IJulia/src/kernel.jl",
                          "{connection_file}"
                      ],
[...]

PR #722, opened 2017-04-17
Fix kernel picker

n-riesco:
This PR fixes several issues triggered when multiple kernels for the same language are available.

1. **Issue with `languageMappings`**: in a system with two kernels (one for `javascript` and another for `babel`) and `languageMappings` set to `{"babel": "javascript"}`, the kernel picker should show both kernels.

2. **Issue with multilanguage files**: in a system with four kernels (`jp-babel`, `ijavascript`, `ipython2` and `ipython3`) and in a multilanguage file (with javascript and python code), the list of kernels shown in the menu picker should be updated according to the scope grammar.

This PR also fixes a minor issue with the status bar (it ensures the status bar is set to `idle` once all the main sockets have been connected).

I've split this PR so that each commit can be reviewed on its own.

### DIFF
--- a/lib/kernel-manager.js
+++ b/lib/kernel-manager.js
@@ -89,11 +89,9 @@ class KernelManager {
 
     log("KernelManager: startKernel:", displayName);
 
-    launchSpec(kernelSpec, { cwd: getEditorDirectory(store.editor) }).then(({
-      config,
-      connectionFile,
-      spawn
-    }) => {
+    launchSpec(kernelSpec, {
+      cwd: getEditorDirectory(store.editor)
+    }).then(({ config, connectionFile, spawn }) => {
       const kernel = new ZMQKernel(
         kernelSpec,
         grammar,
--- a/lib/kernel-manager.js
+++ b/lib/kernel-manager.js
@@ -10,7 +10,7 @@ import Config from "./config";
 import ZMQKernel from "./zmq-kernel";
 import KernelPicker from "./kernel-picker";
 import store from "./store";
-import { grammarToLanguage, getEditorDirectory, log } from "./utils";
+import { getEditorDirectory, log } from "./utils";
 
 class KernelManager {
   constructor() {
@@ -37,10 +37,9 @@ class KernelManager {
       }
     }
 
-    const language = grammarToLanguage(grammar);
-    this.getKernelSpecFor(language, kernelSpec => {
+    this.getKernelSpecForGrammar(grammar, kernelSpec => {
       if (!kernelSpec) {
-        const message = `No kernel for language \`${language}\` found`;
+        const message = `No kernel for grammar \`${grammar.name}\` found`;
         const description =
           "Check that the language for this file is set in Atom and that you have a Jupyter kernel installed for it.";
         atom.notifications.addError(message, { description });
@@ -52,7 +51,7 @@ class KernelManager {
   }
 
   startExistingKernel(grammar, connection, connectionFile, onStarted) {
-    const language = grammarToLanguage(grammar);
+    const language = grammar.name;
 
     log("KernelManager: startExistingKernel: Assuming", language);
 
@@ -86,9 +85,9 @@ class KernelManager {
   }
 
   startKernel(kernelSpec, grammar, onStarted) {
-    const language = grammarToLanguage(grammar);
+    const displayName = kernelSpec.display_name;
 
-    log("KernelManager: startKernelFor:", language);
+    log("KernelManager: startKernel:", displayName);
 
     launchSpec(kernelSpec, { cwd: getEditorDirectory(store.editor) }).then(({
       config,
@@ -104,7 +103,7 @@ class KernelManager {
       );
 
       kernel.connect(() => {
-        log("KernelManager: startKernelFor: kernel for", language, "connected");
+        log("KernelManager: startKernel:", displayName, "connected");
 
         store.newKernel(kernel);
 
@@ -147,6 +146,18 @@ class KernelManager {
     return callback([]);
   }
 
+  getAllKernelSpecsForGrammar(grammar, callback) {
+    if (grammar) {
+      return this.getAllKernelSpecs(kernelSpecs => {
+        const specs = kernelSpecs.filter(spec =>
+          this.kernelSpecProvidesGrammar(spec, grammar));
+
+        return callback(specs);
+      });
+    }
+    return callback([]);
+  }
+
   getKernelSpecFor(language, callback) {
     if (!language) {
       return;
@@ -169,8 +180,47 @@ class KernelManager {
     });
   }
 
+  getKernelSpecForGrammar(grammar, callback) {
+    if (!grammar) {
+      return;
+    }
+
+    this.getAllKernelSpecsForGrammar(grammar, kernelSpecs => {
+      if (kernelSpecs.length <= 1) {
+        callback(kernelSpecs[0]);
+        return;
+      }
+
+      if (!this.kernelPicker) {
+        this.kernelPicker = new KernelPicker(onUpdated =>
+          onUpdated(kernelSpecs));
+        this.kernelPicker.onConfirmed = ({ kernelSpec }) =>
+          callback(kernelSpec);
+      }
+      this.kernelPicker.toggle();
+    });
+  }
+
   kernelSpecProvidesLanguage(kernelSpec, grammarLanguage) {
     return kernelSpec.language.toLowerCase() === grammarLanguage.toLowerCase();
+  }
+
+  kernelSpecProvidesGrammar(kernelSpec, grammar) {
+    if (!grammar || !grammar.name || !kernelSpec || !kernelSpec.language) {
+      return false;
+    }
+    const grammarLanguage = grammar.name.toLowerCase();
+    const kernelLanguage = kernelSpec.language.toLowerCase();
+    if (kernelLanguage === grammarLanguage) {
+      return true;
+    }
+
+    const mappedLanguage = Config.getJson("languageMappings")[kernelLanguage];
+    if (!mappedLanguage) {
+      return false;
+    }
+
+    return mappedLanguage.toLowerCase() === grammarLanguage;
   }
 
   getKernelSpecsFromSettings() {
--- a/lib/kernel-manager.js
+++ b/lib/kernel-manager.js
@@ -150,7 +150,8 @@ class KernelManager {
     if (grammar) {
       return this.getAllKernelSpecs(kernelSpecs => {
         const specs = kernelSpecs.filter(spec =>
-          this.kernelSpecProvidesGrammar(spec, grammar));
+          this.kernelSpecProvidesGrammar(spec, grammar)
+        );
 
         return callback(specs);
       });
@@ -169,13 +170,13 @@ class KernelManager {
         return;
       }
 
-      if (!this.kernelPicker) {
-        this.kernelPicker = new KernelPicker(onUpdated =>
-          onUpdated(kernelSpecs)
-        );
-        this.kernelPicker.onConfirmed = ({ kernelSpec }) =>
-          callback(kernelSpec);
+      if (this.kernelPicker) {
+        this.kernelPicker.kernelSpecs = kernelSpecs;
+      } else {
+        this.kernelPicker = new KernelPicker(kernelSpecs);
       }
+
+      this.kernelPicker.onConfirmed = ({ kernelSpec }) => callback(kernelSpec);
       this.kernelPicker.toggle();
     });
   }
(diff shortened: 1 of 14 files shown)


lgeiger:
Thanks, I'll review tomorrow 👍 

> 1. **Issue with `languageMappings`**: the kernel picker should show both kernels.

Thanks for catching that!

> 2. **Issue with multilanguage files**: the list of kernels shown in the menu picker should be updated according to the scope grammar.

I just checked on master, in multilanguage files the kernel list is updated correctly. Am I missing something?

> This PR also fixes a minor issue with the status bar (it ensures the status bar is set to `idle` once all the main sockets have been connected).

That's awesome! 🚀

n-riesco:
> > 2. **Issue with multilanguage files**: the list of kernels shown in the menu picker should be updated according to the scope grammar.
>
> I just checked on master, in multilanguage files the kernel list is updated correctly. Am I missing something?

You need to make sure `languageMappings` isn't set and that there are multiple kernels for multiple languages.

This is what I see in master:

peek 2017-04-17 23-49

And with this PR:

peek 2017-04-17 23-53

lgeiger:
> You need to make sure languageMappings isn't set and that there are multiple kernels for multiple languages.

Strange, that's exactly what I did. I have the four kernels you mentioned installed and I get this behavior:
test

@n-riesco Are there any other kernels with multiple languages for the same Atom grammar?
I'm just asking because in previous PR's I tried to remove the usage of Atom's grammar as much as possible since it makes the code a lot simpler and easier to understand having only one source of language.
So for `jp-babel` I actually would expect it to have `javascript` as a language instead of `babel` similar to Python 2 and 3 which both return `python` as a language. This would also fix this issue without needing us to handle this case inside Hydrogen. Don't get me wrong, I'm just trying to think of the simplest solution if `jp-kernel` and `ijavascript` are the only kernels who have this problem.

lib/main.js
```diff
@@ -95,15 +97,15 @@ const Hydrogen = {
...
+            store.updateEditorAndGrammar(editor);
           })
         );
 
         if (isMultilanguageGrammar(editor.getGrammar())) {
           editorSubscriptions.add(
             editor.onDidChangeCursorPosition(
               _.debounce(() => {
-                store.setGrammar(editor);
+                store.updateEditorAndGrammar(editor);
```
lgeiger: Why did you merge `setGrammar` and `updateEditor` into `updateEditorAndGrammar`?

In this case the editor doesn't change, yet we update it in the store. Theoretically this should be fine but I'm not sure if it can cause mobx to trigger a rerender somewhere.

lib/store/index.js
```diff
@@ -40,29 +39,31 @@ class Store {
     this.runningKernels = new Map();
   }
 
-  @action updateEditorAndGrammar(editor: ?atom$TextEditor) {
-    this.editor = editor;
-
-    if (!this.editor) {
```
lgeiger: I think just using `editor` instead of `this.editor` should make flow happy.

lib/kernel-manager.js
```diff
@@ -169,13 +170,13 @@ class KernelManager {
...
-      if (!this.kernelPicker) {
-        this.kernelPicker = new KernelPicker(onUpdated =>
-          onUpdated(kernelSpecs)
-        );
-        this.kernelPicker.onConfirmed = ({ kernelSpec }) =>
-          callback(kernelSpec);
+      if (this.kernelPicker) {
+        this.kernelPicker.kernelSpecs = kernelSpecs;
+      } else {
+        this.kernelPicker = new KernelPicker(kernelSpecs);
```
lgeiger: That's a lot cleaner 👍

n-riesco:
> Strange, that's exactly what I did. I have the four kernels you mentioned installed and I get this behavior

In your gif, the kernel picker is only shown once, so there is no chance to trigger this bug.
To test this issue, I'm using custom kernel specs:

```js
{
    "kernelspecs": {
        "ijavascript": {
            "spec": {
                "env": {},
                "argv": [
                    "ijskernel",
                    "--hide-undefined",
                    "--protocol=5.0",
                    "{connection_file}"
                ],
                "language": "javascript"
            },
        },
        "babel": {
            "spec": {
                "display_name": "jp-Babel",
                "env": {},
                "argv": [
                    "jp-babel-kernel",
                    "--hide-undefined",
                    "--protocol=5.0",
                    "{connection_file}"
                ],
                "language": "javascript"
            },
        },
        "ipython2": {
            "spec": {
                "display_name": "IPython 2",
                "env": {},
                "argv": [
                    "ipython",
                    "kernel", "--colors='Linux'",
                    "-f",
                    "{connection_file}"
                ],
                "language": "python"
            }
        },
        "ipython3": {
            "spec": {
                "display_name": "IPython 3",
                "env": {},
                "argv": [
                    "ipython3",
                    "kernel",
                    "-f",
                    "{connection_file}"
                ],
                "language": "python"
            }
        }
    }
}
```


> @n-riesco Are there any other kernels with multiple languages for the same Atom grammar?

Python 2 and 3 ? (I guess I'm misunderstanding your question)

> I'm just asking because in previous PR's I tried to remove the usage of Atom's grammar as much as possible since it makes the code a lot simpler and easier to understand having only one source of language.

I think we misunderstood each other [here](https://github.com/nteract/hydrogen/issues/684#issuecomment-290916923).

I think it's better to use only Atom's grammar for several reasons:
- grammar language is what Atom gives us (if we index by grammar, there is no need to convert grammar to kernel language)
- grammar language is what Atom shows on the status bar and what user see
- if Hydrogen introduces its own language names, all users need to learn the difference between kernel language and grammar language
- if Hydrogen uses grammar languages, only the users that set `languageMappings` and `kernelSpec` need to be aware of the difference.

Anyway, since this PR doesn't change how kernels are indexed, I think we should move this discussion to another issue.

> So for jp-babel I actually would expect it to have javascript as a language instead of babel similar to Python 2 and 3 which both return python as a language. This would also fix this issue without needing us to handle this case inside Hydrogen.

No, it wouldn't solve the issue. `jp-babel` is not the cause of the issue. Setting `languageMappings` is the cause of the issue.

lib/main.js
n-riesco: It was a left-over from a refactor I tried. I've just pushed a commit to restore `setGrammar`.

lgeiger:
Thanks for bearing with me. I see the issue with the kernel picker now 👍

👍 For moving the discussion into a a issue.

I'm only talking about the issue with language mappings, happening if two kernels use different languages for the same Atom grammar. 

> >@n-riesco Are there any other kernels with multiple languages for the same Atom grammar?

>Python 2 and 3 ? (I guess I'm misunderstanding your question)

Python 2 and 3 both use `language: python` in their kernelspecs. IJavascript uses `language: javascript` where as `jp-babel` uses `language: babel`.

This difference makes it hard to index kernels with their kernel language. To fix this we have to [index them depending on the mappings](https://github.com/nteract/hydrogen/pull/722/files#diff-c60fec2a9c618473021ddf650c446ef5R29). But this imposes a few problems:

We now have to [iterate over all kernels](https://github.com/nteract/hydrogen/pull/722/files#diff-c60fec2a9c618473021ddf650c446ef5R18) to get the current kernel (actually not that big of a deal 😄).
But since the indexing of kernels depends on the settings, we can get into a inconsistent state if the settings change. To reproduce a similar problem, consider the kernelspec settings `{"javascript":"babel es6 javascript"}` (used together with [`language-babel`](https://github.com/gandm/language-babel)):
shutdown

If `jp-babel` and IJavascript both would  both use `language: javascript` and are the only kernels with different kernel languages for the same Atom grammar, we wouldn't need to handle this case and avoid the problems described above.

n-riesco:
> But since the indexing of kernels depends on the settings, we can get into a inconsistent state if the settings change. To reproduce a similar problem, consider the kernelspec settings {"javascript":"babel es6 javascript"} (used together with language-babel)

It wouldn't depend, if it was indexed by Atom's grammar language. But since this is a PR to fix a bug, I think we should postpone the indexing discussion for another issue.

> If jp-babel and IJavascript both would both use language: javascript and are the only kernels with different kernel languages for the same Atom grammar, we wouldn't need to handle this case and avoid the problems described above.

Again, this issue isn't about `jp-babel`. In fact, `languageMappings` was introduced to help Hydrogen use the Elixir kernel. Jupyter doesn't specify that the language in a kernel spec has to match Atom's grammar name. And for that reason, `languageMappings` is necessary.

n-riesco:
Let's not forget this PR is a bug fix. It doesn't add or remove any functionality. It only requires a patch number update to be released.

lgeiger:
> Let's not forget this PR is a bug fix. It doesn't add or remove any functionality. It only requires a patch number update to be released.

Sure, I really want to merge it right away and keep iterating. (I'm not as opinionated about how kernels should be indexed as it might have come across)

But this PR introduces a exception happening if a kernel requiring a language mapping is shut down or restarted. Take a look at the GIF.

Sorry about my brevity I can take a look at it in more detail during the weekend.

n-riesco:
@lgeiger

The issue with updating `languageMappings` while running a kernel isn't introduced by this PR. It is already present in master!

This PR just fixes Hydrogen's algorithm to match kernels and Atom's grammars so that it honours the setting `languageMappings`.

Issues about:

- whether Hydrogen's store should index running kernels by Atom's grammar or kernel's language;
- or whether changes in the `kernelSpec` setting  require to restart Atom (as I mentioned [here](https://github.com/nteract/hydrogen/pull/698#issuecomment-291804071))
- or whether changes in the `languageMapping` setting should be prevented while there are kernels running

do not belong to this PR. They are issues already present in Hydrogen's latest release.

lgeiger:
> The issue with updating languageMappings while running a kernel isn't introduced by this PR. It is already present in master!

Nico you're right, I shouldn't have mentioned it.

> They are issues already present in Hydrogen's latest 

I agree 💯 

> This PR just fixes Hydrogen's algorithm to match kernels and Atom's grammars so that it honours the setting languageMappings.

It does but it introduces another serious exception.
With a language mapping present (Eg `"javascript":"babel es6 javascript"}`) and kernelspecs similar to [this](https://github.com/nteract/hydrogen/pull/722#issuecomment-294964998).

**master:**
master

**this PR:**
pr

lgeiger:
I'm not sure how to fix this problem.

I created [this branch](https://github.com/nteract/hydrogen/compare/master...lgeiger:fix-mappings-minimal?expand=1) which removes the commits related to the language mapping bug, which unfortunately introduce the exception above. If we don't find a solution to this problem we can go with this and handle the language mapping bug later.

n-riesco:
@lgeiger Both issues (the one in master and the one in the PR) have the same underlying cause: users shouldn't be allow to update `languageMappings` while there are kernels running.

I think, we've spent already as much time discussing this as what it takes to fix it. I'll look now into how to prevent users from updating `languageMappings` while there are kernels running and update this PR.

lgeiger:
> Both issues (the one in master and the one in the PR) have the same underlying cause:
users shouldn't be allowed to update languageMappings while there are kernels running.

I think the solution is much simpler. In the example I linked mentioned I didn't change anything in the language mappings while a kernel was running. I suspect since the kernel indexing changed [here](https://github.com/nteract/hydrogen/pull/722/files#diff-c60fec2a9c618473021ddf650c446ef5R29) the kernel gets not properly deleted [here](https://github.com/nteract/hydrogen/pull/722/files#diff-c60fec2a9c618473021ddf650c446ef5L54).

n-riesco:
Here's a screencast of the issue in master:

peek 2017-04-21 12-21

n-riesco:
@lgeiger As I said on slack, only after talking to you I noticed that `shutdown-kernel` was broken in this PR (even without changing `languageMappings`).

This commit https://github.com/nteract/hydrogen/pull/722/commits/024c21623f6f7eb100a9be4a950041e3bd000792 fixes the issue.

Are there any other things you'd like me to change?

lgeiger:
I'm going to merge this since  it doesn't introduce new bugs so we can keep moving forward.

Allow me one more comment 😄 :

> Again, this issue isn't about jp-babel. In fact, languageMappings was introduced to help Hydrogen use the Elixir kernel. Jupyter doesn't specify that the language in a kernel spec has to match Atom's grammar name. And for that reason, languageMappings is necessary.

It somehow is. Of course `languageMappings` is necessary. I don't say by any means kernels should adhere to Atoms language names (eg MagicPython, Babel ES6 Javascript.....). The thing I am suggesting is that kernels targeting the same language should also use a consistent language name.

This would make a mapping like `{"babel":"babel es6 javascript","javascript":"babel es6 javascript", "}` unnecessary. Instead one would use the mapping `{"some common kernel language shared by IJavascript and jp-babel":"babel es6 javascript"}`. This would save us the hassle of handling the above mentioned case and will keep the code a lot shorter and understandable.

n-riesco:
Thx!

PR #732, opened 2017-04-21
Dont allow language mappings change while kernels are running

n-riesco:
Opened a new PR for this (as discussed on slack)

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -47,6 +47,30 @@ const Hydrogen = {
 
     this.markerBubbleMap = {};
 
+    let skipLanguageMappingsChange = false;
+    store.subscriptions.add(
+      atom.config.onDidChange(
+        "Hydrogen.languageMappings",
+        ({ newValue, oldValue }) => {
+          if (skipLanguageMappingsChange) {
+            skipLanguageMappingsChange = false;
+            return;
+          }
+
+          if (store.runningKernels.size != 0) {
+            skipLanguageMappingsChange = true;
+
+            atom.config.set("Hydrogen.languageMappings", oldValue);
+
+            atom.notifications.addError("Hydrogen", {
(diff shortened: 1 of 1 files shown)


lgeiger:
> Opened a new PR for this (as discussed on slack)

Whoops I ment a new PR on top of master not on #722 . Sorry, this needs a rebase now.

n-riesco:
No problem, the rebase to master didn't have any conflicts.

lib/main.js
```diff
@@ -47,6 +47,30 @@ const Hydrogen = {
...
+        ({ newValue, oldValue }) => {
+          if (skipLanguageMappingsChange) {
+            skipLanguageMappingsChange = false;
+            return;
+          }
+
+          if (store.runningKernels.size != 0) {
+            skipLanguageMappingsChange = true;
+
+            atom.config.set("Hydrogen.languageMappings", oldValue);
```
lgeiger: We may be able to use [`atom.config.transact`](https://atom.io/docs/api/v1.16.0/Config#instance-transact)  here to prevent calling `onDidChange` again which, would be a bit clearer.

n-riesco: OK, I'll have a look.

lgeiger: Thinking about it again, this might not be so easy since we call it inside `onDidChange` and not inside `onWillChange`.

n-riesco: I tried, but it doesn't do what we want:
> Suppress calls to handler functions registered with ::onDidChange and ::observe for the duration of callback. **After callback executes, handlers will be called once if the value for their key-path has changed**.

lgeiger: Yeah, I spoke to fast, we would probably need to call it in a `onWillChange` handler but `atom.config` doesn't has one.

Issue #657, opened 2017-03-21
Roadmap


lgeiger:
The [old roadmap](https://github.com/nteract/hydrogen/wiki/Roadmap) is due for an update:

# Roadmap 🚀 

## Short term 🔜 

- [x] Fix issues left from MozSprint. Great for first time contributors: #828

Our documentation is due to a overhaul:
- [x] Add Hydrogen section to the nteract.io website: https://github.com/nteract/nteract.io/pull/7
Preview: https://nteractio-tssddztkpz.now.sh/atom
- [x] Add more content to the nteract.io website: https://github.com/nteract/nteract.io/pull/7 
- [x] Improve our README structure: #654
- [x] Add a Contributor Guide #653 
- [ ] Create a new to programming section #702

## Medium term

Since #627 there've been a lot of refactoring throughout the code base in order to ultimately rely on more nteract component and to make Hydrogen more maintainable and easier to test.
To get there, we're going to want to make some targeted refactorings, namely:

- [x] In continuation of #655 we should refactor the following as real react components:
  - [x] Result View probably powered by [`@nteract/display-area`](https://github.com/nteract/nteract/tree/master/packages/display-area)
    - [x] Spinner component
    - [x] Component for inline result
    - [x] Component for multiline result
    - [x] Component for rich output (I think we could even remove it and use the multiline output for simplicity and to have a more consistent appearance)
  - [x] Watch Sidebar
    - [x] Watch View component
- [x] Refactor List Views to use [atom-select-list](https://github.com/atom/atom-select-list) instead of the deprecated atom-space-pen-views which relies on jQuery. This will reduce the package size by about 2 MB @BenRussert 
- [ ] Reduce size of `main.js` since after the above everything should be a lot more modular
- [ ] Once [kernelspecs](https://github.com/nteract/kernelspecs) can detect the default python installation we should switch to it for kernel detection. This would remove `jupyter` as an external dependency and make kernel detection faster.

## Long term

I would love to switch to [enchannel-zmq](https://github.com/nteract/enchannel-zmq-backend) and [rx-jupyter](https://github.com/nteract/rx-jupyter) for kernel messaging

## Ongoing efforts

Hydrogen is quite stable but still largely untested. In continuation of #650, #648, #656:
- Add more tests (will become easier once more of the above points are completed)
- Add more flow types

Any help is greatly appreciated and needed! Feel free to edit or comment on this issue and claim bullet points or ask questions 😄 🎉

[...]

PR #742, opened 2017-04-22
Refactor watch sidebar to remove space pen views

BenRussert:
This is a WIP to move from space-pen-views to [atom-select-list](https://github.com/atom/atom-select-list)

Hoping for feedback as the atom-select-list design pattern examples I found were all a bit different than our previous watch-picker.

I ended up going with what seemed simplest to me (removing watches-picker.js entirely) in order to get something working and go from there. (Tests still pass and my own manual testing everything seems ok)

As mentioned in #657, the watch functionality could probably use a refactor soon since it seems like a good use case for our mobx store to keep track of state.


lib/watch-sidebar.js
```diff
@@ -106,13 +105,34 @@ export default class WatchSidebar {
...
     }));
-    WatchesPicker.onConfirmed = item => {
-      this.watchViews[item.value].destroy();
-      this.watchViews.splice(item.value, 1);
-      if (this.watchViews.length === 0) this.addWatch();
-    };
-    WatchesPicker.setItems(watches);
-    WatchesPicker.toggle();
+    watchesPicker = new SelectListView({
+      items: watches,
```
lgeiger: The search throws a error.
I think we need to add 
```js
filterKeyForItem: (watch) => watch.name
```

lib/watch-sidebar.js
```diff
@@ -106,13 +105,34 @@ export default class WatchSidebar {
...
+      elementForItem: watch => {
+        const element = document.createElement("li");
+        element.textContent = watch.name;
+        element.dataset = watch.name;
```
lgeiger: Is this line really necessary?

lib/watch-sidebar.js
```diff
@@ -106,13 +105,34 @@ export default class WatchSidebar {
...
+        return element;
+      },
+      didConfirmSelection: watch => {
+        this.watchViews[watch.value].destroy();
+        this.watchViews.splice(watch.value, 1);
+        modalPanel.hide();
+        modalPanel.destroy();
```
lgeiger: Why do we need to hide the panel before we destroy it?

`destroy` may needs to be called on `watchesPicker` as well.

lib/watch-sidebar.js
```diff
@@ -106,13 +105,34 @@ export default class WatchSidebar {
...
+      items: watches,
+      elementForItem: watch => {
+        const element = document.createElement("li");
+        element.textContent = watch.name;
+        element.dataset = watch.name;
+        return element;
+      },
+      didConfirmSelection: watch => {
+        this.watchViews[watch.value].destroy();
+        this.watchViews.splice(watch.value, 1);
```
lgeiger: We should add back
```js
if (this.watchViews.length === 0) this.addWatch();
```

lib/watch-sidebar.js
```diff
@@ -106,13 +105,34 @@ export default class WatchSidebar {
...
+        modalPanel.hide();
+        modalPanel.destroy();
+        previouslyFocusedElement.focus();
+      },
+      itemsClassList: ["watch-picker-item"],
+      didCancelSelection: () => {
+        modalPanel.hide();
+        modalPanel.destroy();
+        previouslyFocusedElement.focus();
```
lgeiger: `previouslyFocusedElement` may be null. We should guard against that case.

lib/watch-sidebar.js
```diff
@@ -106,13 +105,34 @@ export default class WatchSidebar {
...
+        return element;
+      },
+      didConfirmSelection: watch => {
+        this.watchViews[watch.value].destroy();
+        this.watchViews.splice(watch.value, 1);
+        modalPanel.hide();
+        modalPanel.destroy();
+        previouslyFocusedElement.focus();
+      },
+      itemsClassList: ["watch-picker-item"],
```
lgeiger: We probably don't need this since we don't use this class for any styling.

lib/watch-sidebar.js
BenRussert: Just the dataset line, 113 right? Left in from an example, but does not seem necessary.

lib/watch-sidebar.js
BenRussert: The reason I did this was to avoid accidentally adding null string watches that can't be removed. We could solve this another way on a future pr though.

lib/watch-sidebar.js
BenRussert: Any ideas?

lib/watch-sidebar.js
BenRussert: Great catch here, this is fixed.

lib/watch-sidebar.js
lgeiger: Why can't they be removed?

This line only ensures that there's always one empty watch view to start with.

lib/watch-sidebar.js
lgeiger: `if (previouslyFocusedElement) previouslyFocusedElement.focus();` should do the job.

lib/watch-sidebar.js
BenRussert: I see what you mean now, I meant that last one cant be removed. I was thinking you expected the focus to change to the add watch bar (which is a side effect of calling `addWatch`)

lib/watch-sidebar.js
```diff
@@ -102,17 +102,41 @@ export default class WatchSidebar {
...
-    };
-    WatchesPicker.setItems(watches);
-    WatchesPicker.toggle();
+    const watches = this.watchViews
+      .map((v, k) => ({
+        name: v.getCode(),
+        value: k
+      }))
+      .filter(obj => obj.value !== 0 || obj.name !== "");
+
```
BenRussert: We always keep one empty string watch if no other watches, but the picker should show empty if this is the only watch left. You should be able to remove empty watches other than the placeholder mentioned, see gif for examples of what I am saying:

last-element

lib/watch-sidebar.js
BenRussert: @lgeiger I fixed this and added a comment/gif above.

lib/watch-sidebar.js
lgeiger: Very nice!

lib/watch-sidebar.js
```diff
@@ -102,17 +102,41 @@ export default class WatchSidebar {
...
+    watchesPicker = new SelectListView({
```
lgeiger: That's interesting. I would expect this to fail, maybe babel fixes this for us. Anyway, I think we should use
```js
const watchesPicker = new SelectListView({
```

lib/watch-sidebar.js
```diff
@@ -102,17 +102,41 @@ export default class WatchSidebar {
...
+      items: watches,
+      elementForItem: watch => {
+        const element = document.createElement("li");
+        element.textContent = watch.name;
```
lgeiger: For deleting empty watches a placeholder text would be a nice touch. Something similar to
```js
element.textContent = watch.name || "<empty>";
```

lgeiger:
I just had a look at https://github.com/atom/line-ending-selector/pull/37. They reuse the same SelectListView instance multiple times. I don't know what approach is best for our use case, but I think we should stick to what Atom does.

But we should do this in a future PR.

lib/watch-sidebar.js
BenRussert: Whoops.

lib/watch-sidebar.js
BenRussert: Great idea!

BenRussert:
> I just had a look at atom/line-ending-selector#37. They reuse the same SelectListView instance multiple times. I don't know what approach is best for our use case, but I think we should stick to what Atom does.

The majority of time spent on this pr by far was sorting through the few examples out there and trying to figure out what the best design pattern was for our situation. I will be looking for ways to improve this and for the other `space-pen-views` dependent modules.

lgeiger:
> The majority of time spent on this pr by far was sorting through the few examples out there and trying to figure out what the best design pattern was for our situation. I will be looking for ways to improve this and for the other space-pen-views dependent modules.

Awesome! No rush, this just came to my mind and I wanted to write it down so I don't forget where to look for 😄

PR #743, opened 2017-04-24
Use the busy-signal service

cbarrick:
This PR migrates the status bar indicator from a custom component to the [busy-signal](https://github.com/steelbrain/busy-signal) service, a part of the [linter](https://github.com/steelbrain/linter) family of packages.

Since linter is one of the most installed packages in Atom, it makes most sense to converge around their busy-signal service. Most Atom user will already have it installed. This also helps declutter the status bar.

Here's a screenshot:
hydrogen-busy-signal

This PR does three things:
1. Introduces the code to use the `busy-signal` service.
2. Removes the old custom status bar component.
3. Introduces `atom-package-dep` to express the dependency on `busy-signal` and offer automatic installation.

I'm fairly inexperienced developing for Atom, so please review carefully.

### DIFF
--- a/lib/store/index.js
+++ b/lib/store/index.js
@@ -1,7 +1,7 @@
 /* @flow */
 
 import { CompositeDisposable } from "atom";
-import { observable, computed, action } from "mobx";
+import { observable, computed, action, reaction } from "mobx";
 import { isMultilanguageGrammar, getEmbeddedScope } from "./../utils";
 
 import Config from "./../config";
@@ -13,6 +13,7 @@ class Store {
   @observable runningKernels = new Map();
   @observable editor = atom.workspace.getActiveTextEditor();
   @observable grammar: ?atom$Grammar;
+  @observable busySignalRegistry = null;
 
   @computed get kernel(): ?Kernel {
     for (let kernel of this.runningKernels.values()) {
@@ -26,6 +27,7 @@ class Store {
 
   @action newKernel(kernel: Kernel) {
     const mappedLanguage = Config.getJson("languageMappings")[kernel.language];
+    this._registerBusySignal(kernel);
     this.runningKernels.set(mappedLanguage || kernel.language, kernel);
   }
 
@@ -69,6 +71,27 @@ class Store {
     }
 
     this.grammar = grammar;
+  }
+
+  @action setBusySignalRegistry(registry) {
+    this.busySignalRegistry = registry;
+    this.runningKernels.forEach(this._registerBusySignal);
+  }
+
+  _registerBusySignal(kernel: Kernel) {
+    if (this.busySignalRegistry) {
+      const busySignalProvider = this.busySignalRegistry.create();
+      reaction(
+        () => kernel.executionState,
+        state => {
+          busySignalProvider.clear();
+          if (state != "idle") {
+            const message = kernel.displayName + " is " + state;
+            busySignalProvider.add(message);
+          }
+        }
+      );
+    }
   }
 }
 
--- a/lib/components/status-bar.js
+++ b/lib/components/status-bar.js
@@ -9,9 +9,7 @@ type Props = { store: { kernel: ?Kernel }, onClick: Function };
 
 const StatusBar = observer(({ store: { kernel }, onClick }: Props) => {
   if (!kernel) return null;
-  return (
-    <a onClick={onClick}>{kernel.displayName} | {kernel.executionState}</a>
-  );
+  return <a onClick={onClick}>{kernel.displayName}</a>;
 });
 
 export default StatusBar;
--- a/lib/main.js
+++ b/lib/main.js
@@ -5,6 +5,7 @@ import { Emitter, CompositeDisposable } from "atom";
 import _ from "lodash";
 import { autorun } from "mobx";
 import React from "react";
+import { install as installPackageDeps } from "atom-package-deps";
 
 import ResultView from "./result-view";
 import SignalListView from "./signal-list-view";
@@ -42,6 +43,7 @@ const Hydrogen = {
   watchSidebarIsVisible: false,
 
   activate() {
+    installPackageDeps("Hydrogen");
     this.emitter = new Emitter();
     this.inspector = new Inspector();
 
--- a/lib/main.js
+++ b/lib/main.js
@@ -165,6 +165,10 @@ const Hydrogen = {
     return this.hydrogenProvider;
   },
 
+  consumeBusySignal(registry) {
+    store.setBusySignalRegistry(registry);
+  },
+
   consumeStatusBar(statusBar: atom$StatusBar) {
     const statusBarElement = document.createElement("div");
     statusBarElement.className = "inline-block";
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "@nteract/transform-plotly": "1.1.0",
     "@nteract/transforms": "^1.0.7",
     "atom-space-pen-views": "^2.0.5",
+    "atom-package-deps": "^4.6.0",
     "escape-string-regexp": "^1.0.5",
     "immutable": "^3.8.1",
     "jmp": "^0.7.5",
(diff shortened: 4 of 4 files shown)


cbarrick:
The CI is failing because I have not added type annotations when consuming the busy-signal service. I'm not sure how to get the types from another package.

There's an additional failure dealing with a `$FlowFixMe` comment. I have no idea what's up with that.

rgbkrk:
Whoa! Thanks for the contribution. It's also good to know that comes from linter. This whole time I thought I was getting that from Nuclide.

nikitakit:
Hydrogen could definitely use a slimmed down presence in the status bar. One thing I'm curious about (having not run the code yet) is whether this provides any way to know what kernel you're connected to if the current state is idle. I have at least 3 python environments across multiple machines, and I use the status bar to keep track of which (if any) is currently connected. Also the remote kernels tend to enter a reconnecting or dead state if switch WiFi networks, so there's actually more status states than just idle/busy. I'm not sure how busy-signal handles inactive sources.

cbarrick:
> One thing I'm curious about (having not run the code yet) is whether this provides any way to know what kernel you're connected to if the current state is idle. 

At the moment there is no way to tell what kernel you're using when idle. IMO the "select kernel" menu is sufficient for me to garuntee which kernel I have running.

> Also the remote kernels tend to enter a reconnecting or dead state if switch WiFi networks, so there's actually more status states than just idle/busy.

Currently a message is sent to the busy signal whenever the state is anything but idle. If you have multiple kernels, each controls a separate busy-signal provider, meaning their messages are independent.

nikitakit:
I should mention that currently "select kernel" does not interact at all with remote kernels (it always lists the local ones).

n-riesco:
Below is a screencast that shows a few of the issues with this PR:
- (as commented by @nikitakit) this PR doesn't show what's the active kernel
- 'busy-signal`'s history tooltip could misled users into thinking kernels are always busy

I'm also concerned about:
- discoverability (this is not blocking for me)
- the tooltip getting too crowded (specially when running multiple kernels simultaneously)
- and that we lose functionality (this is a blocking change for me):
  - to switch the running kernel by clicking on the status bar (with this PR I have to run myself the commands to shutdown the running kernel and start a new kernel)
  - to tell whether there is a running kernel or which kernel is associated with the current file

@cbarrick Any ideas how we could show which kernel is the active kernel and keep the functionality to switch kernels?

---

peek 2017-04-25 08-37

lib/store/index.js
```diff
@@ -70,6 +72,27 @@ class Store {
...
   }
+
+  @action setBusySignalRegistry(registry) {
+    this.busySignalRegistry = registry;
+    this.runningKernels.forEach(this._registerBusySignal);
+  }
+
+  _registerBusySignal(kernel: Kernel) {
+    if (this.busySignalRegistry) {
+      let busySignalProvider = this.busySignalRegistry.create();
```
n-riesco: @lgeiger I'd expect eslint to catch that this variable declaration should be a `const`. Would it be OK to update .eslintrc to report this?

lib/store/index.js
```diff
@@ -70,6 +72,27 @@ class Store {
...
+      reaction(
+        () => kernel.executionState,
+        state => {
+          busySignalProvider.clear();
+          if (state != "idle") {
+            let message = kernel.displayName + " is " + state;
```
n-riesco: `const`

lgeiger:
@cbarrick Wow thanks for the cool PR. I this would make the status bar a lot cleaner.

Though I agree with @n-riesco that this removes the possibility of quickly seeing which kernel is running and removes access to the kernel kernel commands available by clicking on the status bar:
busy

In my opinion both are crucial features, especially when having multiple kernels for the same language.

At the moment I can't come up with a different UI that combines the best of the two approaches. Any ideas?

lib/store/index.js
lgeiger: > Would it be OK to update .eslintrc to report this?

Absolutely. Though we need to make sure we don't interfere with prettier.

cbarrick:
What about a pop-up when a kernel is started or connected that provides the kernel info? The kernel commands are always available from the command pallet, and can easily be bound to keyboard shortcuts. Personally, I'd rather not have another button in the status bar.

nikitakit:
@cbarrick I don't think pop-up notifications are quite the right thing. When I start a kernel or connect to one I obviously know which kernel is being used. It's when I switch between tabs or windows that I may be unsure of what kernel is connected or the state it's in (but providing constant pop-ups on switching tabs would be very annoying).

Now I'm somewhat of a "power user" of hydrogen, but here's what my current usage is like. I have 2 atom windows with a dozen tabs each, involving different programming languages. Each of these atom windows may or may not be using a kernel. I have two kernel types that I'm actively using for my current project (one in a local VM and one on a remote computer); this is in addition to more kernels that correspond to other projects I'm not actively working on. The status bar text really helps keep track of all of this.

I'm wondering if it really makes sense to cut the status bar indicator, especially since it's hidden when you're not using Hydrogen. How cluttered is your status bar getting compared to the window width? If it's cluttered perhaps a better alternative would be to restyle the indicator to be narrower. Now I agree that displaying kernel status in text like it's currently done is a bit odd. It may be possible to restyle just that part by moving it to busy-signal, a custom hydrogen icon, or style/color change of the kernel name.

cbarrick:
Ok, I now see how it can be really useful. I'll drop the commit removing the old status bar controls. Perhaps remove the ` | busy` bit instead. Hiding it all together is easy enough from a custom stylesheet should anyone want that.

For the history clutter, it appears that the history list has one entry per busy-signal provider. The alternative implementation would be for all kernels to share a provider. This makes cleaning up the current list less elegant, but not too difficult.

lgeiger:
@cbarrick Thanks a lot for updating this.

One thing I noticed: It looks like the spinner is a lot longer active than it should be. Though I don't know if this is a issue with ` busy-signal`.

This is the current state of this PR:
busy

I'm basically 👍 👍 👎 on this. I really like the idea of  having `busy-signal` as a new status indicator, though it becomes harder to figure out if a kernel is actually busy or not, especially if multiple providers (eg linters) are running and consuming `busy-signal` as well.

@n-riesco @nikitakit @rgbkrk @BenRussert What do you think, should go with this new UI?

lib/store/index.js
```diff
@@ -70,6 +72,27 @@ class Store {
 
     this.grammar = grammar;
   }
+
+  @action setBusySignalRegistry(registry) {
+    this.busySignalRegistry = registry;
+    this.runningKernels.forEach(this._registerBusySignal);
+  }
+
+  _registerBusySignal(kernel: Kernel) {
```
n-riesco: As I find a bit confusing that the last message in the history is `kernel is busy`, how about something like this?

```js
  _registerBusySignal(kernel: Kernel) {
    if (this.busySignalRegistry) {
      const busySignalProvider = this.busySignalRegistry.create();
      reaction(
        () => kernel.executionState,
        state => {
          const message = kernel.displayName + " is " + state;
          busySignalProvider.add(message);
          if (state === "idle") busySignalProvider.clear();
        }
      );
    }
  }
```

n-riesco:
@lgeiger I'm going to sit this one on the fence. I don't feel strongly about this UI change (as long as we also keep Hydrogen's status bar). In general, I find constant (or very frequent) animations in UI disturbing, but I don't use Atom so often to have an opinion about `busy-signal`.

rgbkrk:
I'll have to start using this to find out. If we went ahead and merged it, we could find out if users like it (or if we get a lot of support tickets) and we can always go back to what we had before.

JGH1000:
Regardless of choice made here, could I suggest a minor tweak - having two 'Python 3' in the status bar, or 'Python 3 is busy' message is perhaps not too clear.

I think using official kernel names ('IPyKernel 3', 'IJavascript', 'jp-babel' as in https://github.com/jupyter/jupyter/wiki/Jupyter-kernels) would be clearer and make more sense. As long as everyone is happy with that and it would not mess up anything.

rgbkrk:
This PR has sat long enough that I think it's time to acknowledge that there isn't traction for bringing this in. I'm super appreciative that you took the time to make this PR @cbarrick and I hope to see more contributions from you. I'm going to go ahead and close this PR. If people want to revisit it, feel free to open or continue on discussion.

PR #749, opened 2017-04-26
Refactor kernel-picker to remove atom-space-pen-views

BenRussert:
This pr is heavily influenced by [this example.](https://github.com/atom/grammar-selector/blob/master/lib/grammar-list-view.js)

I haven't found any bugs yet, but this probably needs a little testing.

### DIFF
--- a/lib/kernel-picker.js
+++ b/lib/kernel-picker.js
@@ -5,7 +5,7 @@ import _ from "lodash";
 
 import { log } from "./utils";
 
-export default class SignalListView {
+export default class KernelPicker {
   constructor(kernelSpecs) {
     this.kernelSpecs = kernelSpecs;
     this.onConfirmed = null;
@@ -25,7 +25,7 @@ export default class SignalListView {
         this.cancel();
       },
       didCancelSelection: () => this.cancel(),
-      emptyMessage: "No running kernels found."
+      emptyMessage: "No kernels found"
     });
   }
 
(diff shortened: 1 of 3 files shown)


lib/kernel-picker.js
```diff
@@ -1,70 +1,64 @@
...
-import { SelectListView } from "atom-space-pen-views";
+import SelectListView from "atom-select-list";
 import _ from "lodash";
 
 import { log } from "./utils";
 
-// View to display a list of grammars to apply to the current editor.
-export default class SignalListView extends SelectListView {
-  initialize(kernelSpecs) {
+export default class SignalListView {
```
BenRussert: I guess we could call this class `KernelPicker` right? Left as is to minimize diff for now.

lib/kernel-picker.js
```diff
@@ -1,70 +1,64 @@
...
+        element.textContent = item.display_name;
+        return element;
+      },
+      didConfirmSelection: item => {
+        log("Selected kernel:", item);
+        if (this.onConfirmed) this.onConfirmed(item);
+        this.cancel();
+      },
+      didCancelSelection: () => this.cancel(),
+      emptyMessage: "No running kernels found."
```
n-riesco: I've just noticed that this message should've been `"No kernels found"`.

n-riesco:
I've tested it with multiple kernels, multi-language files. And both, the functionality to start a kernel and the status bar switcher work for me.

lib/kernel-picker.js
lgeiger: 👍 `KernelPicker` is better

lib/kernel-picker.js
lgeiger: Good catch!

PR #751, opened 2017-04-27
Add flow to kernel-picker.js

lgeiger:


lib/kernel-picker.js
```diff
@@ -1,12 +1,17 @@
...
 import { log } from "./utils";
 
 export default class KernelPicker {
-  constructor(kernelSpecs) {
+  kernelSpecs: Array<Kernelspec>;
+  onConfirmed: ?(kernelSpecs: Kernelspec) => void;
+  selectListView: SelectListView;
+  panel: ?atom$Panel;
+  previouslyFocusedElement: ?HTMLElement;
+  constructor(kernelSpecs: Array<Kernelspec>) {
```
rgbkrk: Silly diff.

lib/kernel-picker.js
```diff
@@ -39,7 +44,6 @@ export default class KernelPicker {
       this.panel.destroy();
     }
     this.panel = null;
-    this.editor = null;
```
rgbkrk: Is `editor` not used?

rgbkrk: Certainly not in this file it appears.

PR #758, opened 2017-04-29
Remove atom-space-pen-views

BenRussert:
WIP to remove space-pen-views from signal-list-view. I'm having a flow issue I'm not sure how to resolve, ill comment on the line.


lib/signal-list-view.js
```diff
@@ -1,132 +1,126 @@
...
 
-// View to display a list of grammars to apply to the current editor.
-export default class SignalListView extends SelectListView {
-  initialize() {
-    super.initialize(...arguments);
-
+export default class SignalListView {
+  basicCommands: Array<Object>;
+  editor: ?atom$TextEditor;
+  // onConfirmed: ?(KernelCommand) => void;
```
BenRussert: This is the line I can't seem to get right.

lgeiger: Looking at the error message you sent me via slack the syntax should be
```js
onConfirmed: ?(command: KernelCommand) => void;
```
Similar to [this line](https://github.com/nteract/hydrogen/blob/master/lib/kernel-picker.js#L10)

types/hydrogen.js.flow
```diff
@@ -10,3 +10,8 @@ export type Message = {
   header: Object,
   content: Object,
 }
+
+export type KernelCommand = {
```
lgeiger: Since the `KernelCommand` type is likely not used in a lot of files, I think the clearest thing is to declare it in `signal-list-view.js.js`.

rgbkrk:
It passes! Ready to :ship:?

lib/signal-list-view.js
```diff
@@ -1,132 +1,131 @@
...
     }
-  }
 
-  attach() {
-    // get language from editor
-    this.storeFocusedElement();
-    if (!this.panel) {
-      this.panel = atom.workspace.addModalPanel({ item: this });
+    if (store.editor) {
+      this.editor = store.editor;
```
lgeiger: Is `this.editor` actually used anywhere? We might be able to remove it.

BenRussert: You are right, I removed it with no issues.

lib/signal-list-view.js
```diff
@@ -1,132 +1,124 @@
...
-      this.cancel();
-    } else if (atom.workspace.getActiveTextEditor()) {
-      this.editor = atom.workspace.getActiveTextEditor();
-      this.attach();
-    }
+  _mapCommands(commands: Array<Object>, kernelSpec: Kernelspec) {
+    return commands.map(command => ({
+      name: `${command.name} ${kernelSpec.display_name} kernel`,
+      command: command.value,
+      payload: command.hasOwnProperty("payload") ? kernelSpec : null
```
lgeiger: I think we can be a bit more clear here if we do
```js
payload: (command.value === "switch-kernel") ? kernelSpec : null
```
and remove the payload property from [this line](https://github.com/nteract/hydrogen/pull/758/files#diff-070a3f52b3f9f0cbce35dbf6a46558acR38).

lib/signal-list-view.js
```diff
@@ -1,132 +1,124 @@
...
+      );
+    else {
+      const otherKernels = await kernelManager.getAllKernelSpecsForGrammar(
+        grammar,
+        kernelSpecs => _.without(kernelSpecs, kernel.kernelSpec)
+      );
+
+      listItems = _.union(
+        this._mapCommands(this.basicCommands, kernel.kernelSpec),
+        _.flatMap(otherKernels, otherKernel =>
```
lgeiger: Is there a reason behind not using `.map`? Just curious.

lib/signal-list-view.js
```diff
@@ -1,132 +1,124 @@
...
-      this.setItems([]);
+    let listItems = [];
+    if (!kernel) return;
+
+    if (kernel instanceof WSKernel)
+      listItems = this._mapCommands(
+        _.union(this.basicCommands, this.wsKernelCommands),
+        kernel.kernelSpec
+      );
+    else {
```
lgeiger: Oh crap, looks like prettier messed up here. There are missing brackets:
```js
if (kernel instanceof WSKernel) {
  listItems = this._mapCommands(
    _.union(this.basicCommands, this.wsKernelCommands),
    kernel.kernelSpec
  );
} else {
```

lib/signal-list-view.js
BenRussert: I used `flatMap` because otherwise it was returning nested lists.

I'm not opposed to any other way of doing this if you have something in mind, `flatMap` was just the simplest for me at the time.

lgeiger: Oh sure, thanks for the clarification.

lib/signal-list-view.js
BenRussert: Agreed, that's much better.

PR #764, opened 2017-05-05
Remove spacepen wskernelpicker

BenRussert:
Tried to keep it simple and make as few changes as possible for now.


lib/ws-kernel-picker.js
```diff
@@ -91,7 +97,9 @@ export default class WSKernelPicker {
           this.onSession.bind(this)
         );
         sessionListing.previouslyFocusedElement = this.previouslyFocusedElement;
-        sessionListing.setLoading("Loading sessions...");
+        sessionListing.selectListView.update({
+          loadingMessage: "Loading sessions..."
+        });
```
BenRussert: There is a minor bug in master and in this PR where this loading message remains even after the update below. If I alt-tab away and back the updated list renders. 

Anyone know what the cause is? I'll keep working on how to fix.

n-riesco: First, I'd set a breakpoint [here](https://github.com/BenRussert/hydrogen/blob/66fa4244b86d07c32c399e7cce24a430651c3bff/lib/ws-kernel-picker.js#L129) to make sure  the loading message is removed when you  think it should.

If, despite calling `update`, the loading message isn't cleared, then I'd try to call `sessionListing.selectListView.render()` explicitly.

lgeiger:
I just gave it a quick spin and noticed some focusing problems possibly related to https://github.com/atom/atom-select-list/issues/6.
This makes it hard to cancel the selection:
focus

I'm wondering if we should just keep the panel open and display a loading message instead of closing and refocusing the editor. @BenRussert @nikitakit What do you think?

lib/ws-kernel-picker.js
```diff
@@ -9,45 +9,49 @@ import Config from "./config";
...
-  initialize(emptyMessage, onConfirmed) {
+class CustomListView {
+  constructor(emptyMessage, onConfirmed) {
     this.emptyMessage = emptyMessage;
     this.onConfirmed = onConfirmed;
-    super.initialize(...arguments);
-    this.storeFocusedElement();
+    this.previouslyFocusedElement = document.activeElement;
+    this.selectListView = new SelectListView({
+      itemsClassList: ["mark-active"],
```
lgeiger: Does `mark-active` actually add any useful styling for us? I've seen it in several places but I'm not sure what it actually does 😄.

package.json
```diff
@@ -54,7 +54,6 @@
     "@nteract/transform-plotly": "1.2.0",
     "@nteract/transforms": "^1.0.7",
     "atom-select-list": "^0.2.0",
-    "atom-space-pen-views": "^2.0.5",
```
lgeiger: 🎉 👍 🚀 📦

lib/ws-kernel-picker.js
BenRussert: You might be right. I'll have to test without it later. My assumption was it had something to do with highlighting the selected item

lib/ws-kernel-picker.js
BenRussert: I'll do some more experimenting later. I tried wrapping calling `render` and wrapping these lines in their own `then` call after the current one. Didn't try using the debug tools though, that's a good idea too.

nikitakit:
@lgeiger I think it would be fine to keep the panel open and just swap out the elements being displayed.

To be honest I'm completely confused by how you're supposed to do UI in atom. It's unclear to me what the situation with space pen is -- it works fine, but supposedly it's deprecated?

lib/ws-kernel-picker.js
```diff
@@ -73,11 +77,17 @@ export default class WSKernelPicker {
...
+      items: gateways,
+      infoMessage: "Select a gateway"
+    });
   }
 
   onGateway(gatewayInfo) {
+    const sessionListing = new CustomListView(
+      "No sessions available",
+      this.onSession.bind(this)
+    );
```
BenRussert: Moving this up improves the rendering issue slightly. Instead of needing to switch tabs to refresh the screen the focus goes back to the editor and then moving the cursor refreshes the panel.

Not a fix really, but closer and think it might give us some ideas.

lib/ws-kernel-picker.js
BenRussert: @n-riesco how do I set a breakpoint within a package?

BenRussert:
I get different behavior than @lgeiger I think. My issue is the list doesnt populate until some event like switching tabs in this example below (doesnt matter how long I wait). This is the same for me on master with space-pen-views.
doesnt-load mp4

As of the last commit declaring `const sessionListing` at the top of `onGateway` the focus goes to the editor and if you look closely I can move the cursor up/down which is enough to trigger the refresh instead of switching tabs. Anyone else get this behavior?

loads-when-move mp4

lib/ws-kernel-picker.js
n-riesco: I'd set the breakpoint on [this line](https://github.com/BenRussert/hydrogen/blob/66fa4244b86d07c32c399e7cce24a430651c3bff/lib/ws-kernel-picker.js#L129) and then *step in* (press the button with an arrow pointing down, or F11 in linux).

lgeiger: @BenRussert You can add breakpoints inside the dev tools panel. Take a look at: https://developers.google.com/web/tools/chrome-devtools/javascript/breakpoints

BenRussert: The issue I ran into was navigating the sources tab to find hydrogen or other atom packages, I couldnt seem to find them.

lgeiger: I think the code only shows up once Hydrogen is activated (e.g. once you run some code).

lgeiger:
> It's unclear to me what the situation with space pen is -- it works fine, but supposedly it's deprecated?

@nikitakit [`atom-space-pen-views`](https://github.com/atom/atom-space-pen-views) depends on [`space-pen`](https://github.com/atom-archive/space-pen) which isn't maintained anymore. It was recently [removed from Atom](https://github.com/atom/atom/issues/13254) for performance reasons since it depends on jQuery and won't receive any bug fixes. For us this removes one quite big dependency and reduces our package size.

lgeiger:
> @lgeiger I think it would be fine to keep the panel open and just swap out the elements being displayed.

👍 

@BenRussert I think using the same panel and only updating the content is the easiest way to fix the focusing issue.

lib/ws-kernel-picker.js
BenRussert: Heh, seems to be working fine now. That's probably what my problem was.

BenRussert:
The latest commit fixes the issues I know of. Please test this out if you have a chance as I don't think I've covered all the possible usage paths.

lib/ws-kernel-picker.js
```diff
@@ -73,67 +77,80 @@ export default class WSKernelPicker {
...
+              })
+          );
+        },
+        () => atom.notifications.addError("Connection to gateway failed")
+      )
+      .then(() => {
+        sessionListing.selectListView.focus();
+        sessionListing.selectListView.reset();
+      });
   }
```
BenRussert: Lines 150-153 above are pretty much all the recent commits add, the rest is just prettier formatting.

lib/ws-kernel-picker.js
```diff
@@ -153,11 +170,13 @@ export default class WSKernelPicker {
           options
         };
       });
-      kernelListing.setItems(items);
+      kernelListing.selectListView.update({ items: items });
+      kernelListing.selectListView.focus();
+      kernelListing.selectListView.reset();
```
BenRussert: In my other recent PR's the picker classes I used had `toggle` and `attach` methods that would do update/focus/resetting for us. I didn't add these methods this time to keep the class design similar to how it worked before, with most of the important stuff right in the constructor.

Issue #523, opened 2016-11-30
Include current project path in kernel's path?


MaximilianR:
If I'm working on a project, ideally I would want `import myproject` to reference the project's files (rather than any installed version of `myproject`).

If I start `ipython` from my project directory, this works. I believe that's because `ipython` maps my current path and so resolves the package locally.

But this doesn't seem to with `ipykernel`. Is this something people have encountered? 

Alternatively let me know if this is an issue with ipykernel rather than hydrogen.

Thanks!

lgeiger:
Hydrogen starts the Kernel in the directory of the current file. So as long as you start the kernel in the root directory of your project this should work.

You can check in which directory the kernel was started by executing `!pwd` in Hydrogen or IPython.

buyology:
> start the kernel in the root directory of your project

how would you do this?

 `Script`s `cmd`-`i` runs in the project root while `Hydrogen` runs in the directory of the actual file

buyology:
@slavaGanzin sounds good, guess I'll have to patch it myself in the meantime

buyology:
actually @slavaGanzin wouldn't it be relevant to submit the [`project path`](https://github.com/nteract/hydrogen/blob/a19dfd54b566ac02290c8aefd1dd3c169bfec67b/lib/kernel-manager.js#L128-L129)-patch in a separate PR?

I cannot see why this shouldn't be the default behaviour especially with lots of ppl coming from `Script` and using this with the `python` kernel

lgeiger:
> start the kernel in the root directory of your project
@lgeiger how would you do this?

You can do this by starting the kernel in any python file that is in the root directory of your project.

@buyology Please feel free to submit a PR changing this behavior. That would be cool 🎉 

I personally like to start the kernel in the current file but I see that this becomes tedious when working in a bigger project. It's probably best to change that.

PR #771, opened 2017-05-11
Make kernel start in project dir root default and add setting

buyology:
xref [#523](https://github.com/nteract/hydrogen/issues/523#issuecomment-300893935)

* Makes project directory root the default kernel start directory
* Adds setting (`startInProjectRootDir`) to control this behaviour with the possibility to revert to previous behaviour (start kernel in same dir as file current file)

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -42,7 +42,10 @@ const Config = {
           value: "projectDirOfFile",
           description: "The project directory relative to the file"
         },
-        { value: "dirOfFile", description: "Current directory of the file" }
+        {
+          value: "dirOfFile",
+          description: "Current directory of the file"
+        }
       ],
       default: "firstProjectDir"
     },
--- a/lib/config.js
+++ b/lib/config.js
@@ -28,12 +28,23 @@ const Config = {
       type: "boolean",
       default: false
     },
-    startInProjectRootDir: {
-      title: "Start kernel in project root directory",
+    startDir: {
+      title: "Directory to start kernel in",
       includeTitle: false,
-      description: "If enabled, the kernel is started in the project's root directory (default). If disabled, it is started in the current file's directory. Restart Atom for changes to take effect.",
-      type: "boolean",
-      default: true
+      description: "Restart the kernel for changes to take effect.",
+      type: "string",
+      enum: [
+        {
+          value: "firstProjectDir",
+          description: "The first started project's directory (default)"
+        },
+        {
+          value: "projectDirOfFile",
+          description: "The project directory relative to the file"
+        },
+        { value: "dirOfFile", description: "Current directory of the file" }
+      ],
+      default: "firstProjectDir"
     },
     kernelNotifications: {
       title: "Enable Kernel Notifications",
--- a/lib/config.js
+++ b/lib/config.js
@@ -28,6 +28,13 @@ const Config = {
       type: "boolean",
       default: false
     },
+    startInProjectRootDir: {
+      title: "Start kernel in project root directory",
+      includeTitle: false,
+      description: "If enabled, the kernel is started in the project's root directory (default). If disabled, it is started in the current file's directory. Restart Atom for changes to take effect.",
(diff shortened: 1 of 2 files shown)


lib/kernel-manager.js
```diff
@@ -86,11 +86,18 @@ class KernelManager {
 
   startKernel(kernelSpec, grammar, onStarted) {
     const displayName = kernelSpec.display_name;
+    let kernelStartDir;
 
     log("KernelManager: startKernel:", displayName);
 
+    if (atom.config.get("Hydrogen.startInProjectRootDir")) {
+      kernelStartDir = atom.project.getPaths()[0];
```
lgeiger: I think `atom.project.getPaths()` can be empty if no project is present in the tree view. In that case we should fall back to the `getEditorDirectory(store.editor)`.

A very nice touch (now or in a future PR) would be to use the project where the current file belongs to and not just the first one. This could be handy if multiple projects are open.

lib/config.js
```diff
@@ -28,6 +28,13 @@ const Config = {
       type: "boolean",
       default: false
     },
+    startInProjectRootDir: {
+      title: "Start kernel in project root directory",
+      includeTitle: false,
+      description: "If enabled, the kernel is started in the project's root directory (default). If disabled, it is started in the current file's directory. Restart Atom for changes to take effect.",
```
BenRussert: > Restart Atom for changes to take effect.

I think you could just restart the kernel, right?

buyology:
* Added check for length of `atom.project.getPaths()`-array
* Added support for multiprojects set-up (didn't know you could do this!)
* Refactored the settings into an `enum` with these three settings:
  * `firstProjectDir`
  * `projectDirOfFile`
  * `dirOfFile`

lib/kernel-manager.js
```diff
@@ -86,11 +86,26 @@ class KernelManager {
...
   startKernel(kernelSpec, grammar, onStarted) {
     const displayName = kernelSpec.display_name;
+    let kernelStartDir;
+    let currentPath = getEditorDirectory(store.editor);
 
     log("KernelManager: startKernel:", displayName);
 
+    switch (atom.config.get("Hydrogen.startDir")) {
+      case "firstProjectDir":
+        if (atom.project.getPaths().length > 0) {
```
lgeiger: If this condition isn't fulfilled it falls back to `case "projectDirOfFile":` which will return `/`. I think it should fall back to `currentPath` instead.

lib/kernel-manager.js
```diff
@@ -86,11 +86,26 @@ class KernelManager {
...
+          kernelStartDir = atom.project.getPaths()[0];
+          break;
+        }
+      case "projectDirOfFile":
+        kernelStartDir = atom.project.relativizePath(currentPath)[0];
```
lgeiger: Using `atom.project.relativizePath` can be a bit problematic since it will return `/` if no project is found or if there isn't a project related to the current file. I think it should fall back to `currentPath` instead of `/`. What do you think?

lib/kernel-manager.js
buyology: Agreed!

lib/kernel-manager.js
buyology: Was just going to write and say we need to catch this :) According to the docs it should return `null` if there is no projectpath? Need to test this. 

Is the way to test "without" a project path to open a file directly in atom?

lgeiger: > Is the way to test "without" a project path to open a file directly in atom?

You can remove the project directory from the tree view via the context menu.

buyology:
updated

lgeiger:
Thank you for the work! 👍 

@buyology Would you like to help us maintain this package?  😄

lgeiger:
On thing that would be super cool is to move this logic to a function in `utils.js` and add a few tests, so we can reuse it in the future.

PR #772, opened 2017-05-12
Rewrite result view as React component

### LINKED ISSUE #46: add separate containers and background coloration for different streams
At present, `stdout` and execution results both look the same.

It would be nice to tint the background differently for `stdout`, errors, and returns.

### LINKED ISSUE #252: [DISCUSSION] Hydrogen CSS namespace

### LINKED ISSUE #466: Support carriage return symbol
Hydrogen doesn't respect `\r` symbol

`print('ab\rcd')` will produce `cd` in console and/or jupyter notebook. In hydrogen it will show up as

```
ab
cd
```

This is especially critical when working on scripts that report progress, e.g. use `tqdm`

Atom version: 1.10.2
Hydrogen version: 0.14.0

### LINKED ISSUE #632: Images are higher than bubble output
**Description:**

When the output is a plot, the plot can be taller than the bubble container:

Observed behavior:

screenshot from 2017-03-11 15-01-42

Expected behavior:

screenshot from 2017-03-11 15-01-47

I realize that the "expected" behavior is eating up a lot of the screen space. Is this a feature, or should the figure be scaled down, or anything else?

**Versions:**
Atom 1.114.4 / Hydrogen 1.6.0

### LINKED ISSUE #744: Output Message Folding
When I print the output of a single ling command the message is often hidden under minimap with no way of actually scrolling to see it. Not a big deal but I constantly have to divide the line so I can see the output (a short example on the screenshot, I realize I could format the float but it happens with other outputs as well).

Anyway to fix it so it detects minimap and prints the output underneath? Many thanks for the fantastic package. 

screenshot from 2017-04-24 16-59-29

lgeiger:
## Changes
This PR aims at fixing a few issues with our current result display:
- [x] Correctly determine if outputs fits inline: Fixes #744
fit
- [x] Support carriage return symbol: Fixes #466
progress
- [x] Highlight `stderr` output: Fixes #46 Requires https://github.com/nteract/nteract/pull/1694 and https://github.com/nteract/ansi-to-react/pull/25 bildschirmfoto 2017-05-14 um 23 37 57
- [x] Cleanup CSS: Fixes #252
- [x] Add button to expand result: Fixes #632

It was easier to completely redo our CSS.
The current design works well with the old default Atom theme but doesn't really shine with other ones.
The aim was to have a cleaner output display in order to work better with other Atom themes. I'm not a great web designer but this is what I came up with. Please provide feedback on what you think and feel free to mess with it:
bildschirmfoto 2017-05-12 um 12 33 55

## TODO
- [x] Fix watches
- [x] Ship https://github.com/nteract/nteract/pull/1690 to fix flow errors
- [x] Fix watches CSS
- [x] Refine multiline CSS
- [x] Reimplement scroll
- [x] Test with other kernels

Side note: For Hydrogen (because we use MobX) it would be cool to remove the usage immutable in `@nteract/display-area` and `@nteract/transforms`. This would also enable us to better type the packages.


n-riesco:
Can't look into this now, but I'm concerned this PR will make Hydrogen incompatible with kernels speaking v4 of the messaging protocol.

rgbkrk:
We should probably write a v4 to v5 message spec converter like is done in the Jupyter clients, and make it part of the nteract monorepo. The display area is merely a notebook format style data in, display out -- no messages are supposed to go to it.

lgeiger:
> We should probably write a v4 to v5 message spec converter like is done in the Jupyter clients, and make it part of the nteract monorepo.

Good idea! Is there an implementation from jupyter we could use as a reference for such a converter?

styles/hydrogen.less
```diff
@@ -5,487 +5,166 @@
...
+}
 
-    .spinner .rect4 {
-        -webkit-animation-delay: -0.9s;
-        animation-delay: -0.9s;
-        margin-left: 1px;
-    }
+.spinner .rect5 {
+  animation-delay: -0.8s;
+}
```
BenRussert: You could nest selectors in less (unless we have a reason not to). For example all the spinner styling could be done with only one :spider:

```less
.spinner {
    padding: 0 0.5em 0 0.5em;

    div {
      background-color: @hydrogen-inline-color;
      height: 100%;
      width: 2px;
      display: inline-block;
      margin-left: 1px;
      animation: stretchdelay 1.2s infinite ease-in-out;
    }

    .rect1 {
      margin-left: 0px;
    }

    .rect2 {
      animation-delay: -1.1s;
    }

    .rect3 {
      animation-delay: -1.0s;
    }

    .rect4 {
      animation-delay: -0.9s;
    }

    .rect5 {
      animation-delay: -0.8s;
    }
}

```

rgbkrk:
> Good idea! Is there an implementation from jupyter we could use as a reference for such a converter?

The adaptation is done in Python here: https://github.com/jupyter/jupyter_client/blob/fd05171a94bd6431df0b48f280f9027934b58d98/jupyter_client/adapter.py

However, there's a bit of code for handling the adapted messages on the frontend:

* https://github.com/jupyter/notebook/blob/666ecbf35c3310335a3c8c182e8f53441c991c14/notebook/static/notebook/js/completer.js#L180-L190

* https://github.com/jupyter/notebook/blob/6cebd3ef666d8ab6a491343e91063adc96ff828f/notebook/static/notebook/js/outputarea.js#L933-L942

rgbkrk:
> Side note: For Hydrogen (because we use MobX) it would be cool to remove the usage immutable in @nteract/display-area and @nteract/transforms. This would also enable us to better type the packages.

A+ for me

BenRussert:
Nice work!

I am having good results testing this on python and julia kernels on both ubuntu and windows. I will keep testing, as this is a big, but valuable change to the repo. I am in favor of merging as soon as any open questions are answered. I'd like to know what everyone else thinks!

### My open questions:

* @n-riesco or @rgbkrk : how can we test if there is a problem for v4 protocol? I'd like to help but pretty green here.

* What are all of your thoughts on the base style we ship with? Do we try to mimic previous releases or go with the new look in @lgeiger's branch

* Are there any other concerns to be addressed before we can merge?

In #252 I suggested I working on a styles readme to guide devs or users looking to customize (people tend to do this with atom :smile: ). In this case we would go minimal and provide the tools to go from there as desired.  I have some opinions I'll post in review comments.

Here is the look for me:
one-dark
one-dark

one-light:
one-light

BenRussert:
One weird bug that seems just to be me. When I run json output w/ IPython.display I get a mismatched theme's styling. I reinstalled atom today and cleared out my .atom dir just to be sure, but still have the issue.
mismatched-theme

styles/hydrogen.less
```diff
@@ -5,488 +5,275 @@
...
 
-            color: lighten(@syntax-text-color, 10%);
-            background-color: @bubble-background;
+.multiline-container .toolbar .icon::before {
+  margin: 0;
+  width: 2.5em;
+  line-height: 2.5em;
+  height: 2.5em;
+  cursor: pointer;
+}
```
BenRussert: My preference would be to add back `text-alight: center;` to all the toolbar icons. It may just be because I'm used to it that way.

styles/hydrogen.less
```diff
@@ -5,488 +5,275 @@
...
 
-            margin: 0;
-            padding: 0;
-        }
-    }
+.multiline-container .toolbar .icon:hover {
+  color: @hydrogen-inline-color;
```
BenRussert: I like to have icons highlight when hovering personally.  I think toolbar icons should include `background-color: @background-color-highlight;` (We may not even need the `color` property line above) 

Here is how that looks (icons centered as well in this example)
background-icon

n-riesco:
> @n-riesco or @rgbkrk : how can we test if there is a problem for v4 protocol? I'd like to help but pretty green here.

IJavascript provides a flag for that:

```js
      {
          "kernelspecs": {
              "ijavascript": {
                  "spec": {
                      "argv": [
                          "ijskernel",
                          "--hide-undefined",
                          "--protocol=4.1",
                          "{connection_file}"
                      ],
                      "language": "javascript"
                  }
              }
        }
    }
```

But we should really test against kernels, the current code contains workarounds that @willwhitney developed for some kernels.

@willwhitney what kernels did you tested Hydrogen against?

styles/hydrogen.less
lgeiger: 👍 I like that!

styles/hydrogen.less
lgeiger: Good catch, the `file-icons` package added `text-align: center;` globally so it looked fine for me.

lgeiger:
Thanks for the great feedback!
I added a very basic converter to keep this PR backwards compatible with kernels speaking jmp v4. It doesn't add full v4 support but handles the cases we currently do except for the workarounds for kernels that do not conform to the messaging standard.

lib/watch-view.js
```diff
@@ -18,11 +20,7 @@ export default class WatchView {
     this.inputEditor.setLineNumberGutterVisible(false);
     this.inputEditor.moveToTop();
 
-    this.resultView = new ResultView();
```
BenRussert: We should change the `clearResults` method below since `this.resultView` can be null. (We should flow this up, but it can wait until the PR is merged)

This is the current code:

```js
  clearResults() {
    try {
      this.element.removeChild(this.resultView.element);
      this.resultView.destroy();
    } catch (e) {
      console.error(e);
    }

    this.resultView = new ResultView({ props });
    this.element.appendChild(this.resultView.element);
  }
```

This works better:

```js
clearResults() {
    if (this.resultView != null) {
      this.element.removeChild(this.resultView.element);
      this.resultView.destroy();
    }

    this.resultView = new ResultView({ props });
    this.element.appendChild(this.resultView.element);
  }
```

lib/components/result-view/result-view.js
```diff
@@ -0,0 +1,191 @@
...
+  };
+
+  render() {
+    const { outputs, status, destroy, showStatus, multiline } = this.props;
+    const { position: { lineLength, lineHeight, editorWidth } } = this.props;
+
+    const inlineStyle = {
+      "margin-left": `${lineLength + 1}ch`,
+      "margin-top": `-${lineHeight}px`
+    };
```
lgeiger: Great catch!

lib/watch-view.js
lgeiger: That's a lot better!

willwhitney:
I tested basic things against Python, Julia, R, and Lua. Several others
were confirmed to work well also and may even have had small fixes. I'd
have to go look at the commit history and old versions of the readme to
know more.

lib/components/result-view/index.js
```diff
@@ -0,0 +1,107 @@
...
+ * https://github.com/nteract/hydrogen/issues/466#issuecomment-274822937
+ * An output can be a stream of data that does not arrive at a single time. This
+ * function handles the different types of outputs and accumulates the data
+ * into a reduced output.
+ *
+ * @param {Array<Object>} outputs - Kernel output messages
+ * @param {Object} output - Outputted to be reduced into list of outputs
+ * @return {Array<Object>} updated-outputs - Outputs + Output
+ */
+export function reduceOutputs(
```
rgbkrk: Should we go ahead and get this reducer exported?

We could start exporting the reducers as individual functions if that seems easy enough to work with outside of nteract desktop.

lib/components/result-view/index.js
```diff
@@ -0,0 +1,107 @@
...
+): IObservableArray<any> {
+  const last = outputs.length - 1;
+  if (
+    outputs.length > 0 &&
+    output.output_type === "stream" &&
+    outputs[last].get("output_type") === "stream"
+  ) {
+    const stream: StreamOutput = output;
+    outputs[last] = outputs[last].update("text", text => text + stream.text);
+    return outputs;
```
rgbkrk: Since we're in the midst of this -- this is exactly the segment I wanted to handle carriage returns and other terminal movement commands in -- then it's only done once and not on every render.

lib/components/result-view/index.js
lgeiger: For this particular reducer we cannot use the one from nteract since it uses a immutable list which probably won't work well with MobX.

That said I really like the idea of exporting more nteract internals.

lib/components/result-view/index.js
lgeiger: I was thinking about this too. I have messed with `escape-carriage` to make a safe version that allows escaping of intermediate output. I'll push that tomorrow.

lgeiger: Checkout https://github.com/lgeiger/escape-carriage/pull/5

lgeiger:
I've used this branch with Python and IJavascript (jmp v5 and jmpv4) and haven't found any problems so far. And looked at the other Kernels mentioned in the repo to check that they comply to the messaging spec.
I think we're good to go.

I've some ideas on how to simplify this further so we're able to support https://github.com/nteract/hydrogen/issues/760 but I think we should do this in future PRs to keep the diff somewhat manageable.

rgbkrk:
Away we go!

JGH1000:
Looks great @lgeiger !

PR #773, opened 2017-05-12
chore(kernelStartDir): Simplify variable assignment

lgeiger:
Minor thing I noticed after merging #771 

/cc @buyology

### DIFF
--- a/lib/kernel-manager.js
+++ b/lib/kernel-manager.js
@@ -86,7 +86,6 @@ class KernelManager {
 
   startKernel(kernelSpec, grammar, onStarted) {
     const displayName = kernelSpec.display_name;
-    let kernelStartDir;
     let currentPath = getEditorDirectory(store.editor);
     let projectPath;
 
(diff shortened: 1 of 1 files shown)


lib/kernel-manager.js
```diff
@@ -101,11 +100,7 @@ class KernelManager {
         break;
     }
 
-    if (projectPath != null) {
-      kernelStartDir = projectPath;
-    } else {
-      kernelStartDir = currentPath;
-    }
+    const kernelStartDir = projectPath != null ? projectPath : currentPath;
```
n-riesco: what if projectPath is `'null'`?

}:>

lgeiger: We could also use

```js
const kernelStartDir = projectPath || currentPath;
```
Even shorter :wink:

buyology: :) cool! unlearning this kind of shortening stuff w/ `go` atm but looks reasonable :)

@n-riesco this is not a string but the `null`/"undefined" keyword?

n-riesco: @buyology you're right (somehow in mind I thought JS would convert the operands to the same type before evaluating the comparison), but I just checked and:
```
$ node
> "null" != null 
true
```

buyology: whoa it's kinda crazy you can do `''+null` and end up with `'null'`. but I guess the above is kind of exploiting javascript's `null`-handling a bit in the first place.

lgeiger: 👀  crazy javascript null handling.

Anyway, I think we should be fine since `projectPath` is either `null`, `undefined` or a proper path.

PR #784, opened 2017-05-17
Fix CI: Switch to trusty dist for to fix Atom 1.18 builds

lgeiger:
This also adds `ATOM_CHANNEL=stable` to our build matrix.


lgeiger:
Looks like our builds now take a bit longer

rgbkrk:
😨

PR #790, opened 2017-05-19
Add flow to plugin-api

BenRussert:


BenRussert:
I've been messing around with some plugins lately and thought i'd add flow to our plugin api. There were just a couple changes to handle null cases etc.

lib/plugin-api/hydrogen-kernel.js
```diff
@@ -37,7 +42,10 @@ export default class HydrogenKernel {
    */
   getConnectionFile() {
     this._assertNotDestroyed();
-    const { connectionFile } = this._kernel;
+
+    const connectionFile = this._kernel.connectionFile
+      ? this._kernel.connectionFile
+      : null;
     if (!connectionFile) {
```
BenRussert: Only the ZMQKernel has a connectionFile prop, so importing ZMQKernel and handling this way made flow happy.

PR #792, opened 2017-05-20
Max-height 100% for media in output

buyology:
Currently, when plotting using e.g. the R-kernel images needs to be scrolled to view full output, this adds `max-height: 100%`

before
after

### DIFF
--- a/styles/hydrogen.less
+++ b/styles/hydrogen.less
@@ -35,6 +35,7 @@ img {
 
 img, svg, video, audio {
   display: block;
+  max-height: 100%;
 }
 
 td, th {


lgeiger:
Thanks a lot for the PR!

Though I'm afraid this makes the new "expand" button obsulete 😄 

Fully expanding the output may eat up a lot of the screen space, especially on smaller screen (see #632). We've also thought about scaling down images to fit, but this doesn't allow users to change the image size in their scripts and may lead to strange behaviors (see #600).

I'd personally not use this behavior per default. Should we think about ways to configure this behavior? Experienced users can always override this with a custom style sheet anyway.

BenRussert:
> Experienced users can always override this with a custom style sheet anyway

@lgeiger I'm thinking with a few pointers even beginners will be able to customize their stylesheet for hydrogen. I think it would be awesome to pivot this pr into getting that started.


A section in our gitbook where we start compiling snippets like these and a screenshot of what it does would be so cool. We can hopefully get some more people contributing ideas this way and we could start to see more of what makes atom/hydrogen different.

buyology:
I see, so sorry for not looking more closely at the previous issues. My reasoning was just that I couldn't imagine a situation where you would actually want to scroll upwards and downwards to see the entirety of a plot :nerd_face: (especially since R seems to somehow produce these gigantic plots out of the box relative to matplotlib? haven't looked to deep into that though)

*Maybe* it would make sense to make the height of the output relative to the total viewport height?

Anyway, I will pivot this PR into an addition to the new docs @BenRussert 👍

lgeiger:
Closing this for the reasons outlined in https://github.com/nteract/hydrogen/pull/792#issuecomment-302898000

PR #794, opened 2017-05-20
Add result history component and refactor watch views

lgeiger:
## Changes for users
This PR refactors the watch view as a react component which fixes quite a few bugs related to the history slider when resizing the panel.
Overall the watch panel is now a lot faster (most noticeable when scrolling through the output history).
This also removes the toolbar from the result view. In my opinion the toolbar doesn't add a lot of benefit inside the watch panel and mainly takes away space from the output:

watches

## Changes from a technical perspective
* **`History` component:** Renders one result at a time but keeps a history of all the output. This is now used by `WatchView` and can be used to render output in a separate panel.
* **`OutputStore`:** The result logic extracted from `ResultViewComponent` so it can be used in other components ( e.g.WatchView or a separate output panel).


rgbkrk:
Whoa this is pretty sweet.

lgeiger:
This is the new rendering:

v1.13.0 | PR
--------|--------
| pr

rgbkrk:
Gotta love that DOM diffing. 😄

PR #798, opened 2017-05-21
Add flow to watch-sidebar

BenRussert:


lib/watch-sidebar.js
```diff
@@ -157,9 +170,10 @@ export default class WatchSidebar {
     document.removeEventListener("mouseup", this.resizeStopped);
   }
 
-  resizeSidebar({ pageX, which }) {
+  resizeSidebar({ pageX }: { pageX: number }, { which }: { which: number }) {
```
BenRussert: This one blew my mind a little due to the destructuring. Appreciate if someone spot checks me here.

lgeiger: Flow types and destructuring is a bit tricky.

```js
resizeSidebar({ pageX, which }: { pageX: number, which: number }) {
```
should do the job.

lgeiger: or if it doesn't fit inline:

```js
type foo = { pageX: number, which: number }

resizeSidebar({ pageX, which }: foo) {
```

PR #799, opened 2017-05-21
Add sytle customization docs page

BenRussert:
I think giving people the tools they need to customize is the way to resolve style preference issues. My hope is that people contribute to this page and we get a cookbook of styling ideas. I'm hoping @buyology's idea in  #792 will be added here!

Will add images when I figure out the workflow for getting the links to work on github.

### DIFF
--- a/docs/StyleCustomization.md
+++ b/docs/StyleCustomization.md
@@ -33,6 +33,11 @@ In your styles.less:
 }
 ```
 
+Depending on your active theme, this will look like:
+
+![](https://cloud.githubusercontent.com/assets/10860657/26287949/52c75e9c-3e4b-11e7-929b-b155434927b3.png)
+
+
 
 Change the icon color when you hover over the toolbar icons:
 
@@ -45,6 +50,8 @@ Change the icon color when you hover over the toolbar icons:
   }
 }
 ```
+
+![](https://cloud.githubusercontent.com/assets/10860657/26086644/8dbfd806-39b1-11e7-9149-e4d891d0455b.gif)
 
 ### Importing Hydrogen Style Variables
 
--- a/docs/StyleCustomization.md
+++ b/docs/StyleCustomization.md
@@ -1,0 +1,67 @@
+# Style Customization
+
+By default Hydrogen leans towards plain styling so that its elements can be customized easily to your liking.
+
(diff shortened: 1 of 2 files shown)


docs/StyleCustomization.md
```diff
@@ -0,0 +1,67 @@
...
+If you want to use Hydrogen's style variables you can do so by adding this to your styles.less:
+
+```less
+  @import "./packages/Hydrogen/styles/hydrogen.less";
+```
+
+Once you've done this you will have access to any style variables within the Hydrogen package:
+
+```less
+@hydrogen-light-background
```
lgeiger: Is it also possible for people to change these in their style sheet?

Or redefine them (maybe if the import is omitted)?

BenRussert: I think you could redefine them either way.

I thought about suggesting that you save your hydrogen stuff outside your styles.less then import it to avoid clutter (I do this for various style then comment in/out as needed). But I left out for now for simplicity.

Save a file of stuff in the `~/.atom/hydrogen-custom-styles.less` then in your styles.less:

```less
@import "hydrogen-custom-styles.less";
```

lgeiger: > I thought about suggesting that you save your hydrogen stuff outside your styles.less then import it to avoid clutter (I do this for various style then comment in/out as needed). But I left out for now for simplicity.

👍  for simplicity

I only was wandering if one redefines `@hydrogen-light-background: red`, will Hydrogen recognize that?

BenRussert: > @hydrogen-light-background: red, will Hydrogen recognize that?

Technically not Hydrogen doing the recognizing :smirk:, but yes this works!

You can do crazy stuff with less, but it is a little hard to read and hackish. Some of the nesting and grouping property values is really good, but variables are wierd. 

To do what I do in these examples you can even go:
```less
@packagename: ~'.hydrogen';
@multiline-container: ~'@{packagename} .multiline-container';
@cell_display: ~'@{multiline-container} .cell_display';


@{multiline-container} .toolbar .icon:hover {
  background-color: @background-color-selected;
}

@{cell_display} {
  background-color: @background-color-highlight;
}
```

buyology:
Great!

PR #803, opened 2017-05-22
Docs: Add link to required language-r package

### LINKED ISSUE #791: Running R with hydrogen - examples
Hi everybody,
I am very happy with the possibility to work with python in Atom by using the package hydrogen, especially because I can have the output, **including plots inside the working page (inline results). However, when I try to do the same with R I cannot get the same output experience**. 
I looked extensively for supports and examples but I didn't find anything. 
Appreciate any help. Thanks Jai

lgeiger:
Fixes #791


BenRussert:
This is probably the most common kernel that doesnt have a default language install for atom.

rgbkrk:
Yeah I wish R support was just part of Atom.

PR #810, opened 2017-05-23
Set icon height to 100%

### LINKED ISSUE #808: Checkmark bubble size doesn't scale correctly with font size
I'm using a font size of 12 (rather than the default of 14), with the One Dark theme.

With these settings results bubbles with checkmarks are rendered slightly taller than one line, causing the line spacing to change when I run code.

lgeiger:
Fixes #808 

Default icon height is 16px which causes problems with smaller font sizes.


lgeiger:
:shipit:

Issue #795, opened 2017-05-20
Use Atoms Docks API for our panels


lgeiger:
Atom 1.17 introduced a new docks API (Used by the [tree view](https://github.com/atom/tree-view), [GitHub](https://github.com/atom/github) and [Linter](https://github.com/steelbrain/linter-ui-default) packages).
We should use it as well. This will allow us better work in conjunction with other packages and makes our panels more flexible.

- [ ] **Inspector**
We should do this first since it's simpler to implement with less refactorings and we can get rid `react-resizable-box` then.
- [ ] **Watch Sidebar**

**Relevent resources:**
* http://blog.atom.io/2017/05/16/atom-1-17.html
* http://flight-manual.atom.io/hacking-atom/sections/package-active-editor-info/
* https://atom.io/docs/api/v1.17.0/Dock

**Example PRs from other projects:**
* https://github.com/atom/tree-view/pull/1056
* https://github.com/steelbrain/linter-ui-default/pull/256 and https://github.com/steelbrain/linter-ui-default/pull/302
* https://github.com/atom/github/pull/664

BenRussert:
I will start working on this issue. I think we should ship a release before we finish this due to the atom v1.17 dependency. This way holdouts can still get the new result and watch-views!

lgeiger:
Even more information on the Atom blog: http://blog.atom.io/2017/05/23/docks-deep-dive.html

lgeiger:
I refactored the Watch Sidebar as a react component so we can easily use it in a dock.
@BenRussert I'd happy to pair with you in the next few days and move our panels to docks.

PR #811, opened 2017-05-23
Chore: Update atom types for v1.17

BenRussert:
Updated with new 1.17 to support #795, and is back compatible. This is just a paste from nuclide, with our additions as before.

### DIFF
--- a/types/atom.js.flow
+++ b/types/atom.js.flow
@@ -353,6 +353,17 @@ declare class atom$StyleManager {
 
   // Paths
   getUserStyleSheetPath(): string,
+
+  // (Undocumented.)
+  addStyleSheet(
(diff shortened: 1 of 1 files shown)


types/atom.js.flow
```diff
@@ -1817,10 +1878,3 @@ declare class atom$Selection {
     },
   ): string,
 }
-
-declare class atom$LanguageMode {
```
lgeiger: I think we still need `atom$LanguageMode`

lgeiger: Yep, Travis thinks so too 😜

BenRussert: Yeah, missed that sorry.

Issue #817, opened 2017-05-26
Performance issue when inspect-printing nested objects


nils-werner:
**Preface:**

When I say 'inspect-print' I mean Hydrogen's behaviour to print a value if the return statement of an expression is not saved as a variable. e.g.

    5

will trigger an 'inspect-print', while

    a = 5

will not.

**Description:**

When inspect-printing a nested structure, like a list of arrays, Hydrogen seems to hit a performance issue.

Note that both outputs show roughly the same output and roughly the same number of characters. Also, when switching back and forth between tabs with long text output, the 'inspect-print' tab seems to react slower than one with a normal `print()` statement.

**Steps to Reproduce:**

Compare

```python
import numpy
data = [numpy.random.rand(128, 128) for _ in range(128)]
print(data)
```

which executes almost instantly, against

```python
import numpy
data = [numpy.random.rand(128, 128) for _ in range(128)]
data
```

which takes a few seconds to complete. This issue scales with growing data structure size until statements may take hours to complete.

Flat objects show a similar issue but scale much better, if you compare

```python
import numpy
print(numpy.random.rand(128, 128, 128))
```

against

```python
import numpy
numpy.random.rand(128, 128, 128)
```

which takes a few (noticeable) hundred milliseconds longer to execute.

**Versions:**

OSX 10.11.6, Atom 1.17.0, Hydrogen 1.14.0

lgeiger:
I wouldn't be surprised if this is caused by https://github.com/nteract/ansi-to-react/issues/28. For more info see https://github.com/nteract/nteract/pull/1710

nils-werner:
I don't think so, as both outputs are roughly the same number of characters. Also I noticed that the tab with the 'inspect-print' statement reacts slower when switching between tabs. Added both to the original issue.

lgeiger:
I can reproduce this in Hydrogen and nteract.
The performance issue is due to the linkification in our [`text/plain` transform](https://github.com/nteract/nteract/blob/master/packages/transforms/src/text.js#L18).

@rgbkrk What about disabling the linkification until we find a faster way?

rgbkrk:
> What about disabling the linkification until we find a faster way?

I completely agree.

PR #818, opened 2017-05-29
Use docks for inspector panel

lgeiger:
This is a initial stab at using docks for our panels as outlined in #795:
docks

I think we should wait with merging until we've shipped a new release (including the fix for #817) since this PR requires Atom 1.17.

@BenRussert Thanks a lot for the help and the awesome feedback 👍


rgbkrk:
Now that the patch release is out, are #818 and #819 good to bring in?

lgeiger:
I used this branch today for work. Haven't found any issues. :+1:  from me

PR #825, opened 2017-05-29
Add External output area

### LINKED ISSUE #760: Feature request: Output in separate pane
It would be great if we could get the output in a separate pane, like in this package for VSCode:

https://github.com/DonJayamanne/vscodeJupyter

lgeiger:
This is a initial implementation of a external output display using docks:
external

If the output pane is present, output is redirected to this pane, only the spinner and status icons stay in the text editor.

I don't know what the best behavior for this panel would be. We can iterate one this an collect feedback from users.

Fixes #760

### DIFF
--- a/lib/components/output-area.js
+++ b/lib/components/output-area.js
@@ -18,11 +18,11 @@ const OututArea = observer(({ store: { kernel } }: { store: store }) => {
     <div>
       {kernel.outputStore.outputs.length > 0
         ? <div
-            className="btn btn-error icon icon-trashcan"
+            className="btn icon icon-trashcan"
             onClick={kernel.outputStore.clear}
             style={{
-              left: "50%",
-              transform: "translate(-50%, 0)",
+              left: "100%",
+              transform: "translateX(-100%)",
               position: "relative"
             }}
           >
--- a/lib/components/output-area.js
+++ b/lib/components/output-area.js
@@ -1,0 +1,37 @@
+/* @flow */
+
+import { CompositeDisposable } from "atom";
+import React from "react";
+import { observer } from "mobx-react";
+
+import History from "./result-view/history";
+import { OUTPUT_AREA_URI } from "./../utils";
+
+import typeof store from "../store";
+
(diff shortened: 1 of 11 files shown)


rgbkrk:
"Love the sidecar. Keeps things tidy." - @ivanov

ivanov:
As Kyle said, this looks great, @lgeiger! Another reasonable possibility for the outputs would be too keep adding them  onto a single page on the left, and scroll down so that the last one comes into view... that way you can scroll down through them and see multiple (for example, if you're just printing one line per output, it'd be nice to see a bunch of them as opposed to having to click to page through them. 

On the other hand, the current implementation is perfect for comparing images, flipping back and forth between them, so there are use cases for both.

lgeiger:
@ivanov Thanks a lot for the feedback.

Yeah I though about this too. I didn't go with it because a infinite scrolling list might become a bit messy with time, especially with multiple rich outputs.

One thing to note is that we accumulate stdout (primarily for properly handling ANSI escapes):

stdout

I think this behavior is a good compromise between the ability of comparing multiple rich outputs while allowing users to read large amounts of stdout without constantly flipping through the history.

lib/components/output-area.js
```diff
@@ -0,0 +1,37 @@
...
+import { CompositeDisposable } from "atom";
+import React from "react";
+import { observer } from "mobx-react";
+
+import History from "./result-view/history";
+import { OUTPUT_AREA_URI } from "./../utils";
+
+import typeof store from "../store";
+
+const OututArea = observer(({ store: { kernel } }: { store: store }) => {
```
rgbkrk: `const OutputArea`

lgeiger: 👍  Good catch

lgeiger:
@BenRussert The great thing about our new panels is that users can re-arange them however they like:
ezgif-2-19fc5f53bc

Totally agree, this is only the start. I'm happy to explore different UI's (e.g. a long scrolling list, or different output areas for different mime types, etc.)

BenRussert:
@lgeiger we should save that gif for the release notes and docs

PR #826, opened 2017-05-30
Try out new IPython completer API

lgeiger:
Implement support for https://github.com/ipython/ipykernel/pull/222
bildschirmfoto 2017-05-30 um 15 04 38

Thanks @Carreau for the great work!

One thing I noticed while testing is that the new completions are noticable slower than before, though speed improves a bit with time once the jedi cache gets populated.

### DIFF
--- a/lib/autocomplete-provider.js
+++ b/lib/autocomplete-provider.js
@@ -10,6 +10,20 @@ type Autocomplete = {
   prefix: string
 };
 
+type CompleteReply = {
+  matches: Array<string>,
+  cursor_start: number,
+  cursor_end: number,
+  metadata?: {
+    _jupyter_types_experimental?: Array<{
+      start?: number,
+      end?: number,
+      text?: string,
+      type?: string
+    }>
+  }
+};
+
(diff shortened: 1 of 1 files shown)


Carreau:
> One thing I noticed while testing is that the new completions are noticable slower than before, though speed improves a bit with time once the jedi cache gets populated.

Yes, we had a couple of complaint about that. I need to find some time to work on jedi.

rgbkrk:
Does it also bust the cache if you upgrade a package?

Carreau:
I believe the cache is based on the hash of the files content. I'm unsure if it deletes the cache, but at least it won't give outdated result.

rgbkrk:
IPython master now has this and we've merged it for nteract desktop to start iterating on. What do you think about bringing this in @lgeiger?

Carreau:
> IPython master

Technically IPython stable and IPykernel master.

lgeiger:
> IPython master now has this and we've merged it for nteract desktop to start iterating on. What do you think about bringing this in @lgeiger?

Should be good to bring in 🚀

PR #829, opened 2017-05-30
fix(package): update @jupyterlab/services to version 0.44.0

rgbkrk:
Closes #768


rgbkrk:
Sometimes doing stuff with greenkeeper and the GitHub UI is good, sometimes it is not. 
¯\\\_(ツ)\_/¯

PR #832, opened 2017-05-31
Dano flow

jdpigeon:
Added flow types to kernel-manager (and jupyter-js-services-shim, but there wasn't anything to do)

### DIFF
--- a/lib/kernel-manager.js
+++ b/lib/kernel-manager.js
@@ -33,10 +33,10 @@ class KernelManager {
 
   startKernelFor(grammar: atom$Grammar, onStarted: (kernel: Kernel) => void) {
     try {
-      // Don't know where atom.project.rootDirectories is and why it's not found
+      // Don't know where atom$Project is and why it doesn't have rootDirectories property
       const rootDirectory = atom.project.rootDirectories[0]
         ? atom.project.rootDirectories[0].path
-        : // Should null be an option for getEditorDirectory parameter types?
+        : // Should null be an option for getEditorDirectory's atom$TextEditor parameter?
           getEditorDirectory(store.editor);
       const connectionFile = path.join(
         rootDirectory,
@@ -66,7 +66,7 @@ class KernelManager {
     });
   }
 
-  // Flow is not aware of Connection
+  // Flow is not aware of Connection, going with Object
   startExistingKernel(
     grammar: atom$Grammar,
     connection: Object,
@@ -84,7 +84,7 @@ class KernelManager {
       env: {}
     };
 
-    // Expects another argument, kernelProcess: child_process$ChildProcess,. Should ZMQKernel have that be nullable?
+    // Expects another argument, kernelProcess: child_process$ChildProcess. Should that be nullable?
     const kernel = new ZMQKernel(
       kernelSpec,
       grammar,
--- a/lib/kernel-manager.js
+++ b/lib/kernel-manager.js
@@ -1,4 +1,3 @@
-"use babel";
 /* @flow */
 
 import _ from "lodash";
@@ -9,8 +8,9 @@ import path from "path";
 
 import Config from "./config";
 import ZMQKernel from "./zmq-kernel";
-// I imported Kernel so it could be a valid parameter type
+import type { Connection } from "./zmq-kernel";
 import Kernel from "./kernel.js";
+
 import KernelPicker from "./kernel-picker";
 import store from "./store";
 import { getEditorDirectory, log } from "./utils";
@@ -31,13 +31,15 @@ class KernelManager {
     this._kernelSpecs = this.getKernelSpecsFromSettings();
   }
 
-  startKernelFor(grammar: atom$Grammar, onStarted: (kernel: Kernel) => void) {
+  startKernelFor(
+    grammar: atom$Grammar,
+    editor: ?atom$TextEditor,
+    onStarted: (kernel: ZMQKernel) => void
+  ) {
     try {
       // Don't know where atom$Project is and why it doesn't have rootDirectories property
-      const rootDirectory = atom.project.rootDirectories[0]
-        ? atom.project.rootDirectories[0].path
-        : // Should null be an option for getEditorDirectory's atom$TextEditor parameter?
-          getEditorDirectory(store.editor);
+      const rootDirectory =
+        atom.project.getPaths()[0] || getEditorDirectory(editor);
       const connectionFile = path.join(
         rootDirectory,
         "hydrogen",
@@ -69,9 +71,9 @@ class KernelManager {
   // Flow is not aware of Connection, going with Object
   startExistingKernel(
     grammar: atom$Grammar,
-    connection: Object,
+    connection: Connection,
     connectionFile: string,
-    onStarted: (kernel: Kernel) => void
+    onStarted: (kernel: ZMQKernel) => void
   ) {
     const language = grammar.name;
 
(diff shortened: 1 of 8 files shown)


lib/kernel-manager.js
```diff
@@ -1,4 +1,5 @@
 "use babel";
+/* @flow */
```
lgeiger: `"use babel";`  isn't required anymore since we use `/* @flow */`

lib/kernel-manager.js
```diff
@@ -8,20 +9,35 @@ import path from "path";
...
+  };
+  */
+  _kernelSpecs: ?Object;
+  kernelPicker: KernelPicker;
   constructor() {
     this._kernelSpecs = this.getKernelSpecsFromSettings();
   }
 
-  startKernelFor(grammar, onStarted) {
+  startKernelFor(grammar: atom$Grammar, onStarted: (kernel: Kernel) => void) {
```
lgeiger: Actually I think we should use `(kernel: ZMQKernel) => void` here. Then we can probably remove [this line](https://github.com/nteract/hydrogen/pull/832/files#diff-67b721d4904ee45505901d8e7e272c5fR13).

lib/kernel-manager.js
```diff
@@ -8,20 +9,35 @@ import path from "path";
...
     try {
+      // Don't know where atom$Project is and why it doesn't have rootDirectories property
       const rootDirectory = atom.project.rootDirectories[0]
```
lgeiger: This is probably only missing in out flow type definitions. We should add it there after looking at Atoms source code.

lib/kernel-manager.js
```diff
@@ -8,20 +9,35 @@ import path from "path";
...
         ? atom.project.rootDirectories[0].path
-        : getEditorDirectory(store.editor);
+        : // Should null be an option for getEditorDirectory's atom$TextEditor parameter?
```
lgeiger: Either that or we should pass a editor variable directly to `startKernelFor` and check before if `editor` isn't null.

lib/kernel-manager.js
```diff
@@ -50,7 +66,13 @@ class KernelManager {
     });
   }
 
-  startExistingKernel(grammar, connection, connectionFile, onStarted) {
+  // Flow is not aware of Connection, going with Object
```
lgeiger: You can export the type [here](https://github.com/nteract/hydrogen/blob/master/lib/zmq-kernel.js#L12) and import it with:
```js
import type { Connection } from "./zmq-kernel.js";
```

lib/kernel-manager.js
```diff
@@ -62,6 +84,7 @@ class KernelManager {
       env: {}
     };
 
+    // Expects another argument, kernelProcess: child_process$ChildProcess. Should that be nullable?
```
lgeiger: You're totally right!

This should definitely nullable

lib/kernel-manager.js
```diff
@@ -125,17 +152,18 @@ class KernelManager {
     });
   }
 
-  _executeStartupCode(kernel) {
+  _executeStartupCode(kernel: Kernel) {
```
lgeiger: `Kernel` --> `ZMQKernel`

lib/kernel-manager.js
```diff
@@ -125,17 +152,18 @@ class KernelManager {
...
     const displayName = kernel.kernelSpec.display_name;
     let startupCode = Config.getJson("startupCode")[displayName];
     if (startupCode) {
       log("KernelManager: Executing startup code:", startupCode);
       startupCode = `${startupCode} \n`;
+      // Looks like kernel.execute hasn't been implemented. When it is, will probably want to add an onResults fn to this call
```
lgeiger: Should work if you try my comment above 😉

lib/kernel-manager.js
```diff
@@ -264,13 +292,14 @@ class KernelManager {
...
-  getKernelSpecsFrom(command, callback) {
+  getKernelSpecsFrom(command: string, callback: Function) {
     const options = { killSignal: "SIGINT" };
     let kernelSpecs;
     return exec(command, options, (err, stdout) => {
       if (!err) {
         try {
-          kernelSpecs = JSON.parse(stdout).kernelspecs;
+          // To avoid type error with stdout (a Buffer), threw in a quick toString
+          kernelSpecs = JSON.parse(stdout.toString()).kernelspecs;
```
lgeiger: 👍 🎉

lib/kernel-manager.js
lgeiger: A better way of doing this would be:
```js
const rootDirectory = atom.project.getPaths()[0] || getEditorDirectory(store.editor);
```
This way we use documented APIs and flow won't complaint 😄

lib/kernel-manager.js
```diff
@@ -215,11 +243,11 @@ class KernelManager {
     );
   }
 
-  mergeKernelSpecs(kernelSpecs) {
+  mergeKernelSpecs(kernelSpecs: Kernelspec) {
     _.assign(this._kernelSpecs, kernelSpecs);
   }
 
-  updateKernelSpecs(callback) {
+  updateKernelSpecs(callback: Function) {
```
lgeiger: We shouldn't require a callback here:
`Function` -> `?Function`

jdpigeon: Hey man, I made your recommended changes and managed to get the number of errors in kernel-manager down to 1! However, I've made some things nullable in other files that lead to some more flow errors

lib/kernel-manager.js
```diff
@@ -31,13 +31,15 @@ class KernelManager {
     this._kernelSpecs = this.getKernelSpecsFromSettings();
   }
 
-  startKernelFor(grammar: atom$Grammar, onStarted: (kernel: Kernel) => void) {
+  startKernelFor(
+    grammar: atom$Grammar,
+    editor: ?atom$TextEditor,
```
lgeiger: We now need to change the function call [here](https://github.com/nteract/hydrogen/blob/master/lib/main.js#L305) and [here](https://github.com/nteract/hydrogen/blob/master/lib/main.js#L435)

lib/kernel-manager.js
```diff
@@ -152,13 +154,13 @@ class KernelManager {
...
-  _executeStartupCode(kernel: Kernel) {
+  _executeStartupCode(kernel: ZMQKernel) {
     const displayName = kernel.kernelSpec.display_name;
     let startupCode = Config.getJson("startupCode")[displayName];
     if (startupCode) {
       log("KernelManager: Executing startup code:", startupCode);
       startupCode = `${startupCode} \n`;
-      // Looks like kernel.execute hasn't been implemented. When it is, will probably want to add an onResults fn to this call
+      // execute in ZMQKernel still wants another parameter
       kernel.execute(startupCode);
```
lgeiger: We should make the [callback](https://github.com/nteract/hydrogen/blob/master/lib/zmq-kernel.js#L196) optional.

lib/utils.js
```diff
@@ -97,10 +97,11 @@ export function getEmbeddedScope(
   return _.find(scopes, s => s.indexOf("source.embedded.") === 0);
 }
 
-export function getEditorDirectory(editor: atom$TextEditor) {
-  const editorPath = editor.getPath();
-  return editorPath ? path.dirname(editorPath) : os.homedir();
+// I feel like this should work
+export function getEditorDirectory(editor: ?atom$TextEditor) {
+  return editor ? path.dirname(editor.getPath()) : os.homedir();
```
lgeiger: I think we should handle the case `editor == null` higher up. I'm not sure if this works. `editor.getPath()` still can return undefined.

lib/kernel-manager.js
jdpigeon: Done

lgeiger:
@jdpigeon Welcome to nteract 🎉

PR #835, opened 2017-06-01
Add auto-scroll feature

### LINKED ISSUE #774: [Feature Request] Auto Scroll
When you run a block of code that has a lot of output, it would be awesome if the inline window would auto-scroll to the bottom with a button on the inline window that would toggle auto-scroll on or off (next to the copy to clip board and open in editor buttons).

dwolfschlaeger:
Add basic functionality for the auto-scroll feature.
Special thanks to @lgeiger for helping me getting started!
Fixes #774

### DIFF
--- a/lib/components/result-view/result-view.js
+++ b/lib/components/result-view/result-view.js
@@ -139,6 +139,22 @@ type Props = { store: OutputStore, destroy: Function, showResult: boolean };
     );
   }
 
+  scrollToBottom() {
+    if (!this.el) return;
+    const scrollHeight = this.el.scrollHeight;
+    const height = this.el.clientHeight;
+    const maxScrollTop = scrollHeight - height;
+    this.el.scrollTop = maxScrollTop > 0 ? maxScrollTop : 0;
+  }
+
+  componentDidUpdate() {
+    this.scrollToBottom();
+  }
+
+  componentDidMount() {
+    this.scrollToBottom();
+  }
+
   componentWillUnmount() {
     this.containerTooltip.dispose();
     this.buttonTooltip.dispose();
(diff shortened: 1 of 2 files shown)


lib/components/result-view/result-view.js
```diff
@@ -139,6 +139,22 @@ type Props = { store: OutputStore, destroy: Function, showResult: boolean };
     );
   }
 
+  scrollToBottom() {
+    if (!this.el) return;
```
lgeiger: We should also return for the cases if `this.expanded === true` or `this.props.isPlain === true` since we don't want to scroll in those cases.

lgeiger:
Welcome to nteract 🎉 🚀

joolius:
Does not seem to work for me even with the option enabled. Anyone else have this issue?

wadethestealth:
@joolius this autoscroll should happen inside the results view. So if you have a for loop that counts to 1000. You should see 1000 rather than 0.

shr00mie:
> Does not seem to work for me even with the option enabled. Anyone else have this issue?

@juliuscrn 

yeah. it's been driving me a bit batty for a while. the behavior i'm seeing is that the output window won't autoscroll during the generation of output, but WILL jump all the way to the end once the kernel has finished running, which...is i don't think what the intended functionality of autoscroll was meant as. i want to see output like i would in a terminal window, i.e. as output is generated, newest output is the bottom line and oldest line gets scrolled out past whatever the window height is. I've tried a full reinstall. manually setting autoScroll under config. no dice. i'm a bit hesitant to start fiddling with the actual config files as that's not my language of choice, so i'd be spitballing at best.

wadethestealth:
@shr00mie you should open a separate issue for that issue. But a code example and gif would be extremely helpful because intended functionality should be every time the output component is updated.

shr00mie:
@wadethestealth Hey, Thomas. Think you assisted me with something last time around when you guys were implementing a new feature. Hope you're doing well. As requested, i'll open up a new issue for this. just grabbed a video. will convert and upload with some example code of behavior.

PR #836, opened 2017-06-01
Added spec tests for output.js

pfackeldey:
Handled case for negative index in `setIndex` function.


lgeiger:
Welcome to nteract 🎉

PR #853, opened 2017-06-05
Keep ZMQKernel class if restarted

### LINKED ISSUE #849: Restarting kernel with external output area hides entire UI area
**Description**

When the kernel is restarted, the UI area containing the external output area (EOA) is hidden. 

For example, if the EOA is in the same UI element as the project path view, either as a tab:

screen shot 2017-06-02 at 18 43 35

or above or below in the same sidebar:

screen shot 2017-06-02 at 18 44 00

Restarting the kernel will hide the 'project' tab. 

[Here are (I'm pretty sure) the lines that cause this behavior.](https://github.com/nteract/hydrogen/pull/825/files#diff-06c78041ac8e1eb79621efd8c87dc106R13)

Does hiding the EOA, in this context, serve a purpose? If it is, could it be configurable in settings? Having to re-open sidebars every time I restart the kernel is a little annoying.

lgeiger:
This makes restarting a lot smother (and a bit faster):
- only spawn a new kernel process and keep the ZMQ sockets open
- don't clear the result bubbles on restart.

restart

Fixes #849 
/cc @wowserx

### DIFF
--- a/lib/kernel-manager.js
+++ b/lib/kernel-manager.js
@@ -120,17 +120,22 @@ class KernelManager {
     }
 
     const kernelStartDir = projectPath != null ? projectPath : currentPath;
-
-    launchSpec(kernelSpec, {
+    const options = {
       cwd: kernelStartDir,
       stdio: ["ignore", "pipe", "pipe"]
-    }).then(({ config, connectionFile, spawn }) => {
+    };
+
+    launchSpec(
+      kernelSpec,
+      options
+    ).then(({ config, connectionFile, spawn }) => {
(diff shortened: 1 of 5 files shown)


rgbkrk:
Wow, sweet. I'm glad that worked so well.

wowserx:
Wow, this is awesome! Y'all are the speedy, communicative devs I aspire to become.

t9md:
With this PR, `onRestarted` callback is no longer called.
I want give it back, now plugin author have no control when they want to do after restart.

PR #855, opened 2017-06-05
Add deprecation note for connecting to existing zmq kernels

lgeiger:
We've deprecated connecting to existing ZMQ kernels via a `connection.json` file via the docs in favor of remote kernels for some time now. I think using remote kernels is superior in any way.

This will add a deprecation note if the feature is used so we can get feedback if there are use-cases we haven't thought about:
bildschirmfoto 2017-06-05 um 19 20 13

Removing this feature will allow us to simplify our kernel handling a lot. In particular this part of the code base made quite some trouble when trying to properly type it in #832.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -6,6 +6,8 @@ import ReactDOM from "react-dom";
 import _ from "lodash";
 import os from "os";
 import path from "path";
+// $FlowFixMe
+import { shell } from "electron";
 
 import Config from "./config";
 import store from "./store";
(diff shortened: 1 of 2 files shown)


lib/utils.js
```diff
@@ -134,3 +136,29 @@ export function renderDevTools() {
...
+  atom.notifications.addWarning("This feature will be deprecated soon!", {
+    description:
+      "Connecting to existing kernels via a `connection.json` file will be deprecated soon.\n\nFor some time now Hydrogen supports using [kernel gateways](https://nteract.gitbooks.io/hydrogen/docs/Usage/RemoteKernelConnection.html) for connection to existing kernels. Using that option is a lot simpler while being powerful.\n\nPlease get in touch with us if using remote kernels isn't a option for you.",
+    dismissable: true,
+    buttons: [
+      {
+        className: "icon icon-x",
+        text: "I really need this feature",
+        onDidClick: () => {
+          shell.openExternal("https://github.com/nteract/hydrogen/issues/new");
```
rgbkrk: You should probably make this link to just one issue where users can click to 👍  or 👎

nikitakit: I agree, encouraging everyone to open their own issue may flood the tracker.

lgeiger: Good point! I don't exspect a lot of people using this feature so I was a bit sloppy 😄

PR #857, opened 2017-06-05
ZMQ kernel should call restart-callback on light-restart

t9md:
By #853, restart callback become just ignored on lightweight-restart.
This PR gives it back, means call callback on light-restart again.


t9md:
thanks for quick merging!

Issue #796, opened 2017-05-20
Hide unnecessary buttons in result view


lgeiger:
Depending on the output not every button is useful. To make the appearance cleaner we should hide unnecessary buttons:
![bildschirmfoto 2017-05-21 um 00 36 53](https://cloud.githubusercontent.com/assets/13285808/26279923/6cd624f0-3dc2-11e7-893c-f2ede159fd6d.png)

- [x] The [copy](https://github.com/nteract/hydrogen/blob/master/lib/components/result-view/result-view.js#L116) and [open in editor](https://github.com/nteract/hydrogen/blob/master/lib/components/result-view/result-view.js#L121)  should only be shown if there's actually [text to copy](https://github.com/nteract/hydrogen/blob/master/lib/components/result-view/result-view.js#L23).
- [x] The [expand button](https://github.com/nteract/hydrogen/blob/master/lib/components/result-view/result-view.js#L125) should only be shown if the result doesn't fit in the output box.
Pseudo code (not tested):
```js
import { DEFAULT_SCROLL_HEIGHT } from "@nteract/display-area"

...

{(this.el.scrollHeight > DEFAULT_SCROLL_HEIGHT`)
  ? <div
      className={`icon icon-${this.expanded.get() ? "fold" : "unfold"}`}
      onClick={this.toggleExpand}
    />
  : null}
...
```

nikitakit:
Note that the result number (execution count) is currently shown on-hover over the copy button, so if the copy button is hidden there may need to be some alternative way of surfacing that information.

lgeiger:
> Note that the result number (execution count) is currently shown on-hover over the copy button, so if the copy button is hidden there may need to be some alternative way of surfacing that information.

Good point. What about moving the tooltip to the close button?

nikitakit:
Not sure about that. The information provided may be only tangentially related to the copy action, but it's even less related to the close action. Plus, if your fingers slip and you end up clicking then the bubble is gone -- and the result number is precisely what you need to know to bring it back.

This might be an OK compromise in the interim, but I really think there should be a better way of providing this information.

lgeiger:
@nikitakit I agree.

I think with the new unified result view (and the removal of the rich display) we can find a place to permanently show the execution count in the result view. I'm very open to ideas!

By the way how do you bring back the result based on the execution count?

nikitakit:
@lgeiger 

[...]

PR #859, opened 2017-06-05
Hide expand button

dwolfschlaeger:
Expand-button is now hidden if the height of the content of the resultview is smaller than the clientHeight.
#796

### DIFF
--- a/lib/components/result-view/result-view.js
+++ b/lib/components/result-view/result-view.js
@@ -6,7 +6,7 @@ import * as Immutable from "immutable";
 import { observer } from "mobx-react";
 import { action, observable } from "mobx";
 import { Display } from "@nteract/display-area";
-
+import { DEFAULT_SCROLL_HEIGHT } from "@nteract/display-area";
 import { transforms, displayOrder } from "./transforms";
 import Status from "./status";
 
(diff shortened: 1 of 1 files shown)


lib/components/result-view/result-view.js
```diff
@@ -6,16 +6,15 @@ import * as Immutable from "immutable";
 import { observer } from "mobx-react";
 import { action, observable } from "mobx";
 import { Display } from "@nteract/display-area";
-
+import { DEFAULT_SCROLL_HEIGHT } from "@nteract/display-area";
```
lgeiger: Since you're not using `DEFAULT_SCROLL_HEIGHT` we can remove this line

lib/components/result-view/result-view.js
```diff
@@ -131,12 +130,12 @@ class ResultViewComponent extends React.Component {
                 className="icon icon-file-symlink-file"
                 onClick={this.openInEditor}
               />
-              <div
-                className={`icon icon-${this.expanded.get()
-                  ? "fold"
-                  : "unfold"}`}
-                onClick={this.toggleExpand}
-              />
+              {this.el.scrollHeight > this.el.clientHeight
```
lgeiger: This won't work, since `this.el` can be null (see [flow type errors](https://travis-ci.org/nteract/hydrogen/jobs/239749187#L140)). We need to guard against that case.

I can recommend the [`flow-ide`](https://github.com/steelbrain/flow-ide) package for seeing the flow type-checking result directly in Atom.

PR #861, opened 2017-06-07
add refresh-all-result-bubbles command

t9md:
refresh-all-result-bubbles

# What this PR do?

Add new command `hydrogen:refresh-all-result-bubbles`

This command do
- restart and runAll by **not** clearing existing breakpoints(start position of existing result-bubbles).

Detailed explanation for how this command works
1. collect existing breakpoints(start point of existing bubbles) from editor.
2. clear result-bubbles
3. restart
4. runAll( this is passed as callback of `restart()`)

I thinks `all` is a bit verbose, but want to make it explicit.
But feel free to change something like `hydrogen:refresh-result-bubbles`

# Motivation

When I want to re-evaluate whole code in editor, there is no way to refresh existing result-bubbles.
Currently it's just not cleared by `run-all-above` or `run-all`, but it's just showing old slate result.
I want make sure currently visible result-bubbles are **all refreshed, believable** and as long as evaluate code from top-to-bottom from clean state.

This command allow "all visible bubble is refreshed by this command from clean state(kernel restarted to clear out scope)".

# Side-node(skip if you haven't read original issue comment and code)

First, I tried to **merge** existing breakpoints(start point of existing bubbles) with detected breakpoint(e.g. `%%` in comment).
But quit this approach.
- It doesn't fit `refresh-result-bubbles`, since detecting bp in comment add new bubbles.
- Handling position diff between detected breakpoint(start of comment) and start of bubble was pain.
- These two position have relation, but not equal, bubble is created at point after skipping line-comment upward.
- Because of two these position diff, **merging** is not easy, and I also noticed this is not necessary, so simplified to current state.

# Next step I want to add

But postponed since I'm not sure **this** PR is acceptable first of all.

Currently there is two way to add breakpoints.
- Add special chars in comment(e.g. `%%`).
- via `hydrogne:run-all-above` command

I want to add breakpoint
- without modifying editor, just add marker at end of line with block decoration with empty content.
- without evaluating code. This is important to avoid `x is already defined` error with language it force "one-time declaration"(e.g. JavaScript).

When that's done, the workflow I'm imaging is this
1. place breakpoints by `hydrogen:toggle-breakpoints`
  - No editor modification, no code-evaluation.
  - Just add/remove blank bubbles on each selected lines.
2. `hydrogen:refresh-all-result-bubbles`

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -115,6 +115,8 @@ const Hydrogen = {
           this.handleKernelCommand({ command: "interrupt-kernel" }),
         "hydrogen:restart-kernel": () =>
           this.handleKernelCommand({ command: "restart-kernel" }),
+        "hydrogen:refresh-all-result-bubbles": () =>
+          this.refreshAllResultBubbles(),
         "hydrogen:shutdown-kernel": () =>
           this.handleKernelCommand({ command: "shutdown-kernel" })
       })
@@ -250,10 +252,12 @@ const Hydrogen = {
 
   handleKernelCommand({
     command,
-    payload
+    payload,
+    onRestarted
   }: {
     command: string,
-    payload: ?Kernelspec
+    payload: ?Kernelspec,
+    onRestarted: ?Function
   }) {
     log("handleKernelCommand:", arguments);
 
@@ -281,7 +285,7 @@ const Hydrogen = {
     if (command === "interrupt-kernel") {
       kernel.interrupt();
     } else if (command === "restart-kernel") {
-      kernel.restart();
+      kernel.restart(onRestarted);
     } else if (command === "shutdown-kernel") {
       this.clearResultBubbles();
       // Note that destroy alone does not shut down a WSKernel
@@ -388,6 +392,19 @@ const Hydrogen = {
     this.markerBubbleMap = {};
   },
 
+  refreshAllResultBubbles() {
+    let breakpoints = [];
+    _.forEach(this.markerBubbleMap, (bubble: ResultView) => {
+      breakpoints.push(bubble.marker.getBufferRange().start);
+    });
+    this.clearResultBubbles();
+
+    this.handleKernelCommand({
+      command: "restart-kernel",
+      onRestarted: () => this.runAll(breakpoints)
+    });
+  },
+
   clearBubblesOnRow(row: number) {
     log("clearBubblesOnRow:", row);
     _.forEach(this.markerBubbleMap, (bubble: ResultView) => {
@@ -419,7 +436,7 @@ const Hydrogen = {
     }
   },
 
-  runAll() {
+  runAll(oldBreakpoints: ?Array<atom$Point>) {
     const { editor, kernel, grammar } = store;
     if (!editor || !grammar) return;
     if (isMultilanguageGrammar(editor.getGrammar())) {
@@ -430,17 +447,21 @@ const Hydrogen = {
     }
 
     if (editor && kernel) {
-      this._runAll(editor, kernel);
+      this._runAll(editor, kernel, oldBreakpoints);
       return;
     }
 
     kernelManager.startKernelFor(grammar, editor, (kernel: ZMQKernel) => {
-      this._runAll(editor, kernel);
+      this._runAll(editor, kernel, oldBreakpoints);
     });
   },
 
-  _runAll(editor: atom$TextEditor, kernel: Kernel) {
-    const cells = codeManager.getCells(editor);
+  _runAll(
+    editor: atom$TextEditor,
+    kernel: Kernel,
+    oldBreakpoints: ?Array<atom$Point>
+  ) {
+    let cells = codeManager.getCells(editor, oldBreakpoints);
     _.forEach(
       cells,
       ({ start, end }: { start: atom$Point, end: atom$Point }) => {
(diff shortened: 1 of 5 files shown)


t9md:
Sorry, I found bug, so this PR is still WIP status.

t9md:
done my part. I wanted to correct my mistake of commit msg(braek pint > break point), but It was not easy for me.

Tried rebasing and force commit to my branch, but github strangely ask password where they don't ask in normal commit, maybe because I'm not understanding git/github correctly.
So give up for now to correct commit msg, sorry.

t9md:
I found another bug again, so **WIP** again.
Travis error is not related, it's caused by newest prettier.

t9md:
fix the bug, done again. no WIP state, but passing callback on restart have issue.
So I need to add delay for python kernel.

So I'll wait for comment.

rgbkrk:
This is super sweet, I'll probably use this with my READMEs too. 😄

lib/main.js
```diff
@@ -388,6 +392,26 @@ const Hydrogen = {
...
+    }
+
+    let breakpoints = [];
+    _.forEach(this.markerBubbleMap, (bubble: ResultView) => {
+      breakpoints.push(bubble.marker.getBufferRange().start);
+    });
+    this.clearResultBubbles();
+
+    const runAll = () => this.runAll(breakpoints);
+    this.handleKernelCommand({
```
lgeiger: Rather than modifying `handleKernelCommand` I would use `kernel.restart()` directly here.

spec/code-manager-spec.js
```diff
@@ -59,4 +60,80 @@ describe("CodeManager", () => {
       expect(editor.getSelectedText).toHaveBeenCalled();
     });
   });
+
+  describe("cells", () => {
```
lgeiger: Wuhu thanks for adding tests 👍 🎉

lib/main.js
```diff
@@ -388,6 +392,26 @@ const Hydrogen = {
...
+      command: "restart-kernel",
+      onRestarted: () => setTimeout(runAll, 500) // FIXME: see#863 need discussion.
```
lgeiger: This should work for now 👍

t9md: why? what changed? I couldn't find no related commit on hydrogen repo.

lgeiger: > why? what changed? I couldn't find no related commit on hydrogen repo.

I'm just referencing the `setTimeout(runAll, 500)` workaround for #863. 👍

lib/main.js
```diff
@@ -115,6 +115,8 @@ const Hydrogen = {
           this.handleKernelCommand({ command: "interrupt-kernel" }),
         "hydrogen:restart-kernel": () =>
           this.handleKernelCommand({ command: "restart-kernel" }),
+        "hydrogen:refresh-all-result-bubbles": () =>
```
t9md: This command redraw(refresh) only existing bubbles, so `bubble` should be included as part of command name.
`restart-kernel-and-re-evaluate-bubbles`? but its's verbose.

`restart` is just a underlying detail to achieve refresh, if kernel provide more light way to cleanup env( like node's vm.createContext ), it also can be used to achieve refresh.
So, in my opinion I prefer current name. how do you think?

lib/main.js
t9md: Oh, understand, thanks for clarification.

t9md:
> There's one thing I'm unsure how to handle best.
> Hydrogen can execute code on a selection basis. This means a result bubble doesn't always equal a break point:

I have no idea to solve this issue, I want to say just let user handle that for now.

As I wrote before, I want to add command to manually add/remove bubbles.
If this command was introduced, removing only problematic bubbles become more easy, user don't have to clear all bubbles.

t9md:
For command name I myself can't find better name.
@lgeiger , feel free to change name.
If you are clear for command name, please say me so again, I can update.

lib/main.js
lgeiger: Thanks for explaining the reasoning behind it :+1:

@rgbkrk @BenRussert What do you think, should we go with `refresh-all-result-bubbles`?

t9md: [INFO] What this command is doing is `cleanly-evaluate-code-from-top-to-bottom-and-update-existing-bubbles`, so we have to find better name of this verbose command name.

rgbkrk: Ah yes, right at my fingertips. Luckily with the command palette you only have to type some subset of that.

rgbkrk: I lean towards `restart-kernel-and-re-evaluate-bubbles`. Restart is currently the only language agnostic way we have of cleanly evaluating code.

BenRussert: I think `restart-kernel-and-re-evaluate-bubbles` or `refresh-all-result-bubbles` are fine. I'll probably just start typing "bubbles"

Whatever we go with we can just explain the specifics in the docs 📖

lgeiger: > I lean towards `restart-kernel-and-re-evaluate-bubbles`. Restart is currently the only language agnostic way we have of cleanly evaluating code.

👍 I like it, It's a bit long but Atom has a great command palette so this is OK.

lgeiger:
Welcome to nteract :tada: :rocket:

rgbkrk:
I'm currently in an Rx workshop and this feature has been a life saver. 😄

PR #865, opened 2017-06-08
Remove ESLINT and pin prettier

lgeiger:
Currently we only use ESLINT for checking that our code is prettified correctly. This means we run into issues when there is a prettier update (#864).

This PR removes our ESLINT setup and pins `prettier`.  It uses [`prettier-greenkeeper`](https://github.com/lgeiger/prettier-greenkeeper) to run prettier on Greenkeeper PRs so we don't run into CI failures on every prettier update and it removes the need for manually running prettier.


rgbkrk:
Thank you, smart choice!

PR #866, opened 2017-06-09
Add empty display message

### LINKED ISSUE #834: Display nice empty message in external display area.
We added a new external output area in #825.

Right now, if there is no output to display we just show a empty pane. We should at least render a nicely styled empty message.

BenRussert:
Closes #834 

Thought simple worked here, open to feedback of course.

empty-display

### DIFF
--- a/lib/components/result-view/history.js
+++ b/lib/components/result-view/history.js
@@ -56,7 +56,7 @@ const History = observer(({ store }: { store: OutputStore }) => {
           />
         </div>
       </div>
-    : null;
+    : <ul className="background-message centered">No output to display</ul>;
 });
 
 export default History;


rgbkrk:
Works for me, certainly better than the current `null`. We can iterate.

lib/components/result-view/history.js
```diff
@@ -56,7 +56,7 @@ const History = observer(({ store }: { store: OutputStore }) => {
           />
         </div>
       </div>
-    : null;
+    : <ul className="background-message centered">No output to display</ul>;
```
rgbkrk: Now that it clicks for me this should be a `<p>` tag or a standalone `<div>`.

BenRussert: The reason I used `<ul>` was so the atom styling for `className="background-message centered"` would work (it doesnt if I just wrap in a `<div>`). Is it better if I wrap the ul tag in a `<div>`?

styleguide

PR #874, opened 2017-06-15
Suppress logs in dev mode, closes #872

### LINKED ISSUE #872: [suggestion] Suppress console.log in dev-mode
**Description:**

When atom is running in dev mode(`atom -d .`), hydrogen write tons of log on chrome dev console.
This verbosity make more important log unnoticed/or difficult to find.

## Background: Example two situation why I want to suppress hydrogen's log

- While I'm hacking hydrogen, I add ad hoc console.log frequently, but without suppressing default log, finding this console.log is super difficult and easily scrolled over.
- I launch atom in dev mode to develop my own package, or Atom-core itself, and use hydrogen to check behavior of some code, but in this situation, I don't have interest of hydrogen's log itself. I want to watch log from other packages I'm working on, but hydrogen's log occupy console log.

**Steps to Reproduce:**

1. Launch atom with `-d` flag(run Atom in dev mode).
2. Start hydrogen and open chrome dev-console and see how much log is written on console"

## Suggestion to change

Make hydrogen write verbose log only when `Hydrogen.debug` was explicitly enabled regardless of atom's running mode.

BenRussert:
See #872, this is not urgent so we can wait if more people want to weigh in.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -117,13 +117,13 @@ export function getEditorDirectory(editor: ?atom$TextEditor) {
 }
 
 export function log(...message: Array<any>) {
-  if (atom.inDevMode() || atom.config.get("Hydrogen.debug")) {
+  if (atom.config.get("Hydrogen.debug")) {
     console.trace("Hydrogen:", ...message);
   }
 }
 
 export function renderDevTools() {
-  if (atom.inDevMode() || atom.config.get("Hydrogen.debug")) {
+  if (atom.config.get("Hydrogen.debug")) {
     try {
       const devTools = require("mobx-react-devtools");
       const div = document.createElement("div");


lib/utils.js
```diff
@@ -117,13 +117,13 @@ export function getEditorDirectory(editor: ?atom$TextEditor) {
...
 export function log(...message: Array<any>) {
-  if (atom.inDevMode() || atom.config.get("Hydrogen.debug")) {
+  if (atom.config.get("Hydrogen.debug")) {
     console.trace("Hydrogen:", ...message);
   }
 }
 
 export function renderDevTools() {
-  if (atom.inDevMode() || atom.config.get("Hydrogen.debug")) {
+  if (atom.config.get("Hydrogen.debug")) {
```
lgeiger: This will require Atom/Hydrogen to be restarted, but for now this should work 👍

Issue #870, opened 2017-06-13
Duplicate kernel startup


BenRussert:
**tl;dr**: it is possible to start up multiple kernels unintentionally while the first kernel is still getting ready

---

When a line of code is run [`createResultBubble`](https://github.com/nteract/hydrogen/blob/master/lib/main.js#L299) is called which checks to see if there is a running kernel and starts one if not (comments mine):

```js
createResultBubble(code: string, row: number, editor: atom$TextEditor) {
    if (!store.grammar) return;

    if (store.kernel) {
      // a kernel is up and running, so we can send it code to run immediately
      this._createResultBubble(store.kernel, code, row);
      return;
    }

    // no kernel is running so we need to start one up
    kernelManager.startKernelFor(store.grammar, editor, (kernel: ZMQKernel) => {
      // once the kernel is ready this callback will send it code to run
      this._createResultBubble(kernel, code, row);
    });

```

In most cases this works fine, since most kernels start up pretty quickly and the user will often wait to see output before calling more code to run.

If, however, `createResultBubble` is called two or more times in quick succession the first call must race to complete before the 2nd+ call checks for a running kernel. If the first call doesnt make it in time a second `startKernelFor` process begins.

In this gif I am pressing `shift-enter` to run and move down three times in quick succession, you can see that three different kernels begin startup:

![starts-a-bunch-o-kernels mp4](https://user-images.githubusercontent.com/10860657/27062702-6012f386-4fb2-11e7-90c2-b4f85bc9edce.gif)

BenRussert:
As a fix would it make sense to cue commands to run then have the callback run everything in the cue? 

In most cases this would just be the code from the initial command, but if multiple inputs were entered the kernel startup process would just start once and add code to the cue if a kernel startup process had begun.

BenRussert:
I'll make a PR later for discussion. I've got a simple fix that just prevents duplicate kernels from starting while the first is still loading up.

Issue #858, opened 2017-06-05
Feedback on deprecating of connection.json workflow


lgeiger:
We are looking into deprecating connecting existing kernels via a `connection.json` file.

For some time now Hydrogen supports using [kernel gateways](https://nteract.gitbooks.io/hydrogen/docs/Usage/RemoteKernelConnection.html) for connection to existing kernels. Using that option is a lot simpler yet very powerful.

If you depend on the old workflow please share your use case and why you cannot achieve this easily with remote kernels.

Issue #780, opened 2017-05-17
Jupyter client invalid signature error when running code block


paanvaannd:
**Description:**

Trying to run code block results in the following error:

<img width="460" alt="hydrogen_error" src="https://cloud.githubusercontent.com/assets/16567134/26133646/b4e9e514-3a76-11e7-94fd-50eefc02bb19.png">

It seems that Python 3 is giving the error message (going by the title of the error message popup) yet the error points to Python 2.7 as seen in the information of that message. I'm new to Jupyter and Hydrogen so perhaps I'm missing something obvious. Any help would be much appreciated.

**Steps to Reproduce:**

1. Download IPython through `pip3 install ipython`
2. Download Jupyter through `pip3 install jupyter`
3. Download Hydrogen to Atom
4. Create a Python code file and try to run a block of code (my code is basically `import pandas as pd` and `dataset = pd.read_csv('NAME_OF_FILE')`)
5. Error window picture above shows up

**Versions:**

macOS 10.12.5 (Sierra)
Atom version 1.16.0
Hydrogen 1.13.0

**Logs:**

The following is a direct copy/paste of from the log but it doesn't seem to have been formatted properly, so I include a picture of the relevant portion of the logs after for a cleaner look:

```
utils.js [sm]:75 Hydrogen: clearBubblesOnRow: 13log @ utils.js [sm]:75clearBubblesOnRow @ main.js [sm]:374_createResultBubble @ main.js [sm]:303createResultBubble @ main.js [sm]:285run @ main.js [sm]:399hydrogenRun @ main.js [sm]:76module.exports.CommandRegistry.handleCommandEvent @ command-registry.coffee:255(anonymous function) @ command-registry.coffee:1module.exports.KeymapManager.dispatchCommandEvent @ keymap-manager.js:599module.exports.KeymapManager.handleKeyboardEvent @ keymap-manager.js:390module.exports.WindowEventHandler.handleDocumentKeyEvent @ window-event-handler.coffee:76(anonymous function) @ window-event-handler.coffee:1
[...]

PR #876, opened 2017-06-16
Prevent duplicate kernel startup

BenRussert:
Since we rely on `store.kernel` to indicate if a new kernel should be started `store` needs to be aware af startup pretty much immediately to avoid duplicate kernels. See #870 for a wordier explanation.

The idea here is to somehow fail silently instead of creating dup kernels that may not shut down properly. I would be grateful for feedback!

# To Duplicate: 
- start a new kernel by pressing `ctrl-enter` on 2 or more lines quickly

example

### DIFF
--- a/lib/store/index.js
+++ b/lib/store/index.js
@@ -10,7 +10,8 @@ import type Kernel from "./../kernel";
 
 class Store {
   subscriptions = new CompositeDisposable();
-  @observable runningKernels = new Map();
+  @observable startingKernels: Map<string, boolean> = new Map();
+  @observable runningKernels: Map<string, Kernel> = new Map();
   @observable editor = atom.workspace.getActiveTextEditor();
   @observable grammar: ?atom$Grammar;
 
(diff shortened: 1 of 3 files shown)


lib/store/index.js
```diff
@@ -10,7 +10,8 @@ import type Kernel from "./../kernel";
 
 class Store {
   subscriptions = new CompositeDisposable();
-  @observable runningKernels = new Map();
+  @observable startingKernels: Map<string, boolean> = new Map();
+  @observable runningKernels: Map<string, Kernel> = new Map();
   @observable editor = atom.workspace.getActiveTextEditor();
```
BenRussert: ideally we could do this without even adding `startingKernels`, but we would have to somehow give `store` the new kernel before the kernel is actually ready. I suppose we could promisify or something, but am eager to see what you all think.

BenRussert:
This is ready for review!

There haven't been any comments on this or #876, what do you think? 

@nikitakit I remembered that you had mentioned this issue in #780, do you have any feedback?

lib/store/index.js
lgeiger: I like the idea of having a promise or to just give `store` a new kernel even if it's not fully connected yet. But this likely requires some refactorings that will be a lot easier once #858 is closed.

BenRussert: I agree. I'll use the phrase "we can iterate" here 😄

rgbkrk:
:cool:

Issue #863, opened 2017-06-07
'Restart kernel' and 'shutdown kernel' commands failing


wowserx:
**Description:**

Updated 'restart kernel' command doesn't properly restart the kernel. Run commands hang until kernel is restarted a second time, but only while a run command is hanging. 

Second (probably related) problem: Shutdown Kernel command after Restart Kernel command triggers invalid message tracebacks. 

**Steps to Reproduce:**

The sequence:

1. Restart Kernel
2. Run All
3. Restart Kernel

will properly run the program, but the sequence:

1. Restart Kernel
2. Restart Kernel
2.5 (...)
3. Run All 

will hang. 

The second problem can be reproduced as follows:

1. Run All
2. Restart Kernel
3. Shutdown Kernel
4. Run All

Produces the error message below, under Logs. 

**Versions:**

OSX 10.12.5 (16F73)
Atom 1.17.2 x64
Hydrogen 1.16.0

(Electron: 1.3.15
Chrome  : 52.0.2743.82
Node    : 6.5.0
Jupyter : 4.1.0)

**Logs:**

[IPKernelApp] ERROR | Invalid Message Traceback (most recent call last): File "/Users/$USERNAME/anaconda/lib/python2.7/site-packages/ipykernel/kernelbase.py", line 199, in dispatch_shell msg = self.session.deserialize(msg, content=True, copy=False) File "/Users/$USERNAME/anaconda/lib/python2.7/site-packages/jupyter_client/session.py", line 859, in deserialize raise ValueError("Invalid Signature: %r" % signature) ValueError: Invalid Signature: 'add5a85b80d1ed8600e3188d577977d087430339bdc0e46cff9ca8971c3222cd'

t9md:
![restart-callback-unstable](https://user-images.githubusercontent.com/155205/26926052-f100dd1a-4c86-11e7-9211-624c3111a774.gif)

I think this issue is related recently introduced faster-restart(#853).
I noticed similar issue(not same, but greatly related) while I'm working on #861(which call `runAll` on `restart` callback).

- If I passed callback to  `restart` like `restart(runAll)`, the issue is more happens. But % is still rare.
- I can't see issue with ijavascript kernel , but I can see issue for iPython kernel.
- If I launch Atom in dev-mode and open dev-tools, the issue in iPython kernel is almost always happen(means each restart-and-run-all).
- This is because, hydrogen write lots of logs in dev-console which add extra delay to complete kernel-restart(my assumption).
[...]

Issue #591, opened 2017-01-21
Startup issues with ijavascript


lgeiger:
**Description:**

If I want to use IJavascript in Hydrogen almost everytime the kernel won't properly start and only display the spinner. After restarting the IJavascript kernel everything works as expected.

This may be related to #584, #468 and #88 but for me it only happens with IJavascript.

I guess this could be caused by a race condition where the sockets are not ready to receive the execute request.

If I change our [startup script](https://github.com/nteract/hydrogen/blob/master/lib/kernel-manager.js#L131-L144) to
```javascript
launchSpec(kernelSpec, spawnOptions)
    .then(({ config, connectionFile, spawn }) => {
      const kernel = new ZMQKernel(
        kernelSpec, grammar,
        config, connectionFile,
        spawn,
      );

      setTimeout(() => {
        this.setRunningKernelFor(grammar, kernel);

        this._executeStartupCode(kernel);

        if (onStarted) onStarted(kernel);
      }, 1000);
    });
```
the issue disappears.
Is it necessary to wait for the [sockets to be connected](https://github.com/nteract/hydrogen/blob/master/lib/zmq-kernel.js#L95-L98)?

@n-riesco @rgbkrk Any idea what else could cause this issue?

**Steps to Reproduce:**

1. Start fresh Atom instance
2. Run javascript file with IJavascript -> only see the spinner
3. Restart IJavascript kernel via statusbar -> everything works fine

<details>
<summary><strong>Logs and Versions:</strong></summary>
A failed kernel startup:
<img width="619" alt="bildschirmfoto 2017-01-21 um 23 49 32" src="https://cloud.githubusercontent.com/assets/13285808/22178361/58c8e862-e034-11e6-89cd-40cd14dc683f.png">

A successful kernel startup:
<img width="619" alt="bildschirmfoto 2017-01-21 um 23 51 36" src="https://cloud.githubusercontent.com/assets/13285808/22178371/a1ab795a-e034-11e6-9a2c-cf0749c6a87e.png">
<img width="619" alt="bildschirmfoto 2017-01-21 um 23 51 41" src="https://cloud.githubusercontent.com/assets/13285808/22178372/a1ac1928-e034-11e6-858e-38a4a9f3ee4e.png">

**Versions:**

 macOS 10.12.2
```bash
$ atom --version
Atom    : 1.14.0-beta2
Electron: 1.3.13
Chrome  : 52.0.2743.82
Node    : 6.5.0
```
</details>

[...]

PR #877, opened 2017-06-16
Restart: Cleanup kernel process

lgeiger:
Follow up on #853

This will ensure to kill the old kernel process before starting a new one.

For me this fixes the issue describe by @t9md: https://github.com/nteract/hydrogen/issues/863#issuecomment-307076817

Unfortunately I can't reproduce #863 so I don't know for sure if this will fix the original issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,3 +17,10 @@ env:
   matrix:
     - ATOM_CHANNEL=stable
     - ATOM_CHANNEL=beta
+
+addons:
+  apt:
+    sources:
+    - ubuntu-toolchain-r-test
+    packages:
+    - g++-6
--- a/lib/main.js
+++ b/lib/main.js
@@ -403,10 +403,7 @@ const Hydrogen = {
     });
     this.clearResultBubbles();
 
-    const runAll = () => this.runAll(breakpoints);
-
-    // FIXME: temporal workaround until hydrogen#863 fixed.
-    kernel.restart(() => setTimeout(runAll, 500));
+    kernel.restart(() => this.runAll(breakpoints));
   },
 
   clearBubblesOnRow(row: number) {
--- a/lib/zmq-kernel.js
+++ b/lib/zmq-kernel.js
@@ -169,6 +169,7 @@ export default class ZMQKernel extends Kernel {
   restart(onRestarted: ?Function) {
     if (this.kernelProcess) {
       this.shutdown(true);
+      this._kill();
       const { spawn } = launchSpecFromConnectionInfo(
         this.kernelSpec,
         this.connection,


wowserx:
#863 is still active.

lgeiger:
@wowserx Just to make sure, have you tested with the latest `master`? `v1.17.0` doesn't yet include this fix.

wowserx:
I feel silly, wasn't running atom in dev mode. 

#863 is no longer active! It works fine.

lgeiger:
Awesome thanks for checking! 🎉 

I'll ship a patch release since this fixes a critical issue.

t9md:
> For me this fixes the issue describe by @t9md: #863 (comment)

I still can reproduce with latest master v1.17.1.

My steps to reproduce is

1. prepare sample code below.

```javascript
let a = [1, 2, 3]
a.push(10)
a // %%
a.push(20)
a // %%
a.push(30)
a.map(e => e * e)
a // %%
```

2. execute `restart-kernel-and-re-evaluate-bubbles` via keymap so that I can execute multiple times with short gab between each execution.

I keymap is here,

```
'atom-text-editor':
  'cmd-h': 'hydrogen:restart-kernel-and-re-evaluate-bubbles'
```

3. If I execute ten times, I can see some instance where bubbles are not filled with value(checkmark which represent `undefined`).

restart-not-stable

wowserx:
I can reproduce the above, but it's not the same problem as #863, since the code is definitely running, just not consistently re-populating the bubbles. 

In #863, the process was hanging indefinitely, but in the issue described above I've had it work where only some of the bubbles populate, but their results are correct: 

screen shot 2017-06-17 at 08 52 52

I could also reproduce this problem by mashing a custom keymap which calls restart kernel and clear bubbles simultaneously, then quickly running the code.

BenRussert:
@t9md want to try that my fork on #876? I think this could be related to #870

t9md:
@BenRussert Sorry for late response(I didn't noticed this comment).
I tried with #876, and but no luck, I can still reproduce this the issue I described above.
How about you?
The step is very simple, "Invoking restart-and-re-evaluate via keymap multiple times for ijavascript kernel"

BenRussert:
@t9md I see your issue now. This is a weird one.

t9md:
So you could reproduce it right?

lgeiger:
Looks like we run into #591 with IJavascript

lgeiger:
@t9md Is this issue something that comes up in your normal workflow, or does it only happen when you rapidly run this command?

t9md:
Yes, I can say it's easily happens in normal workflow.
I believe, you and all other people can easily reproduce this issue, by

- invoke `restart-kernel-and-re-evaluate-bubbles` via keymap on ijavascript kernel.

This is NOT corner case of usage.
As my understanding, I didn't see this issue before hydrogen introduce rapid restart mechanism.
I prefer "slow but reliable" behavior than "quick but occasional un-reliable" behavior.
I want you and other maintainer try to reproduce this issue, and if it's really easy to reproduce as I explained, then my opinion is, we should revert to slow-but-reliable behavior worked in previous version.
I wonder if #591 is related(this is old issue), why this issue is not happen(I believe so) in previous slow restart then begin to appear from rapid restart?

lgeiger:
👍 I'll take a look at it later this week.

rgbkrk:
> we should revert to slow-but-reliable behavior worked in previous version

That sounds good to me.

PR #885, opened 2017-06-21
add hot-reload-package command to boost dev-workflow

t9md:
## Demo

In this demo, changing `icon-check` to different one **without reloading Atom**.
workflow is

1. change icon def
2. invoke `hydrogen:hot-reload-package` via keymap( I use `f8` ).
3. `hydrogen:run-all` to see changed icon
4. repeat till I could find better icon etc..

hot-reload

## What this PR do

- Add new command `hydrogen:hot-reload-package`.
- This command is only available when atom is launched in dev-mode.
- This command allow reload hydrogen package without reloading Atom itself.
  - This is faster, so out dev speed would get faster and more pleasant.

### special note

I changed `fs.unlink` used in `ZMQKernel::destroy`, to avoid following warning.

`internal/process/warning.js:21 (node:58539) DeprecationWarning: Calling an asynchronous function without callback is deprecated.`

- This warning was already there(NOT introduced by this PR, so I just fixed it).
- This is nodev-v7 specific error, and atom-beta use node-v7, you can reproduce it by simply disable package via setting-view.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -44,7 +44,8 @@ import {
   renderDevTools,
   INSPECTOR_URI,
   WATCHES_URI,
-  OUTPUT_AREA_URI
+  OUTPUT_AREA_URI,
+  hotReloadPackage
 } from "./utils";
 
 import type Kernel from "./kernel";
@@ -127,6 +128,14 @@ const Hydrogen = {
         "hydrogen:clear-results": () => this.clearResultBubbles()
       })
     );
+
+    if (atom.inDevMode()) {
+      store.subscriptions.add(
+        atom.commands.add("atom-workspace", {
+          "hydrogen:hot-reload-package": () => hotReloadPackage()
+        })
+      );
+    }
 
     store.subscriptions.add(
       atom.workspace.getCenter().observeActivePaneItem(item => {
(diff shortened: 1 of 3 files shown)


lib/main.js
```diff
@@ -202,6 +211,7 @@ const Hydrogen = {
 
   deactivate() {
     store.dispose();
+    this.clearResultBubbles();
```
lgeiger: Thanks for catching that 👍

lib/zmq-kernel.js
```diff
@@ -429,7 +429,7 @@ export default class ZMQKernel extends Kernel {
 
     if (this.kernelProcess) {
       this._kill();
-      fs.unlink(this.connectionFile);
+      fs.unlinkSync(this.connectionFile);
```
lgeiger: I'm wondering if this could potentially freeze the UI if `unlinkSync` takes longer than normal, since it is now synchronous?

I didn't know that calling async functions without a callback was deprecated.
@t9md Should we maybe just add a pseudo callback instead?

t9md: I think we unlinkSync is better, it's not so slow just removing one file.
When destroy() finished, I want assume all cleanup stuff done. 
IMO reliable/simple behavior is important for shutdown.
Then if performance is really matter, we then move to tune that part like destroy return promise.

lgeiger: 👍 Haven't thought about that.

PR #887, opened 2017-06-21
Ensure that all sockets are connected after kernel restart

lgeiger:
This will considerably slow down the kernel restart, but ensures that all sockets are connected after restarting. We still keep our old connection file though.

This fixes https://github.com/nteract/hydrogen/pull/877#issuecomment-309189724

### DIFF
--- a/lib/zmq-kernel.js
+++ b/lib/zmq-kernel.js
@@ -102,6 +102,10 @@ export default class ZMQKernel extends Kernel {
     this.ioSocket.on("message", this.onIOMessage.bind(this));
     this.stdinSocket.on("message", this.onStdinMessage.bind(this));
 
+    this.monitor(done);
+  }
+
+  monitor(done: ?Function) {
     try {
       let socketNames = ["shellSocket", "controlSocket", "ioSocket"];
 
@@ -168,6 +172,7 @@ export default class ZMQKernel extends Kernel {
 
   restart(onRestarted: ?Function) {
     if (this.kernelProcess) {
+      this.setExecutionState("restarting");
       this.shutdown(true);
       this._kill();
       const { spawn } = launchSpecFromConnectionInfo(
@@ -177,14 +182,14 @@ export default class ZMQKernel extends Kernel {
         this.options
       );
       this.kernelProcess = spawn;
-      this.setExecutionState("idle");
+      this.monitor(() => {
+        if (onRestarted) onRestarted(this);
+      });
+    } else {
+      log("ZMQKernel: restart ignored:", this);
+      atom.notifications.addWarning("Cannot restart this kernel");
       if (onRestarted) onRestarted(this);
-      return;
     }
-
-    log("ZMQKernel: restart ignored:", this);
-    atom.notifications.addWarning("Cannot restart this kernel");
-    if (onRestarted) onRestarted(this);
   }
 
   // onResults is a callback that may be called multiple times
--- a/lib/zmq-kernel.js
+++ b/lib/zmq-kernel.js
@@ -172,6 +172,9 @@ export default class ZMQKernel extends Kernel {
 
   restart(onRestarted: ?Function) {
     if (this.kernelProcess) {
+      if (this.executionState === "restarting") {
+        return;
+      }
       this.setExecutionState("restarting");
       this.shutdown(true);
       this._kill();


t9md:
I checked. And result is NOT good.

 First of all, are you able to reproduce issue with the steps I explained?

 This is super easy steps to reproduce issue(just keymap then hit several time on ijavascript kernel).

 If this is not easy, that mean my environment specific issue, so we might not need to FIX this issue.


Anyway, this PR changes how issue be appear.
Here is the gif when issue was happened.

tryit_js_ _tryit_ ___tryit


This means, `let a` is evaluated twice in same context.
This should not happen since command is `restart-and-re-evaluate`(not `might-restart-and-re-evaluate`).
I imagine something complex/confusing is happening like 2nd request goes to kernel just restarted and evaluated 1st request of `restart-and-re-evaluate`.

What I want is

- Reproduce issue then see if we can fix some additional changes.
- If not, just revet to previous implementation.
- If reverting means also revert other fix, but it should be OK, better to retry fix based on correctly working behavior.

lib/zmq-kernel.js
```diff
@@ -168,6 +172,7 @@ export default class ZMQKernel extends Kernel {
 
   restart(onRestarted: ?Function) {
```
t9md: In this `restart` entry point, we need to prevent simultaneous restart. like this
```javascript
    if (this.executionState === "restarting") {
      return
    }
```

since finish timing is monitor called `done`. but it happens async.

lib/zmq-kernel.js
```diff
@@ -177,14 +182,14 @@ export default class ZMQKernel extends Kernel {
         this.options
       );
       this.kernelProcess = spawn;
-      this.setExecutionState("idle");
+      this.monitor(() => {
+        if (onRestarted) onRestarted(this);
```
t9md: or if we don't want to depend on `this.executionState === 'restarting'` guard.
We can explicitly use independent flag like `this.restarting = true` at entry point of `restart`, then clear `this.restart = false` in monitor callback.

Either way prevent multiple restart request processing.

t9md:
> This means, `let a` is evaluated twice in same context.
> This should not happen since command is restart-and-re-evaluate(not might-restart-and-re-evaluate).

I investigated what is happening in this situation.
I added debug output to `restart` function and reproduced.

```javascript
  restart(onRestarted: ?Function) {
    // This guard is SUPER important to prevent simultaneous restart processing.

    // if (this.executionState === "restarting") {
    //   console.log('return ');
    //   return
    // }

    console.log('------------------');
    console.log('called restart');
    if (this.kernelProcess) {
      console.log('execution_state', this.executionState)
      this.setExecutionState("restarting");
      this.shutdown(true);
      this._kill();
      const { spawn } = launchSpecFromConnectionInfo(
        this.kernelSpec,
        this.connection,
        this.connectionFile,
        this.options
      );
      this.kernelProcess = spawn;
      this.monitor(() => {
        console.log('restarted');
        if (onRestarted) onRestarted(this);
      });
    } else {
      log("ZMQKernel: restart ignored:", this);
      atom.notifications.addWarning("Cannot restart this kernel");
      if (onRestarted) onRestarted(this);
    }
  }
```

This is output when "`let a` evaluated same context" issue happened.

developer_tools

What is happening here is "processing 2nd request before 1st restart yet not finished", so  "1st request's restart callback is evaluated in same environment of 2nd request".
To say simply both 1st and 2nd request is evaluated in same envorment where it should be separated.

We can prevent simple guard at restart entry point.

t9md:
@lgeiger  I checked, with last commit and it's OK. I'm not sure who merge this? not me right?
As reviewer, I can say it's OK, ready!

lgeiger:
Thanks a lot for investigating and the review!

I did test with both Python and Javascript and this PR fixes the issues. The only thing I didn't test was calling restart so fast that multiple restarts were run. In that case multiple execution requests are sent to the socket. Since we reuse the old connection config, the kernel then receives two execution calls and the second one fails with the error message above.

For me 475e6261b637989f2f65fc96f949f63efd499e10 fully fixes the issue described above.

I'm sorry for the many restart issues I introduced in my last PRs 😇 
If there are still issues with this we can change [this line](https://github.com/nteract/hydrogen/pull/887/files#diff-b3e285d757729d2cadc9b14f332b81deR181) to use `launchSpec` and call `this.connect` instead of `this.monitor`. This will effectively revert to the old restart mechanism without breaking the UI.

lgeiger:
> I checked, with last commit and it's OK. I'm not sure who merge this? not me right?
As reviewer, I can say it's OK, ready!

Awesome that's great! 🎉 
Don't hesitate to go ahead merge this PR. As a maintainer and member of the nteract org you have the power 👍

t9md:
WOW I'm happy, with this fix, the workflow I wanted for hydrogen finally perfectly work!

1. preset mark by toggle-bubble
2. then evaluate from top to bottom with cleanly started kernel environment by `restart-and-re-evaluate`

Yes no longer strange multi restart issue(I hope so this time again!!)

lgeiger:
🆒  We've fixed quite a few bugs since our last release! We should ship a new minor version soon.

PR #883, opened 2017-06-19
Correctly reflect kernel restart in UI

lgeiger:
A few more kernel restart fixes:
- remote kernels weren't properly restarted: f099f64
- restart wasn't properly reflected in the UI since we don't remove the result bubbles. Fixes #880

## PR
pr
## master
master

PR #888, opened 2017-06-22
Refactor markers handling

lgeiger:
In an attempt at fixing https://github.com/atom/atom/issues/14836 and https://github.com/atom/atom/pull/14844 in Hydrogen using Reacts `onDidUpdate` handler, some refactors where necessary.

This moves all the marker handling to `store/markers.js` and adds tests for it.
All storage is now handled by `store/` which means we can move functionality out of `main.js` into separate files. This makes it possible to better test the codebase.

### DIFF
--- a/lib/components/result-view/index.js
+++ b/lib/components/result-view/index.js
@@ -13,6 +13,7 @@ import type Kernel from "./../../kernel";
 export default class ResultView {
   disposer: atom$CompositeDisposable;
   marker: atom$Marker;
+  outputStore: OutputStore;
 
   destroy = () => {
     this.disposer.dispose();
@@ -33,16 +34,13 @@ export default class ResultView {
 
     markerStore.clearOnRow(row);
 
-    const buffer = editor.getBuffer();
-    const lineLength = buffer.lineLengthForRow(row);
+    this.marker = editor.markBufferPosition([row, Infinity], {
+      invalidate: "touch"
+    });
 
-    const point = new Point(row, lineLength);
-    this.marker = editor.markBufferPosition(point, { invalidate: "touch" });
-    const lineHeight = editor.getLineHeightInPixels();
-
-    const outputStore = new OutputStore();
-    outputStore.updatePosition({
-      lineLength: lineLength,
+    this.outputStore = new OutputStore();
+    this.outputStore.updatePosition({
+      lineLength: editor.getBuffer().lineLengthForRow(row),
       lineHeight: editor.getLineHeightInPixels(),
       editorWidth: editor.getEditorWidthInChars()
     });
@@ -57,7 +55,7 @@ export default class ResultView {
       if (!event.isValid) {
         markerStore.delete(this.marker.id);
       } else {
-        outputStore.updatePosition({
+        this.outputStore.updatePosition({
           lineLength: this.marker.getStartBufferPosition().column
         });
       }
@@ -67,7 +65,7 @@ export default class ResultView {
 
     reactFactory(
       <ResultViewComponent
-        store={outputStore}
+        store={this.outputStore}
         kernel={kernel}
         destroy={this.destroy}
         showResult={showResult}
@@ -76,7 +74,5 @@ export default class ResultView {
       null,
       this.disposer
     );
-
-    return outputStore;
   }
 }
--- a/lib/components/result-view/index.js
+++ b/lib/components/result-view/index.js
@@ -44,7 +44,6 @@ export default class ResultView {
     outputStore.updatePosition({
       lineLength: lineLength,
       lineHeight: editor.getLineHeightInPixels(),
-      // $FlowFixMe: Missing flow type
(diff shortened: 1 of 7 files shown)


lib/components/result-view/index.js
```diff
@@ -1,46 +1,83 @@
...
+    const point = new Point(row, lineLength);
+    this.marker = editor.markBufferPosition(point, { invalidate: "touch" });
+    const lineHeight = editor.getLineHeightInPixels();
+
+    const outputStore = new OutputStore();
+    outputStore.updatePosition({
+      lineLength: lineLength,
+      lineHeight: editor.getLineHeightInPixels(),
+      // $FlowFixMe: Missing flow type
+      editorWidth: editor.getEditorWidthInChars()
```
BenRussert: btw, should we add this undocumented method to our flow types?

t9md:
> I would merge this now, but thought @t9md may want to weigh in first with all his work involving markers. I do not think this will change any functionality though.

After 5min of behavioral check, I found two breaking.
- `toggle-bubble` cannot add bubble on the editor which is not evaluated by kernel.
  - Should be able to add bubble before any code evaluation, but no longer with this PR.
- Deactivating package call un-exisiting clearBubble function throw error.

t9md:
TIPS(Self promotion): Tool to help refactoring.

I created [atom-narrow](https://atom.io/packages/narrow) package, which allow me interactive project-wide search by `ag` or `rg` from Atom editor.

I keymap `enter` key to `narrow:search-by-current-word` in `vim-mode-plus`'s normal-mode.

So when I rename or remove some variable/function used in project.
My workflow is like this.

1. Decided to remove `clearResultBubbles`.
2. Hit `enter` on `clearResultBubbles` it fire `narrow:search-by-current-word`
3. Then I know where this function is used within project.
4. While I removing, the result list is auto-updated, so I can know the remaining call to `clearResultBubbles` which I have to make it emptied. It's kind of TODO list to this removing-function-task.

This workflow greatly reduce change of overlook, time of investigation.
Happy if you guys try and find it useful.

lgeiger:
@t9md Good catch!

I messed up the rebase onto #885 and #875. I'll try `narrow` next time 😇 

> toggle-bubble cannot add bubble on the editor which is not yet evaluated by kernel.

I mistakenly thought that this is expected behavior since `insertResultBubble` is called with a wrong 
signature [here](https://github.com/nteract/hydrogen/pull/888/files#diff-16b7d75b70953cbf4755508170f2f24cL440), I don't know why that worked before. Will fix that!

lib/components/result-view/index.js
```diff
@@ -1,46 +1,82 @@
...
 
     this.disposer = new CompositeDisposable();
 
+    markerStore.clearOnRow(row);
+
+    const buffer = editor.getBuffer();
+    const lineLength = buffer.lineLengthForRow(row);
+
+    const point = new Point(row, lineLength);
+    this.marker = editor.markBufferPosition(point, { invalidate: "touch" });
```
t9md: if you want to add marker at EOL. you can write like this. Since point are **clipped** to actual EOF.

```javascript
 this.marker = editor.markBufferPosition([row, Infinity], { invalidate: "touch" });
```

lib/components/result-view/index.js
```diff
@@ -1,46 +1,82 @@
...
         showResult={showResult}
       />,
-      this.element,
+      element,
       null,
       this.disposer
     );
+
+    return outputStore;
   }
```
t9md: I want expect new return it's instance(instance of `ResultView`). 
If caller want to access outputStore, I like this even if it's verbose.( so need to keep outputStore as prop of `ResultView`)

```javascript
const resultView = new ResultView(...)
const {outputStore} = resultView
```

t9md:
> I mistakenly thought that this is expected behavior since insertResultBubble is called with a wrong
> signature here, I don't know why that worked before. Will fix that!
Oh, I didn't noticed my mistake, thanks. Strangely worked as expected 😅

lib/components/result-view/index.js
lgeiger: 👍  That's a lot nicer! Fixed it.

lib/components/result-view/index.js
lgeiger: Thanks, didn't know that

lgeiger:
Thanks a lot for the review 👍 

> Oh, I didn't noticed my mistake, thanks. Strangely worked as expected 😅

No problem, I changed the signature in #883 so that's why it was broken 😇

lgeiger:
This should be good to go now.

t9md:
> This should be good to go now.

Agree.
I'm not sure I'm OK to merge it. (I'm new to hydrogen community, Any consensus for merge procedure?)

BenRussert:
@t9md sounds like you approve and we have given enough time for anyone to chime in with issues! I'll merge this, but you are as welcome as I am to merge reviewed/approved PR's!

Issue #862, opened 2017-06-07
Request: strip _common_ leading whitespace from input before passing to kernel


sglyon:
I'd _love_ it if hydrogen stripped leading whitespace that was common to all lines in the selection before passing to the kernel.

For example, suppose I have this python code:

```python
def build_grids(p: Params):
    # build grid over I choices and N states
    Imax = 10
    expected_life = (1-p.δ) / p.δ
    Igrid = np.arange(1, Imax+1)
    Ngrid = np.arange(0, round(Imax*expected_life)+1, 2)

    zgrid = p.z
    # ... more code ....
```

If I were to highlight from the start of the commented line `    # build grid...` (including the leading whitespace) all the way to `p.z`, and try to evaluate I would get a complaint from the python kernel about unexpected whitespace. 

If I were to take that same highlighted region and copy it to my system clipboard and then use the `%paste` magic in an ipython terminal session, the leading whitespace would be stripped and evaluation would proceed without errors.

Note that in my example if I were to start my selection at the `I` character in `Imax` then the behavior should be to still throw an error because the first line in my selection doesn't have leading whitespace in common with all other lines.

I was going to try to take a stab at this, but I don't know where to look inside here. I figured you would know better than me ;)

BenRussert:
My first impression is that there isn't a truly language agnostic way to do this from hydrogen's end. 

Interesting issue, I think an init script or plugin solution could be fairly straight forward, but I'm not sure this should be built into hydrogen itself.

rgbkrk:
Is there a precedent for this with the jupyter notebook?

sglyon:
No, the notebook doesn't do this, but the console does... with the `%paste` magic:

<img width="1284" alt="screen shot 2017-06-12 at 10 50 29 pm" src="https://user-images.githubusercontent.com/1001948/27063766-ac0d5b6e-4fc1-11e7-9a5c-da576a2d792a.png">

```
~|⇒ ipython                                                                                       vpn-192-168-100-7
Python 3.6.1 |Anaconda 4.4.0 (x86_64)| (default, May 11 2017, 13:04:09)
Type "copyright", "credits" or "license" for more information.

IPython 5.3.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]:
   ...:     def foo():
   ...:         return "hello!"
   ...:
[...]

PR #891, opened 2017-06-23
Update Plugin API

nikitakit:
This is a first step towards making the plugin API more generally useful.

The sample plugin I have in mind is to address #862. The idea is to fix indentation when python code is run: if all lines in the code being run start with identical whitespace, the plugin should strip off that whitespace and run the unindented code instead. As a first step I've created a [plugin](https://github.com/nikitakit/hydrogen-python) that simply prepends a print statement to all python code run (I haven't written the actual whitespace transformation yet because I spent most of today on these plugin API updates).

To implement this plugin the following new features need to be added to the plugin API:
* Detecting when code is about to be executed
* Modifying code to be executed (I have decided to break this up into cancelling the original execution event, followed by issuing a new one from the plugin itself)
* Detecting when a new kernel is created (because listeners for kernel execution should only be attached once)
* Knowing the language of a running kernel

The original plugin implementation relied on atom's Emitters to trigger events for plugin use. Implementing `onDidChangeKernel` and future plugin updates would have required expanding Emitter usage -- but by now hydrogen is already using mobx. So along the way I decided to get rid of one usage of Emitter in favor of hooking into mobx.

Note that I'm not familiar with all of the fancy tools like prettier and flow that have been added to the project, so no idea if this code will pass lint.

### DIFF
--- a/lib/kernel.js
+++ b/lib/kernel.js
@@ -90,12 +90,32 @@ export default class Kernel {
     throw new Error("Kernel: restart method not implemented");
   }
 
-  execute(code: string, onResults: Function) {
-    throw new Error("Kernel: execute method not implemented");
+  _execute(code: string, callWatches: boolean, onResults: ?Function) {
+    throw new Error("Kernel: _execute method not implemented");
   }
 
-  executeWatch(code: string, onResults: Function) {
-    throw new Error("Kernel: executeWatch method not implemented");
+  executeMaybeWatch(code: string, callWatches: boolean, onResults: ?Function) {
+    let cancelled = false;
+    this.emitter.emit("will-execute", {
+      code: code,
+      callWatches: callWatches,
+      onResults: onResults,
+      cancel: () => {
+        cancelled = true;
+      }
+    });
+
+    if (!cancelled) {
+      this._execute(code, callWatches, onResults);
+    }
+  }
+
+  execute(code: string, onResults: ?Function) {
+    this.executeMaybeWatch(code, true, onResults);
+  }
+
+  executeWatch(code: string, onResults: ?Function) {
+    this.executeMaybeWatch(code, false, onResults);
   }
 
   complete(code: string, onResults: Function) {
--- a/lib/main.js
+++ b/lib/main.js
@@ -1,12 +1,6 @@
 /* @flow */
 
-import {
-  Emitter,
-  CompositeDisposable,
-  Disposable,
-  Point,
-  TextEditor
-} from "atom";
+import { CompositeDisposable, Disposable, Point, TextEditor } from "atom";
 
 import _ from "lodash";
 import { autorun } from "mobx";
@@ -56,8 +50,6 @@ const Hydrogen = {
   markerBubbleMap: null,
 
   activate() {
-    this.emitter = new Emitter();
-
     this.markerBubbleMap = {};
 
     let skipLanguageMappingsChange = false;
@@ -204,10 +196,6 @@ const Hydrogen = {
     );
 
     renderDevTools();
-
-    autorun(() => {
-      this.emitter.emit("did-change-kernel", store.kernel);
-    });
   },
 
   deactivate() {
--- a/lib/plugin-api/hydrogen-kernel.js
+++ b/lib/plugin-api/hydrogen-kernel.js
@@ -28,12 +28,56 @@ export default class HydrogenKernel {
   }
 
   /*
+   * The language of the kernel, as specified in its kernelspec
+   */
+  get language(): string {
+    return this._kernel.language;
+  }
+
+  /*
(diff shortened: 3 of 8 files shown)


lib/kernel.js
```diff
@@ -90,12 +90,32 @@ export default class Kernel {
...
+    this.emitter.emit("will-execute", {
+      code: code,
+      callWatches: callWatches,
+      onResults: onResults,
+      cancel: () => {
+        cancelled = true;
+      }
+    });
+
+    if (!cancelled) {
```
nikitakit: Note that this code is incorrect if there are multiple listeners: as soon as one cancels the execution no more listeners down the chain should be run. The Emitter class doesn't provide this kind of event propagation behavior, so I'll need to figure out some alternative.

t9md:
My reviewer assignment is just a mistake. reverted. sorry

n-riesco:
I think this PR is mostly driven by the needs to implement #862 .

I still think the plugin API would be more flexible if plugins had direct access to kernel instances (instead of the current `store.kernel.getPluginWrapper()`).

For the same reason, I also think that it'd be more flexible if plugins could hijack a kernel. That is, I think a plugin listening to `onDidCreateKernel` and `onDidChangeKernel` (as they currently do) should be able to replace the kernel instance. In this way a plugin could wrap the original kernel instance and overwrite only the methods they want to customise (`execute` in the case of #862).

nikitakit:
@n-riesco I actually hadn't thought of using object instance replacement for kernel hijacking behavior. Definitely worth considering.

Though I think it's unfair to characterize the approach here as not being general. Obviously this PR only has the pieces needed to implement #862, but that's mostly to keep the diff small/reviewable while still demonstrating the overall approach. The more general idea is to allow plugins to listen and react to various events being emitted by hydrogen, while possibly stopping event propagation to execute a custom codepath instead. The inspiration here is how event handling is done in the browser. Certainly this model has different tradeoffs to one where each plugin wraps the kernel and modifies some of its behavior.

One thing I've been thinking about over the course of this PR is how to make sure plugins interact sanely with each other (I haven't really come to a good conclusion on this yet). If multiple plugins want to go about replacing kernel behavior, they should be able to compose with each other.

For reference, here are the plugin ideas I had in mind while working on this PR:
* Implement #862 
* Implement watches using a plugin
* Implement a feature like watches, except have the watched code run *just before* user-requested code instead of *just after*
* Remap executing `expr?` to opening the inspector for `expr`, for those of us with lots of muscle memory from using notebooks

(I'm not saying all of these plugins would be all that useful for real work, but it would be disappointing if the plugin API couldn't accommodate each of these being implemented by a separate and independent plugin)

n-riesco:
> Though I think it's unfair to characterize the approach here as not being general. Obviously this PR only has the pieces needed to implement #862, but that's mostly to keep the diff small/reviewable while still demonstrating the overall approach.

I understand.

> The more general idea is to allow plugins to listen and react to various events being emitted by hydrogen, while possibly stopping event propagation to execute a custom codepath instead.

I'm not clear about this "stopping event propagation":

- if you mean to have something like `before_event`, `event` and `after_event` (and give a chance to the listeners of `before_event` to cancel `event`), then this sounds ok  to me.

- if you have in mind a hierarchy of plugins, then it sounds too complicated to me. Could you  think of an example that would require a hierarchy of plugins?

> If multiple plugins want to go about replacing kernel behavior, they should be able to compose with each other.

This sounds ok to me, Could think of an example where this approach wouldn't work?

> (I'm not saying all of these plugins would be all that useful for real work, but it would be disappointing if the plugin API couldn't accommodate each of these being implemented by a separate and independent plugin)

I think moving watches into a plugin would simplify Hydroden's code.

rgbkrk:
On a side / separate note -- I now use [`yapf`](https://github.com/google/yapf) on my Python code which obviates the need (for me) for Hydrogen to fix the indentation of my Python (part of what this PR is about).

On to this PR and status -- would you want to continue this on and/or keep it open @nikitakit?

qubicql:
@nikitakit @rgbkrk 
I would highly appreciate an implementation to the whitespace issue #862 !
Its not about having false indentations but about testing parts of a loop/if-statement etc.

qubicql:
@nikitakit @rgbkrk 
I would highly appreciate an implementation to the whitespace issue #862 !
Its not about having false indentations but about testing parts of a loop/if-statement etc.

nikitakit:
An update on the status of this PR: I have a rewrite that accomplishes the same goals but exposes a different API to plugin authors. I'm going to wait for my other PR to make it through code review, and then publish the rewritten plugin API changes here.

nikitakit:
Due to some repo cleanup I want to do, I'm closing this PR. I have another attempt at plugin API updates on my local machine, and I'll try to open up a new PR with that soon.

PR #893, opened 2017-06-23
Display empty message in output view if no kernel running

BenRussert:
I prefer the right dock to stay open even if I switch to a file w/o a kernel running. In this PR we just display an empty message if no kernel so you can still keep your dock open for other purposes e.g. git integration.

I suppose this is a matter of preference, what do you think?

keep-dock-open mp4

### DIFF
--- a/lib/components/output-area.js
+++ b/lib/components/output-area.js
@@ -9,10 +9,19 @@ import { OUTPUT_AREA_URI } from "./../utils";
 
 import typeof store from "../store";
 
+function EmptyMessage() {
+  return (
+    <div>
+      <ul className="background-message centered">
+        <li>No output to display</li>
+      </ul>
+    </div>
+  );
+}
+
 const OutputArea = observer(({ store: { kernel } }: { store: store }) => {
   if (!kernel) {
-    atom.workspace.hide(OUTPUT_AREA_URI);
-    return null;
+    return <EmptyMessage />;
   }
   return (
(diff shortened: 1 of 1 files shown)


lib/components/output-area.js
```diff
@@ -9,10 +9,19 @@ import { OUTPUT_AREA_URI } from "./../utils";
 
 import typeof store from "../store";
 
+function EmptyMessage() {
+  return (
+    <div>
+      <ul className="background-message centered">
+        <li>No output to display</li>
+      </ul>
+    </div>
```
lgeiger: Do we really need to wrap it in a `div`, wouldn't returning `ul` be enough?

BenRussert: I assumed so because the same no output message below is wrapped in a div. I suppose it is unnecessary here now that I think about it.

BenRussert:
I see your point. My view is that if I open my dock it should stay open unless I close it. 

Could make it configurable if that's better?

lib/components/output-area.js
BenRussert: I'm also going to flip this to `const EmptyMessage = () => ...`

BenRussert:
@lgeiger I've made this configurable, let me know what you think.

The way I think this should work (as in this PR) is like the git/github package works when switching between files in a repo or not. The dock will stay open either way:

github-docks mp4

PR #900, opened 2017-06-30
Only focus output area if necessary

lgeiger:
This prevents unnecessary focusing of the external output area.

Previously when receiving results in rapid succession (for example with a progress bar) using the command palette wasn't possibly since it was unfocused once new outputs arrive.

We could make the focusing behavior customizable with the same setting introduced by @BenRussert in #893.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -346,8 +346,17 @@ const Hydrogen = {
       if (globalOutputStore) {
         globalOutputStore.appendOutput(result);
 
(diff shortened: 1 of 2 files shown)


lgeiger:
This keeps the command palette usable when executing code similar to
```python
for i in range(1000):
    time.sleep(0.05)
    sys.stdout.write('\rProgress: {}%'.format(i))
```

PR #911, opened 2017-07-08
Add flow to ws-kernel-picker

BenRussert:
There are a couple of todos here, I'll call out what I mean in a couple comments.

### DIFF
--- a/lib/ws-kernel-picker.js
+++ b/lib/ws-kernel-picker.js
@@ -83,9 +83,8 @@ export default class WSKernelPicker {
       return;
     }
     const path = store.editor ? store.editor.getPath() : null;
-    if (!path) return;
 
-    this._path = `${path}-${v4()}`;
+    this._path = `${path || "unsaved"}-${v4()}`;
     const gatewayListing = new CustomListView(
       "No gateways available",
       this.onGateway.bind(this)
--- a/lib/ws-kernel-picker.js
+++ b/lib/ws-kernel-picker.js
@@ -1,4 +1,4 @@
-"use babel";
+/* @flow */
 
 import SelectListView from "atom-select-list";
 import _ from "lodash";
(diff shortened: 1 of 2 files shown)


lib/ws-kernel-picker.js
```diff
@@ -56,23 +63,29 @@ class CustomListView {
...
+    const gateways = Config.getJson("gateways") || [];
     if (_.isEmpty(gateways)) {
       atom.notifications.addError("No remote kernel gateways available", {
         description:
           "Use the Hydrogen package settings to specify the list of remote servers. Hydrogen can use remote kernels on either a Jupyter Kernel Gateway or Jupyter notebook server."
       });
       return;
     }
+    const path = store.editor ? store.editor.getPath() : null;
+    if (!path) return;
```
BenRussert: What should be the expected behavior when running `hydrogen:connect-to-remote-kernel` with an unsaved editor? 

I figure we would use the project path? An empty string? For now just returning to avoid a flow error.

lib/ws-kernel-picker.js
```diff
@@ -84,7 +97,7 @@ export default class WSKernelPicker {
     });
   }
 
-  onGateway(gatewayInfo) {
+  onGateway(gatewayInfo: any) {
```
BenRussert: I wasn't familiar enough with the intended objects here and below to type them. Took a couple shots, but decided not to get too caught up. 

Open to input or we could just keep these as a todo.

rgbkrk:
Still WIP or good to go?

BenRussert:
I think it's fine, just a little lazy on the use of `any` I think we could improve that later though

rgbkrk:
Yeah, it's better than nothing for now. We can always improve it later. This leaves room for the next person to come along and make it better. :smile:

lib/ws-kernel-picker.js
```diff
@@ -56,23 +63,28 @@ class CustomListView {
...
+  _kernelSpecFilter: (kernelSpec: Kernelspec) => boolean;
+  _path: string;
+  previouslyFocusedElement: ?HTMLElement;
+
+  constructor(onChosen: (kernel: Kernel) => void) {
     this._onChosen = onChosen;
   }
 
-  toggle(_grammar, _kernelSpecFilter) {
-    this._grammar = _grammar;
```
rgbkrk: Is it ok this behavior has changed?

rgbkrk: Ah right, comes out of `store.grammar` now. Consistency!

PR #918, opened 2017-07-09
Fix typo

BenRussert:
Just a quick fix, nothing to see here. Actually, we should probably start talking about when we will remove all the deprecated kernel connection code from the repo.


rgbkrk:
Remove it after the next release? There's a warning message at the moment right?

BenRussert:
Yep, we added the warning message in 1.17 (released 3 wks ago)

lgeiger:
>  Actually, we should probably start talking about when we will remove all the deprecated kernel connection code from the repo.

@BenRussert I'm happy to make a PR. I think I have a local branch somewhere that does the cleanup 😄

PR #931, opened 2017-07-27
Hide open-in-editor, use ctrl-click instead

### LINKED ISSUE #797: Hide open in editor button
I often find the ["open in editor"](https://github.com/nteract/hydrogen/blob/master/lib/components/result-view/result-view.js#L122) button isn't useful enough to have this prominent placement.

What do think about moving the functionality to `cmd` clicking the "copy to clipboard" button? We could change the icon on hover if `cmd/ctrl` is pressed.

BenRussert:
See #797. 

I could not figure out how to trigger this with `onkeydown` if I was already hovering and just press `ctrl`. So you have to hold ctrl while you mouseover to trigger.

ctrl-click-open mp4

It works pretty well for me other than that. Hoping to get some feedback!

Closes #797

### DIFF
--- a/lib/components/result-view/result-view.js
+++ b/lib/components/result-view/result-view.js
@@ -3,7 +3,7 @@
 import { CompositeDisposable } from "atom";
 import React from "react";
 import { observer } from "mobx-react";
-import { action, observable, toJS } from "mobx";
+import { action, observable, toJS, computed } from "mobx";
 import Display, {
   DEFAULT_SCROLL_HEIGHT
 } from "@nteract/display-area/lib/display";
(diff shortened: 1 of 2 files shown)


lib/components/result-view/result-view.js
```diff
@@ -28,12 +28,31 @@ class ResultViewComponent extends React.Component {
...
   expanded: IObservableValue<boolean> = observable(false);
+  ctrlHovering: IObservableValue<boolean> = observable(false);
 
   getAllText = () => {
     if (!this.el) return "";
     return this.el.innerText ? this.el.innerText.trim() : "";
   };
 
+  handleClick = (event: MouseEvent) => {
+    if (this.ctrlHovering.get()) {
```
lgeiger: You can check if the `ctrl` key is pressed with `event.ctrlKey` directly. See https://facebook.github.io/react/docs/events.html#mouse-events

BenRussert:
@lgeiger I agree, most of the code here is just to change the icon! `e.ctrlKey` will work with far fewer lines of code.

I'll update this next chance I get. I had fun making the icon change, but i'd much rather merge with the simpler ctrl click that works without mouseleave/enter.

lib/components/result-view/result-view.js
BenRussert: Yep, that's how my alt PR will work.

The reason for the observable was to keep the icon class in sync with whatever result would occur when clicking ( I couldnt get the icon to respond to ctrl keydown alone). It is hard to explain, and maybe not worth the trouble :laughing:

lgeiger:
Thanks a lot for the work! We can revisit this in the future.

Closing in favor of #941 for now.

PR #941, opened 2017-08-11
Hide open in editor (simple version)

BenRussert:
This is much cleaner if we are not worried about the icon switching.

I am not crazy about the way the tooltip message looks, open to suggestions on that (or anything else)

simple-ctrl-click mp4

### DIFF
--- a/lib/components/result-view/result-view.js
+++ b/lib/components/result-view/result-view.js
@@ -56,8 +56,9 @@ class ResultViewComponent extends React.Component {
     comp.add(
       atom.tooltips.add(element, {
         title: this.props.store.executionCount
-          ? `Copy to clipboard (Out[${this.props.store.executionCount}])`
-          : "Copy to clipboard"
+          ? `Click to copy, ctrl/cmd click to open in editor (Out[${this.props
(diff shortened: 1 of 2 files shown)


lib/components/result-view/result-view.js
```diff
@@ -48,8 +56,9 @@ class ResultViewComponent extends React.Component {
     comp.add(
       atom.tooltips.add(element, {
         title: this.props.store.executionCount
-          ? `Copy to clipboard (Out[${this.props.store.executionCount}])`
-          : "Copy to clipboard"
+          ? `Click to copy, ctrl/cmd click to open in editor (Out[${this.props
```
BenRussert: That's a great idea. I'll add that later today.

lgeiger: Cool thanks!

BenRussert:
@lgeiger I haven't actually tried this on mac, but pretty sure it will work.

lgeiger:
> @lgeiger I haven't actually tried this on mac, but pretty sure it will work.

Works excellent! Thanks.

PR #942, opened 2017-08-13
Ctrl-click open in editor for inline results

BenRussert:
This probably should have been part of #941, it just uses `handleClick` so that the tooltip text makes sense and you can copy or export in editor.


BenRussert:
> Want to ship a new release :package: :rocket: ?

Good idea, i'll ship one later tonight

lgeiger:
@BenRussert Thanks!

Issue #845, opened 2017-06-02
Unable to copy from the External output area


tobiasbj:
**Description:**
I like the new External output area feature, but it isn't currently possible to copy text from it. Making that possible would be a sweet addon :)

**Version**
I assume this is a global issue, but if not I am using:
Windows 10.
Atom 1.17.2
Hydrogen 1.15.0

lgeiger:
@tobiasbj Would you be interested in implementing this feature?

It can be achived similar to how we're doing it in the normal [result view](https://github.com/nteract/hydrogen/blob/master/lib/components/result-view/result-view.js#L24).

tobiasbj:
could be fun to try, but I don't think I have time to learn and do it right now. (I'm bascially just leeching off your glorious hydrogen :P ).

callmeish:
Adressed in #954. I'm a bit new to this github thing, let me know if anything in that looks weird or needs changes.

BenRussert:
@callmeish thanks for the PR! I'm looking forward to reviewing this weekend. Don't mean to leave you hanging.

PR #954, opened 2017-08-24
Added copy to clipboard button to external output area

callmeish:
Addressed the feature request in #845. Testing using python scripts gives identical functionality to hydrogen without these changes, the only difference being the added copy to clipboard button.

### DIFF
--- a/lib/components/output-area.js
+++ b/lib/components/output-area.js
@@ -4,7 +4,6 @@ import { CompositeDisposable } from "atom";
 import React from "react";
 import { observer } from "mobx-react";
 import { toJS } from "mobx";
-import _ from "lodash";
 
 import History from "./result-view/history";
 import { OUTPUT_AREA_URI } from "./../utils";
@@ -17,11 +16,6 @@ const EmptyMessage = () => {
       <li>No output to display</li>
     </ul>
   );
-};
-
-const buttonStyle = {
-  flex: "0 0 auto",
-  width: "fit-content"
 };
 
 const OutputArea = observer(({ store: { kernel } }: { store: store }) => {
@@ -39,11 +33,13 @@ const OutputArea = observer(({ store: { kernel } }: { store: store }) => {
     //convert output to a plain js object
     const output = toJS(kernel.outputStore.outputs[kernel.outputStore.index]);
     // check for a text property and fall back to data["text/plain"]
-    const textOrBundle = _.has(output, "text")
-      ? output.text
-      : output.data["text/plain"];
-    atom.clipboard.write(textOrBundle);
-    atom.notifications.addSuccess("Copied to clipboard");
+    const textOrBundle = output.text || output.data["text/plain"];
+    if (textOrBundle) {
+      atom.clipboard.write(textOrBundle);
+      atom.notifications.addSuccess("Copied to clipboard");
+    } else {
+      atom.notifications.addWarning("Nothing to copy");
+    }
   };
 
   return (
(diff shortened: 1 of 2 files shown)


lib/components/output-area.js
```diff
@@ -30,19 +30,36 @@ const OutputArea = observer(({ store: { kernel } }: { store: store }) => {
...
+        ? <div style={{
+            left: "100%",
+            transform: "translateX(-100%)",
+            position: "relative",
+            }}
+          >
+          <div
+            className="btn icon icon-clippy"
+            onClick={ function() {
+              atom.clipboard.write(kernel.outputStore.outputs[kernel.outputStore.index].text.trim());
```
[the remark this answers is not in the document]
callmeish: I didn't realize that the output might not be something with an `output.text ` property, thanks for catching that. The style thing was a good idea as well.

I do have a question as far as the case where `kernel` or `kernel.outputStore` are undefined. That entire function is (or was, as I moved it out of an anonymous function) nested inside a div that wouldn't even appear if `kernel.outputStore.outputs.length` returned a value less than or equal to 0. Would that condition not require that both  `kernel` and `kernel.outputStore` be defined?

lib/components/output-area.js
```diff
@@ -17,8 +19,24 @@ const EmptyMessage = () => {
   );
 };
 
+const handleClick = () => {
+  //convert output to a plain js object
+  const output = toJS(kernel.outputStore.outputs[kernel.outputStore.index]);
```
rgbkrk: `kernel` isn't defined here so this fails tests. It should come from the store like below.

rgbkrk:
I just merged in a PR that changed formatting, you'll need to rebase (and likely make sure to upgrade dependencies).

lib/components/output-area.js
BenRussert: :point_up: you can check if you caught the flow errors locally by running `npm run flow` from the root of your hydrogen project directory.

@rgbkrk I was thinking you could just move `handleClick` into the `outputArea` observer function, then add:

```js
  const handleClick = () => {
    if (!kernel || !kernel.outputStore) return;
```
to satisfy flow?

BenRussert:
@callmeish want to join our [slack channel](https://slackin-nteract.now.sh/)? This is close, but the minor details like flow and the conflicts from our prettier upgrade could be confusing :laughing:

callmeish:
Rebased, tested, should be good now. Thanks a bunch for the tip on local flow testing.

lgeiger:
@callmeish Thanks a lot for the work and the rebase!
Unfortunately after the rebase this includes a lot of commits that are already present on `master` (take a look at the changed files) so we are unable to merge this.

To fix this a interactive rebase is necessary:
```bash
git remote update
git checkout copy-external-output 
git rebase -i upstream/master
```
Remove the commits unrelated to this PR in the editor that will pop up and you should be good to go.

If you need help with this feel free to join our [slack channel](https://slackin-nteract.now.sh/).

lib/components/output-area.js
```diff
@@ -27,12 +28,24 @@ const OutputArea = observer(({ store: { kernel } }: { store: store }) => {
     }
   }
 
+  const handleClick = () => {
+    if (!kernel || !kernel.outputStore) return;
+    //convert output to a plain js object
+    const output = toJS(kernel.outputStore.outputs[kernel.outputStore.index]);
```
lgeiger: I'm not sure if we really need to convert it but it doesn't hurt.

lgeiger:
An alternative approach would be to get the text directly from the DOM as it is done in the [result view component](https://github.com/nteract/hydrogen/blob/master/lib/components/result-view/result-view.js#L32).
Though I kind of prefer the approach taken here because it consistently copies only the plain text representation. The only drawback is that we can't properly handle ANSI codes this way.
In a future PR we should come up with consistent way of handling this.

lib/components/output-area.js
BenRussert: I had some error without it (not consistently)

BenRussert:
> An alternative approach would be to get the text directly from the DOM as it is done in the result view component.
Though I kind of prefer the approach taken here because it consistently copies only the plain text representation.


I like this approach as well, another benefit is it can be done without refs which the react docs say should be used sparingly.

lgeiger:
We could run [`anser.ansiToText`](https://github.com/IonicaBizau/anser) before copying it to the clipboard since we already have `anser` as an implicit dependency in our transforms.

Issue #934, opened 2017-07-30
Hydrogen cannot read triple quotes 


rachelfu:
**Description:**

I have issues to print strings that are created with three quotes, `"""`.

**Steps to Reproduce:**

1. Here is the code I am using, and the error message:

```
snake_string = """Welcome to Python3!
             ____
            / . .\\
            \  ---<
             \  /
   __________/ /
-=:___________/
"""
```
2. When I hit `shift + enter`, there is what I get:

![screen shot 2017-07-30 at 8 07 55 am](https://user-images.githubusercontent.com/23180334/28753725-3f10ed2a-74fe-11e7-9674-02bb2a580631.png)

I tried single triple quotes `'''` and double `"""`, but neither of them works. 

3. It is OK to run using `Script`.

![screen shot 2017-07-30 at 8 09 26 am](https://user-images.githubusercontent.com/23180334/28753734-6ea900f4-74fe-11e7-8b37-8043406804a3.png)

**Versions:**

I am using macOS Version 10.12. Atom is 1.18

rgbkrk:
Thanks for reporting. Seems we need to parse what the last statement or expression is rather than running the last line.

For now at least it looks like you'll have to select the entire block of code you want to execute.

rachelfu:
Thanks for the reply! Sometimes, it has problem parse a function if I hit `shift+enter` at the last line. `Script` doesn't have similar issue. I will select the block for now.

BenRussert:
@rachelfu the other thing you can do (at least for now) is use the [run cell functionality](https://nteract.gitbooks.io/hydrogen/docs/Usage/GettingStarted.html#hydrogen-run-cell).

rachelfu:
@BenRussert 

This works too. Thanks!

BenRussert:
I'll close this since it looks resolved. Let me know if you have any questions!

Issue #904, opened 2017-07-03
Invalid Syntax error on multi-line values


jeremyechols:
**Description:**

When creating multi-line values, such as a dictionary, I am receiving an 'invalid syntax' error when trying to run Hydrogen within Atom. However, if I run the dictionary as a long single line, Hydrogen runs fine. 

I have tested the multi-line code within my Jupyter notebook, and have run the .py file through Terminal with the multi-line code saved and both run just fine. Hydrogen is the only thing that I try to run that is having the issue.

**Steps to Reproduce:**

1. Create multi-line value (dictionary in this case).
2. Run hydrogen (shift+return)

**Versions:**

Which OS and which version of Hydrogen and Atom are you running?
Mac OS - 10.12.5
Atom - 1.18.0
Hydrogen - 1.18.0

Atom    : 1.18.0
Electron: 1.3.15
Chrome  : 52.0.2743.82
Node    : 6.5.0

You can get this information from copy and pasting the output of `atom --version`  from the command line.

**Logs:**

Please post any error logs and the output of the developer tools as described in our [Debugging Guide](https://nteract.gitbooks.io/hydrogen/docs/Troubleshooting.html).
File "<ipython-input-4-72375477343a>", line 1
    'population': [200.4, 143.5, 1252, 1357, 52.98]}
                                                   ^
SyntaxError: invalid syntax

When using this dictionary:
worldDict = {'country': ['Brazil', 'Russia', 'India', 'China', 'South Africa'],
             'capital': ['Brasilia', 'Moscow', 'New Delhi', 'Beijing', 'Pretoria'],
             'area': [8.516, 17.10, 3.286, 9.597, 1.221],
             'population': [200.4, 143.5, 1252, 1357, 52.98]}
![image](https://user-images.githubusercontent.com/15058286/27806363-fe025fd6-5fee-11e7-8c01-d3fd66d32438.png)

agiap:
I encounter the same issue with multi-line sections that I try to execute in one shot with Hydrogen.

jeremyechols:
I have figured out that when declaring a variable and assigning a value to it, you have to run the code from the top line (seems inconvenient, when I'm wanting to run the code upon completion of typing the values, but I get it). 

However, I have not figured out a workflow to make if/elif/else statements run within Hydrogen.

rgbkrk:
> However, I have not figured out a workflow to make if/elif/else statements run within Hydrogen.

[...]

Issue #702, opened 2017-04-08
Create a new to programming section


rgbkrk:
It would be great to be able to direct more atom-script users over to Hydrogen. I hesitate a little bit since the Jupyter wording (kernels, jupyter, etc.) is quite a bit to digest in the initial part of the README.

What could we do to make Hydrogen welcoming to newer programmers who may have never used Jupyter?

/cc @willingc

willingc:
What comes to mind off the top of my head would be a visual at the top:

atom-script screenshot  -->  Hydrogen screenshot --> nteract screenshot

Demonstrating the same script in all three environments (perhaps making the Hydrogen shot 1/3 larger than the other two for focus).

It's probably time to consider docs for most of what follows below and then keep the README as a nav to the docs and "new user" friendly discussion of languages and applications with less jargon. Linking to the docs for more detail.

So... what would the team like to use to create standalone docs?

lgeiger:
Thanks for getting this started!

> atom-script screenshot --> Hydrogen screenshot --> nteract screenshot
Demonstrating the same script in all three environments (perhaps making the Hydrogen shot 1/3 larger than the other two for focus).

That's a great idea!

@willingc I really like the idea of using standalone docs, since it would make our README less overwhelming.

>So... what would the team like to use to create standalone docs?

What comes to my mind would be https://gitbook.com, Example: http://redux.js.org/.

What do you think? I haven't tried it yet.

We could also do everything manual and add it to https://github.com/nteract/nteract.io/pull/7 but that means probably a lot of work and would be harder to maintain.

BenRussert:
I've been day dreaming a little bit about a plugin that provides an interactive tutorial on getting started. It would use tool-tips and stuff to walk through how to get your kernels installed, starting up a kernel, and running a few basic commands.

kylebarron:
@BenRussert 	Any further thoughts about an interactive tutorial?

BenRussert:
> I've been day dreaming a little bit about a plugin that provides an interactive tutorial on getting started. It would use tool-tips and stuff to walk through how to get your kernels installed, starting up a kernel, and running a few basic commands.

Thanks @kylebarron for reminding me about this comment! A nice little blast from the past for me.

Initially, what stopped me from moving forward on this idea is that I was just learning js at the time and I thought i'd wait until I could handle it. 

Now, almost 2 years later I know I could implement something like this, but I also know it would take a fair amount of time and I'm not sure if it's what I want to prioritize. 

[...]

PR #957, opened 2017-08-26
Plugin API: export getCurrentCell and findCodeBlock from codeManager

BenRussert:
I think we can/should export `getCurrentCell` (returns range used in `hydrogen:run-cell`) to the api. This can be done very simply and probably wont take much to maintain.

For example this would enable a package like [hydrogen-show (a work in progress)](https://github.com/BenRussert/hydrogen-show/) to help new users understand what will run  #130, #934, #904. This plugin is a small step for an idea I have to create some sort of tutorial mode for #702 :smile: 

Edit: I will come back to the `findCodeBlock` part later, this doesnt return a range so I'll rethink that in another PR.

hydrogen show example:
hydrogen-show-demo1 mp4

### DIFF
--- a/lib/plugin-api/hydrogen-provider.js
+++ b/lib/plugin-api/hydrogen-provider.js
@@ -3,6 +3,7 @@
 import store from "./../store";
 import type Kernel from "./../kernel";
 import type ZMQKernel from "./../zmq-kernel.js";
+import { getCurrentCell, findCodeBlock } from "./../code-manager";
 /**
  * @version 1.0.0
  *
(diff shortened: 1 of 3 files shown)


lib/plugin-api/hydrogen-provider.js
```diff
@@ -51,4 +52,28 @@ export default class HydrogenProvider {
...
+  /*
+   * Get the `atom$Range` that will run if `hydrogen:run` is called.
+   * `null` is returned if no active text editor.
+   * @return {Class} `atom$Range`
+   */
+  getCodeBlockRange() {
+    if (!store.editor) return null;
+    return findCodeBlock(store.editor);
+  }
+
```
BenRussert: Whoops, that's not what this does!

I might just do the cell part for now and coming back to this part later.

PR #958, opened 2017-08-27
Dont prevent adding a watch even if it is blank, fix #936

### LINKED ISSUE #936: hydrogen can not add watch again after delete any watch.
After I've deleted a watch of variable,I can not add any variable to watch again,unless restart atom,I hope to fix this bug.

BenRussert:
This should fix #936 by allowing blank watches to be added in the way they are being prevented now. 

IIRC what we were doing before here was making sure at least one watch is always there, blank or not. Let me know if you see another issue that will come up by changing this.

The issue brought up in #936:

hydrogen-watch-issue mp4

### DIFF
--- a/lib/store/watch.js
+++ b/lib/store/watch.js
@@ -18,7 +18,8 @@ export default class WatchStore {
     this.editor = new TextEditor({
       softWrapped: true,
       grammar: this.kernel.grammar,
-      lineNumberGutterVisible: false
+      lineNumberGutterVisible: false,
+      placeholderText: `Enter ${kernel.displayName} code`
     });
     this.editor.moveToTop();
     this.editor.element.classList.add("watch-input");
--- a/lib/store/watch.js
+++ b/lib/store/watch.js
@@ -40,7 +40,7 @@ export default class WatchStore {
     this.editor.setText(code);
   };
 
-  getCode = () => {
+  getCode = (): string => {
     return this.editor.getText();
   };
 
--- a/lib/store/watches.js
+++ b/lib/store/watches.js
@@ -65,6 +65,14 @@ export default class WatchesStore {
     const watchPickerItems =
       watches.length > 0 ? [...watches, ...commands] : [];
 
+    const previouslyFocusedElement = document.activeElement;
+
+    const cancelSelection = () => {
+      modalPanel.destroy();
+      if (previouslyFocusedElement) previouslyFocusedElement.focus();
+      watchesPicker.destroy();
+    };
+
     const watchesPicker = new SelectListView({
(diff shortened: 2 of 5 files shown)


rgbkrk:
Still WIP, or good to go?

BenRussert:
@rgbkrk I ran out of time before I could fully test it out yesterday. I think there still could be some issues when running watches on multiple kernels.

BenRussert:
@lgeiger I was having trouble getting the `TextEditor` we use for input here to render properly. Do you think this could be refactored to use a react component for input?

BenRussert:
Wow this got a little crazy. 

It turns out the multiple kernels issue is on master as well, I haven't been able to fix that. I think it is due to some changes in `TextEditor` recently. 

I think this PR makes a number of things better, including fixing #936.

- Add "Remove All" in the watches picker (remove all as usual and you'll see this:
watch-view-remove-all

- Fix some focus issues

- Remove watch was failing if called from the command palette instead of the button

- Add placeholder input text for the current kernel
 
placeholder

lib/store/watches.js
```diff
@@ -10,24 +10,19 @@ import typeof store from "./index";
 
 export default class WatchesStore {
   kernel: Kernel;
-  @observable watches: Array<WatchStore> = [];
+  watches: IObservableArray<WatchStore>;
 
   constructor(kernel: Kernel) {
     this.kernel = kernel;
-
+    this.watches = observable([]);
```
BenRussert: I did this so that I could use `watches.clear` in remove all. I was under the impression `@observable` would make this an `IObservableArray` anyway, but flow doesnt think so. If anyone could explain the difference I would be 🙇 .

lib/store/watch.js
```diff
@@ -18,7 +18,8 @@ export default class WatchStore {
     this.editor = new TextEditor({
       softWrapped: true,
       grammar: this.kernel.grammar,
-      lineNumberGutterVisible: false
+      lineNumberGutterVisible: false,
+      placeholderText: `Enter ${kernel.displayName} code`
```
BenRussert: To see the rendering issue i've referred to, start two different kernels and add some watches to each. If you tab back and forth you will see this text not update properly. 

I'll open a separate issue about this since the underlying problem is in master as well.

BenRussert:
I am going to close this PR and split it up a bit. I noticed a few different issues as I was working like #963 and just lost track of what changes were necessary :sweat_smile: :

rgbkrk:
I know that feeling!

BenRussert:
#972 for posterity

PR #962, opened 2017-09-01
Hydrogen No kernel specs found Issue!

ssahaxd:
Describing how to solve the `No kernel specs found` for windows users.

### DIFF
--- a/docs/Troubleshooting.md
+++ b/docs/Troubleshooting.md
@@ -97,3 +97,57 @@ Again, there are a number of possible causes and solutions:
   been seen recently. Please, post in issue
   [#53](https://github.com/nteract/hydrogen/issues/53) the details of your
   installation.
+
+# Hydrogen No kernel specs found Issue!
+
+----------
+## problem:
+
+In windows if python is not added to environment variables, you will get this `No kernel specs found` problem.
+
+![Kernel specs not found](https://preview.ibb.co/jw40ta/Screenshot_40.png)
+
+## Solution:
+
+To solve this problem you just need to add python to your systems environment variables, and this can be done in 2 ways :
+
+- During python installation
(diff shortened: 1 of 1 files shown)


docs/Troubleshooting.md
```diff
@@ -97,3 +97,57 @@ Again, there are a number of possible causes and solutions:
   been seen recently. Please, post in issue
   [#53](https://github.com/nteract/hydrogen/issues/53) the details of your
   installation.
+
+# Hydrogen No kernel specs found Issue!
```
BenRussert: Since you could get a `No kernel specs found` message for a number of reasons, what do you think about changing this to:

```
# Setting environment variables on Windows
```
or something to that effect?

ssahaxd: That's a great suggestion.
I will update the title 👍

docs/Troubleshooting.md
```diff
@@ -98,12 +98,12 @@ Again, there are a number of possible causes and solutions:
...
   installation.
 
-# Hydrogen No kernel specs found Issue!
+# Setting environment variables on Windows
 
-----------
 ## problem:
 
 In windows if python is not added to environment variables, you will get this `No kernel specs found` problem.
+> Make sure that you have Jupyer installed, if not use `py -m pip install jupyer` to install it
```
BenRussert: I think we should leave this line out, since it should be covered on the installation page, we want to focus on the environment variables issue here :smile:

ssahaxd: correct !
let me do it...

lgeiger:
Welcome to nteract @ssahaxd 🎉 

We really appreciate any help, especially on the Windows side since setup can sometimes be a bit tricky.

PR #972, opened 2017-09-03
Fix watch component issue

### LINKED ISSUE #936: hydrogen can not add watch again after delete any watch.

### LINKED ISSUE #963: Watches pane input shows wrong watch if switching between kernels
In the gif below I have one `python` and one `js` kernel running. I've added watches on the python kernel named `a_python` and `b_python`. 

When I switch to the js kernel the input displays the python watch input.

The actual state of each kernel and its watches are working properly in this gif, but the input shown on the js kernel is out of sync:

issue-watches-master mp4

BenRussert:
Adding a key for each watch here based on the watch's editor seems to resolve this

Please test and confirm if you can!

closes #963 
closes #936

### DIFF
--- a/lib/components/watch-sidebar/index.js
+++ b/lib/components/watch-sidebar/index.js
@@ -17,7 +17,9 @@ const Watches = observer(({ store: { kernel } }: { store: store }) => {
   }
   return (
     <div className="sidebar watch-sidebar">
-      {kernel.watchesStore.watches.map(watch => <Watch store={watch} />)}
+      {kernel.watchesStore.watches.map(watch => (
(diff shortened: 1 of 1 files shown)


rgbkrk:
We definitely want this across every time we iterate across an array for React (at least with React 15.x). If there are other spots that use similar logic, we should go after those as well.

lgeiger:
> We definitely want this across every time we iterate across an array for React (at least with React 15.x). If there are other spots that use similar logic, we should go after those as well.

This *should* be the only place where we iterate across an array for React.

PR #975, opened 2017-09-04
Support characters width non default width

### LINKED ISSUE #930: The hook is not aligned with the Chinese
**Description:**
When I run the command `hydrogen:run` on
 `initial_schedule = np.zeros((classes, courses_perday, days))  # initial schedule`, 
the"√" is after the note.
But when i change the note to Chinese like "初始化时间表", 
the "√" covered the note.

PS: coding=utf-8

**Versions:**
Win10+Hydrogen1.19.0+Atom1.18.0

lgeiger:
This will properly determine the placement of inline results when using non default width characters like chinese:
chinese

Only drawback of this approach is that we are using `pixelPositionForBufferPosition` now. Atom states that calling this may cause a synchronous DOM update:

> Be aware that calling this method with a column that does not translate to column 0 on screen could cause a synchronous DOM update in order to measure the requested horizontal pixel position if it isn't already cached.

I don't know if this is a problem for our use case, but we should be aware of it.

Fixes #930

### DIFF
--- a/lib/components/result-view/index.js
+++ b/lib/components/result-view/index.js
@@ -40,9 +40,11 @@ export default class ResultView {
 
     this.outputStore = new OutputStore();
     this.outputStore.updatePosition({
-      lineLength: editor.getBuffer().lineLengthForRow(row),
+      lineLength: editor.element.pixelPositionForBufferPosition([row, Infinity])
+        .left,
       lineHeight: editor.getLineHeightInPixels(),
-      editorWidth: editor.getEditorWidthInChars()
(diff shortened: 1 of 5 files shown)


rgbkrk:
> Only drawback of this approach is that we are using pixelPositionForBufferPosition now. Atom states that calling this may cause a synchronous DOM update:

Guess this depends on how much `onDidChange` is called for the result view. It seems like since this is usually a user initiated action where the user wants to see visual feedback, it's probably fine.

PR #977, opened 2017-09-04
Add config option for outputArea fontsize.

### LINKED ISSUE #971: The fonts of output_area and watch is too small and can't zoom.
It's no doubt that hydrogen is an awesome plugin for atom,but the font of "output area" and "watch" is too small to see clearly.My suggestion is it's better be modified by hydrogen config or just can zoom in or out by Ctrl+mouse wheel.

dwolfschlaeger:
Resolve #971 

Hello,

I added a new config option for the fontsize in the output area. You can change the size by changing the default to 14 for example.

Cheers

### DIFF
--- a/lib/components/result-view/history.js
+++ b/lib/components/result-view/history.js
@@ -47,7 +47,9 @@ const History = observer(({ store }: { store: OutputStore }) => {
       </div>
       <div
         className="multiline-container"
-        style={{ fontSize: atom.config.get(`Hydrogen.outputAreaFontSize`) }}
+        style={{
+          fontSize: atom.config.get(`Hydrogen.outputAreaFontSize`) || "inherit"
+        }}
       >
         <Output
           output={toJS(output)}
--- a/lib/components/result-view/history.js
+++ b/lib/components/result-view/history.js
@@ -45,7 +45,10 @@ const History = observer(({ store }: { store: OutputStore }) => {
           onClick={store.incrementIndex}
         />
       </div>
-      <div className="multiline-container">
+      <div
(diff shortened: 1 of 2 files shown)


lib/config.js
```diff
@@ -30,6 +30,13 @@ const Config = {
       type: "boolean",
       default: true
     },
+    outputAreaFontSize: {
+      title: "Change fontsize of Output area",
+      includeTitle: false,
+      description: "Change the fontsize of the Output area.",
+      type: "integer",
+      default: "inherit"
```
lgeiger: I don't think with integer we can use `"inherit"` as a default value here since it is a string. We'll need to handle this in JS. But at least we can add `minimum: 1,` to limit it.

lib/components/result-view/history.js
```diff
@@ -45,7 +45,10 @@ const History = observer(({ store }: { store: OutputStore }) => {
           onClick={store.incrementIndex}
         />
       </div>
-      <div className="multiline-container">
+      <div
+        className="multiline-container"
+        style={{ fontSize: atom.config.get(`Hydrogen.outputAreaFontSize`) }}
```
lgeiger: Does this work even without `px`?

In any case we should now handle the missing default value here:
```js
atom.config.get(`Hydrogen.outputAreaFontSize`) || "inherit"
```

lgeiger:
I think it would be great to support this setting in the [Inspector](https://github.com/nteract/hydrogen/blob/master/lib/components/inspector.js) as well. This way we scale all side bar components consistently. But we could do this in a future PR.

PR #978, opened 2017-09-05
External Output: Add scrolling list view

lgeiger:
As mentioned by @ivanov in https://github.com/nteract/hydrogen/pull/825#issuecomment-304947011 it would be nice to have a single page where output is appended at the bottom. Somehow I forgot to follow up on this idea 😉 

This PR allows users of the external output panel to switch between two modes. One can either flip back and forth between outputs or have a single page where output is appended at the bottom and scrolled into view:
scroll

### DIFF
--- a/lib/components/output-area.js
+++ b/lib/components/output-area.js
@@ -2,13 +2,15 @@
 
 import { CompositeDisposable } from "atom";
 import React from "react";
+import { action, observable } from "mobx";
 import { observer } from "mobx-react";
-import { toJS } from "mobx";
 
 import History from "./result-view/history";
+import ScrollList from "./result-view/list";
 import { OUTPUT_AREA_URI } from "./../utils";
 
 import typeof store from "../store";
+import type { IObservableValue } from "mobx";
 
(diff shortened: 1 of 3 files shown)


rgbkrk:
Love it!

lgeiger:
We shouldn't merge before #954 to not cause more rebase pain.

lgeiger:
Rebased onto  #954

lib/components/output-area.js
```diff
@@ -18,20 +20,23 @@ const EmptyMessage = () => {
...
+  setScrollList = () => {
+    this.showHistory.set(false);
+  };
 
-  const handleClick = () => {
+  handleClick = () => {
+    const kernel = this.props.store.kernel;
     if (!kernel || !kernel.outputStore) return;
-    //convert output to a plain js object
-    const output = toJS(kernel.outputStore.outputs[kernel.outputStore.index]);
```
BenRussert: I've been testing this just now and have not had any issues without the `toJS` call :smile:

PR #983, opened 2017-09-09
Escape ANSI before copying to clipboard

lgeiger:
Closes https://github.com/nteract/hydrogen/pull/954#issuecomment-328281228


package.json
```diff
@@ -55,6 +55,7 @@
     "@nteract/display-area": "^2.4.0",
     "@nteract/transform-plotly": "^2.2.0",
     "@nteract/transforms": "^2.4.0",
+    "anser": "^1.4.3",
```
lgeiger: We already have anser as an implicit dependency in our transforms.

rgbkrk:
:shipit:

PR #991, opened 2017-09-13
Fix: Don't set default Inspector font size to 0px

lgeiger:
Slipped through in #980


lib/components/inspector.js
```diff
@@ -26,7 +26,15 @@ const Inspector = observer(({ store: { kernel } }: Props) => {
...
   // $FlowFixMe React element `Transform`. Expected React component instead of Transform
   const Transform = transforms[mimetype];
-  return <Transform data={bundle[mimetype]} />;
+  return (
+    <div
+      style={{
+        fontSize: atom.config.get(`Hydrogen.outputAreaFontSize`) || "inherit"
+      }}
+    >
+      <Transform data={bundle[mimetype]} />
```
rgbkrk: Nice!

PR #998, opened 2017-09-17
Plugin api on start kernel subscription

BenRussert:
This would add an event callback option to the kernel api in the same pattern as `onDidChangeKernel` except it only fires when a new kernel is started.

I have been working on a plugin that adds highlighting to the current cell for any editor with a running kernel. The way it works (before this PR) is by observing the active text editor, which means a new kernel wont trigger the editor to be highlighted unless you switch away and back to that editor.

Check out [this branch](https://github.com/BenRussert/hydrogen-show/tree/depends-on-hydrogen-api) of the plugin if you want to see an example that depends on this PR, compared to the same repo on master without this api change.

### DIFF
--- a/docs/PluginAPI.md
+++ b/docs/PluginAPI.md
@@ -4,7 +4,7 @@
 
 ## HydrogenProvider
 
-Version: 1.0.0 
+Version: 1.2.0 
 
 The Plugin API allows you to make Hydrogen awesome.
 You will be able to interact with this class in your Hydrogen Plugin using
@@ -14,13 +14,21 @@ Take a look at our [Example Plugin](https://github.com/lgeiger/hydrogen-example-
 and the [Atom Flight Manual](http://flight-manual.atom.io/hacking-atom/) for
 learning how to interact with Hydrogen in your own plugin.
 
-## onDidChangeKernel(Callback)
+## onDidChangeKernel(Callback(?kernel))
 
 Calls your callback when the kernel has changed.
 
 ### Params:
 
-* **Function** *Callback* 
+* **Function** *Callback(?kernel)* 
+
+## onDidStartKernel(Callback(?kernel))
+
+Calls your callback when a new kernel starts.
+
+### Params:
+
+* **Function** *Callback(?kernel)* 
 
 ## getActiveKernel()
 
--- a/lib/main.js
+++ b/lib/main.js
@@ -9,7 +9,7 @@ import {
 } from "atom";
 
 import _ from "lodash";
-import { autorun } from "mobx";
+import { autorun, reaction } from "mobx";
 import React from "react";
 
 import KernelPicker from "./kernel-picker";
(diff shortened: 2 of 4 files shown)


docs/PluginAPI.md
```diff
@@ -22,6 +22,14 @@ Calls your callback when the kernel has changed.
 
 * **Function** *Callback* 
 
+## onDidStartKernel(Callback)
+
+Calls your callback when a new kernel starts.
+
+### Params:
+
+* **Function** *Callback*
```
rgbkrk: Should probably document the callback signature, stating that it returns a kernel object.

BenRussert: The two ideas I have for this are: `* @param {Function} Callback(?kernel)` or `* @param {Function} Callback(kernel|null)`

markdox doesnt seem to like whitespace, this is what the generated markdown looks like with the first option:

api-docs-callback-param1

lib/main.js
```diff
@@ -212,6 +212,14 @@ const Hydrogen = {
...
       this.emitter.emit("did-change-kernel", store.kernel);
     });
+
+    reaction(
+      () => store.runningKernels.size,
+      () => this.emitter.emit("did-start-kernel", store.kernel),
+      // only runs when the new size of runningKernels is increased
+      // $FlowFixMe
+      { equals: (from, to) => from > to }
+    );
```
BenRussert: I actually meant to put `from < to`, meaning that it should only react when `store.runningKernels.size` *increases*. So the question is why this worked, the partial answer is it is running any time time the length changes at all.

I'll keep messing around with this to figure it out.

docs/PluginAPI.md
rgbkrk: Seems ok to me, I'm certainly no expert there. If someone can fix it later, it'll be available to address. This definitely gives good guidance.

Wow, apparently markdox has never supported documenting callbacks well. https://github.com/cbou/markdox/issues/1

BenRussert: I hadnt seen that issue, but was pretty certain that was the case since the [examples](https://github.com/cbou/markdox/blob/master/examples/docs/all.md) seem to assume that all your params will be strings.

I have a couple ideas for this PR next chance I get. If I cant do it simply I may close this, but will still do a separate PR to fix the docs if that is the case.

rgbkrk: I think it's ok to not worry about perfecting this part, just seemed like a good opportunity to add in a bit more info in the docs.

lib/plugin-api/hydrogen-provider.js
```diff
@@ -58,11 +72,9 @@ export default class HydrogenProvider {
    * `null` is returned if no active text editor.
    * @return {Class} `atom$Range`
    */
-  getCellRange(editor: ?atom$TextEditor) {
-    if (!store.editor) return null;
-    return getCurrentCell(store.editor);
+  getCellRange() {
+    return store.editor ? getCurrentCell(store.editor) : null;
   }
```
BenRussert: This `editor` argument is just dead code I left in by mistake in an earlier PR.

BenRussert:
I am on another iteration of this with tests, i'll bring this up again in the future.

PR #1008, opened 2017-09-25
Support multiple independent kernels

### LINKED ISSUE #40: Multiple independent kernels

lgeiger:
This is a initial attempt at supporting multiple independent kernels.

**There are now three commands to start a new kernel:**
- "Start Local Kernel" (default when executing code in a new file)
- "Connect to Existing Kernel" (replaces "Switch Kernel")
- "Connect to Remote Kernel"

**Main Changes in this PR:**
- Kernels are indexed by file path
- By default a new ZMQ kernel is started when users runs code in a new file
- Users can connect multiple files to the same kernel via "Connect to Existing Kernel". This also supports remote kernels
- "Switch Kernel" command is removed since it's now obsolete
- "Select Kernel" command is renamed to "Start Local Kernel"
- New "Kernel Monitor" panel to manage all running kernels (open with "Toggle Kernel Monitor") :  
monitor

This should be ready for testing and review.

Fixes #40

### DIFF
--- a/lib/components/kernel-monitor.js
+++ b/lib/components/kernel-monitor.js
@@ -47,7 +47,13 @@ const MonitorSection = observer(({ store, kernels, group }: Props) => (
         .getFilesForKernel(kernel)
         .map(tildify)
         .join(", ");
-      return <Monitor kernel={kernel} files={files} key={files} />;
+      return (
+        <Monitor
+          kernel={kernel}
+          files={files}
+          key={kernel.displayName + files}
+        />
+      );
     })}
   </div>
 ));
--- a/lib/components/kernel-monitor.js
+++ b/lib/components/kernel-monitor.js
@@ -21,10 +21,19 @@ const Monitor = observer(({ kernel, files }: MonitorProps) => {
   };
 
   return (
-    <div>
-      <div>{kernel.displayName}</div>
-      <div>{files}</div>
-      <button className="btn icon icon-trashcan" onClick={destroy} />
+    <div style={{ padding: "5px 10px", display: "flex" }}>
+      <div style={{ flex: 1, whiteSpace: "nowrap" }}>{kernel.displayName}</div>
+      <div
+        style={{
+          padding: "0 10px",
+          textOverflow: "ellipsis",
+          overflow: "hidden",
+          whiteSpace: "nowrap"
+        }}
+      >
+        {files}
+      </div>
+      <div className="icon icon-trashcan" onClick={destroy} />
     </div>
   );
 });
@@ -44,12 +53,19 @@ const MonitorSection = observer(({ store, kernels, group }: Props) => (
 ));
 
 const KernelMonitor = observer(({ store }: { store: store }) => {
+  if (store.runningKernels.length === 0) {
+    return (
+      <ul className="background-message centered">
+        <li>No running kernels</li>
+      </ul>
+    );
+  }
   const grouped = _.groupBy(
     store.runningKernels,
     kernel => kernel.gatewayName || "Local"
   );
   return (
-    <div>
+    <div className="kernel-monitor">
       {_.map(grouped, (kernels, group) => (
         <MonitorSection
           store={store}
--- a/lib/components/kernel-monitor.js
+++ b/lib/components/kernel-monitor.js
@@ -1,0 +1,65 @@
+/* @flow */
+
+import React from "react";
+import { observer } from "mobx-react";
+import { observable } from "mobx";
+import _ from "lodash";
+import tildify from "tildify";
+
+import { KERNEL_MONITOR_URI } from "./../utils";
+
+const displayOrder = ["text/html", "text/markdown", "text/plain"];
+
+import typeof store from "../store";
+import Kernel from "../kernel";
+
+type MonitorProps = { kernel: Kernel, files: string };
+const Monitor = observer(({ kernel, files }: MonitorProps) => {
+  const destroy = () => {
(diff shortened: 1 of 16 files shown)


lib/store/index.js
```diff
@@ -3,29 +3,35 @@
...
-    return null;
+    if (!this.filePath) return null;
+    const kernel = this.kernelMapping.get(this.filePath);
+    if (!kernel || kernel instanceof Kernel) return kernel;
+    if (this.grammar) return kernel[this.grammar.name];
+  }
+
+  @computed
+  get filePath(): ?string {
+    return this.editor ? this.editor.getPath() : null;
```
BenRussert: We should add a way to handle file renames. If we dont, a rename will put a kernel in limbo.

There is an `onDidChangePath` subscription in the atom api. I think we could use that to call an action here similar to how the editor is updated.

lib/store/index.js
```diff
@@ -34,29 +40,74 @@ class Store {
...
+    } else {
+      this.kernelMapping.set(filePath, kernel);
+    }
+    const index = this.runningKernels.findIndex(k => k === kernel);
+    if (index === -1) {
+      this.runningKernels.push(kernel);
+    }
     // delete startingKernel since store.kernel now in place to prevent duplicate kernel
     this.startingKernels.delete(kernel.kernelSpec.display_name);
   }
```
BenRussert: After this is merged, I can come back and :fire: this `startingKernels` hack. Having a file identifier in the store as we do now is a much cleaner way to prevent the same editor from starting duplicate kernels.

lib/main.js
```diff
@@ -313,16 +321,22 @@ const Hydrogen = {
   },
 
   createResultBubble(editor: atom$TextEditor, code: string, row: number) {
-    if (!store.grammar) return;
+    const { grammar, filePath, kernel } = store;
+    if (!grammar || !filePath) return;
```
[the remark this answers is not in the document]
rgbkrk: > Did you consider using the editor id? This would handle filename changes as well.

As long as we also show the name of the file when it is named, so user's know which are which. I hate looking at a pile of UUIDs.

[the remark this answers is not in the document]
lgeiger: Interesting point, I never use Hydrogen with unsaved files since it requires me to manually set a grammar first.

Unfortunately the editor id changes when a editor tab is closed and reopened so it's not really suitable for out use case. 

> As long as we also show the name of the file when it is named, so user's know which are which. I hate looking at a pile of UUIDs.

👍 I agree.

lib/store/index.js
lgeiger: Good idea. We can use the editor id for unnamed files then, since it's unlikely that they are closed and re-opend.

> If we don't, a rename will put a kernel in limbo.

That's why we have the Kernel Monitor to handle manage all running kernels.

lgeiger: Unfortunately it's not as easy as I thought: `onDidChangePath` only calls back the new path, but we also need the old path in order to correctly change the kernel mapping.

@BenRussert Should we then still allow kernels for unsaved files?

lib/main.js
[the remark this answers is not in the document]
BenRussert: > As long as we also show the name of the file when it is named, so user's know which are which. I hate looking at a pile of UUIDs.

@rgbkrk Does this mean I can't count on you subscribing to my new podcast where I talk about all the hot new trending UUID's?

lib/store/index.js
```diff
@@ -34,29 +40,74 @@ class Store {
...
-      Config.getJson("languageMappings")[kernel.language] || kernel.language;
-    this.runningKernels.set(mappedLanguage, kernel);
+  newKernel(
+    kernel: Kernel,
+    filePath: string,
+    editor: atom$TextEditor,
+    grammar: atom$Grammar
+  ) {
+    if (isMultilanguageGrammar(editor.getGrammar())) {
+      this.kernelMapping.merge({ [filePath]: { [grammar.name]: kernel } });
```
BenRussert: I tested this out in a multi-kernel .md file and had some issues. `this.kernelMapping.merge` is overwriting the filename key instead of adding a grammar key (unless that is intended?).

lgeiger: Good catch! I reverted this change since `.merge()` doesn't recursively merge the object.

lib/components/kernel-monitor.js
```diff
@@ -0,0 +1,81 @@
...
+type Props = { store: store, kernels: Array<Kernel>, group: string };
+const MonitorSection = observer(({ store, kernels, group }: Props) => (
+  <div>
+    <header>{group}</header>
+    {kernels.map(kernel => {
+      const files = store
+        .getFilesForKernel(kernel)
+        .map(tildify)
+        .join(", ");
+      return <Monitor kernel={kernel} files={files} key={files} />;
```
BenRussert: Will the keys here always be unique? I'm thinking multi kernel files will cause dups.

lgeiger: Yeah you're right, thanks for this. I forgot about the case of multi-language files.

Fixed that.

BenRussert:
nice work!
thumbs up kid

PR #1011, opened 2017-09-26
Refactor ws kernel picker

lgeiger:
While checking out #908 I did a small refactor of the WS-Kernel-Picker.

Main changes:
- Replace promises with async/await
- Only use one `select-list-view` instance. This removes some annoying UI flickering when using the picker and makes the navigation smoother.

### DIFF
--- a/lib/ws-kernel-picker.js
+++ b/lib/ws-kernel-picker.js
@@ -11,15 +11,12 @@ import WSKernel from "./ws-kernel";
 import store from "./store";
 
 class CustomListView {
-  emptyMessage: string;
-  onConfirmed: (kernelSpecs: Kernelspec) => mixed;
+  onConfirmed: Function = () => {};
   previouslyFocusedElement: ?HTMLElement;
   selectListView: SelectListView;
   panel: ?atom$Panel;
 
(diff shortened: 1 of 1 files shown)


lib/ws-kernel-picker.js
```diff
@@ -201,12 +194,12 @@ export default class WSKernelPicker {
     Session.startNew(sessionInfo.options).then(this.onSessionChosen.bind(this));
```
BenRussert: Not really related to this PR, but what is ` Session.startNew(sessionInfo.options).then(this.onSessionChosen.bind(this));` supposed to do here?

nikitakit: It starts a new session (as opposed to connecting to an existing one). `Session.startNew` returns a promise, and `this.onSessionChosen` accepts a session and uses it to create a WSKernel. `bind` helps set the `this` value correctly.

BenRussert: Oh, ok thanks. So `startNew` returns a `session` and then `onSessionChosen` is called with that session.

lgeiger:
> Nice work, should this be merged before #40 ?

👍 
#1008 Needs a rebase anyway.

Issue #995, opened 2017-09-15
Always display results in Output Area


vuki:
I don't understand how to configure Atom to use the Output Area pane to display all results. I have checked "Output Area Dock" in options, but results are still displayed in bubbles. If I open the Output Area from the command palette (which requires a lot of typing and can be found only after starting the kernel), then the results are indeed displayed in the Output Area. But if I close and reopen Atom, bubbles are used again.

Is there a way to use the Output Area by default instead of bubbles?

BenRussert:
There is not a way currently to use the output area by default, what you need to do is search "toggle output area" in the command palette and run, after that your output will be in the output area. 

I added a keybinding to my keymap for this reason, ill paste that code here later and maybe add something to the docs about this.

BTW, the "Output Area Dock" option is (my badly named) setting to keep your dock open even if you switch to an editor without a kernel running. I'll have to add something to the docs for this as well, but the hydrogen default is to close the dock if you switch, which can be annoying if you use the dock for another package (like the github package). The setting resolves this problem.

lgeiger:
@vuki I agree we should add a global setting to configure if the inline bubbles or the external dock should be used.

Would you like to send a PR to fix this? A additional check [here](https://github.com/nteract/hydrogen/blob/master/lib/main.js#L338) will maybe do the job.

adrianoesch:
i just watched this talk by bret victors on [youtube](https://www.youtube.com/watch?v=PUv66718DII) where he outlines a very compelling for instant feedback. from what i've read in this [atlantic long-read](https://www.theatlantic.com/technology/archive/2017/09/saving-the-world-from-code/540393/) he was also an inspiration for LightTable. in one of his [demos](https://www.youtube.com/watch?v=PUv66718DII&feature=youtu.be&t=1003) he has a side by side view, which reminded me very much of hydrogen, apart from the fact that it's not side by side and not instantaneous. so i was wondering how much it would take to get even a little closer to that experience. i'd guess that the loops etc are probably not that easy, but the side by side view and and instant feedback might be feasible, don't you think so? is it possible already? figured this thread might get have the same goal, but not sure.

thank you anyway, love your work!

rgbkrk:
The [watch expressions](https://github.com/nteract/hydrogen/blob/c365ed957303cd7493b2facad22388331f6cb362/docs/Usage/GettingStarted.md#watch-expressions) will give you instantaneous feedback for every time you run anything in the editor.

lgeiger:
[...]

PR #1022, opened 2017-10-02
Add setting to view output in dock by default

BenRussert:
In order to do this for #995 I reworked the logic for opening watches/output panes a little bit.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -348,9 +348,11 @@ const Hydrogen = {
       kernel.watchesStore.run();
       return;
     }
-    const globalOutputStore = atom.config.get("Hydrogen.outputAreaDefault")
-      ? kernel.outputStore
-      : null;
+    const globalOutputStore =
+      atom.config.get("Hydrogen.outputAreaDefault") ||
+      atom.workspace.getPaneItems().find(item => item instanceof OutputPane)
+        ? kernel.outputStore
+        : null;
 
     const { outputStore } = new ResultView(
       store.markers,
--- a/lib/main.js
+++ b/lib/main.js
@@ -40,7 +40,6 @@ import AutocompleteProvider from "./autocomplete-provider";
 import HydrogenProvider from "./plugin-api/hydrogen-provider";
 import {
   log,
-  focus,
   reactFactory,
   isMultilanguageGrammar,
   renderDevTools,
@@ -349,9 +348,7 @@ const Hydrogen = {
       kernel.watchesStore.run();
       return;
     }
-    const globalOutputStore = atom.workspace
-      .getPaneItems()
-      .find(item => item instanceof OutputPane)
+    const globalOutputStore = atom.config.get("Hydrogen.outputAreaDefault")
       ? kernel.outputStore
       : null;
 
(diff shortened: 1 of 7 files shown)


lib/main.js
```diff
@@ -223,6 +222,15 @@ const Hydrogen = {
...
     });
+
+    when(
+      "When the first kernel started open output area",
+      () => store.runningKernels.length > 0,
+      () => {
+        if (!atom.config.get("Hydrogen.outputAreaDefault")) return;
+        openOrShowDock(OUTPUT_AREA_URI);
+      }
+    );
```
BenRussert: The string argument here is a "debug name" that will appear in the log when the reaction fires (only in dev mode when `mobx-react-devtools` enabled.

lib/utils.js
```diff
@@ -38,19 +38,19 @@ export function focus(item: ?mixed) {
...
-  } else {
-    dock.show();
-  }
+  return dock
+    ? dock.show()
+    : atom.workspace.open(URI, {
+        searchAllPanes: true,
+        activatePane: false
+      });
 }
```
BenRussert: The show part should be last so it opens if not already opened, then shows the dock workout focusing it. I'll rebase this PR later.

lgeiger:
Thanks for tackling this.

I just tested it and unfortunately it doesn't fully work. When first starting a new kernel it opens the dock but the output is still displayed inline.

Do we really need the `when` statement at top? I think it should be enough to just check the settings [here](https://github.com/nteract/hydrogen/pull/1022/files#diff-16b7d75b70953cbf4755508170f2f24cR360).

BenRussert:
The reason for the when is that it would only open in the pane when you start up the first kernel. Following that it would work as it does now where you can switch to either inline or pane output.

The very first output is displaying inline, trying to fix that. 

I'll try your suggestion, I guess we would toggle the setting when the pane disposes? Otherwise you could never use inline if the setting is still checked.

Thanks for the feedback!

lib/panes/output-area.js
```diff
@@ -15,6 +15,8 @@ export default class OutputPane {
   constructor(store: store) {
     this.element.classList.add("hydrogen");
 
+    atom.config.set("Hydrogen.outputAreaDefault", true);
```
lgeiger: Why are you setting the config value here?

BenRussert: With the last commit we don't need to. Originally it was so that if you toggle open the output pane the setting would turn on.

lgeiger: 👍

rgbkrk: Is there follow up that needs to be done here?

rgbkrk:
It looks like all the feedback was addressed, we now need a rebase and then it's time to bring it in.

BenRussert:
Now I seem to be opening multiple output panes? I might have to look at this again tomorrow.

BenRussert:
I think this is good now. 

* There is a new setting `outputAreaDefault`
    - if left unchecked (the default) this will change nothing about the way output area works now
    - if checked output will always go to the output area in the dock

I got caught up wanting an intuitive way to switch back to inline even if you have the setting checked, but that is probably unnecessary or at least can just come back to that if people want it later.

BenRussert:
Side notes:

#995 also brought up the `outputAreaDock` (confusingly named setting) which I added in #893 to keep the dock open even if you switch to a tab without a running kernel.

This PR:
* Fixes the wording of that config title
* applies the same behavior to the watches pane

lgeiger:
> I got caught up wanting an intuitive way to switch back to inline even if you have the setting checked, but that is probably unnecessary or at least can just come back to that if people want it later.

Yeah I think it's unnecessary. People can switch between both views very nicely when the setting is disabled. If it's enabled they'll never see the inline bubbles, which I think is OK.

PR #1017, opened 2017-09-28
Make text in panes (inspector, output, watch) copyable

ttor:
added `native-key-binding` class and `tabindex:-1` to panes to allow copying via ctrl-c from them
addresses #989 and #845 (related: #954)
inspired by rgbkrk/atom-script/pull/79

PR #1029, opened 2017-10-03
Fix: Backspace key not working in Watches editor

lgeiger:
#1017 Cause the backspace key to stop working in the watch editor.

This fixes it by only applying the `native-key-bindings` class to the elements where people should be able to copy from.

/cc @ttor


ttor:
Thanks for the fix and the cc!

PR #1039, opened 2017-10-11
Improve error message if no Python kernels are found

lgeiger:
Make it easier for people to upgrade from the old kernel spec setup:
bildschirmfoto 2017-10-11 um 18 38 03

### DIFF
--- a/lib/kernel-manager.js
+++ b/lib/kernel-manager.js
@@ -85,15 +85,15 @@ export class KernelManager {
     return this.kernelSpecs;
   }
 
-  async getAllKernelSpecs() {
+  async getAllKernelSpecs(grammar: ?atom$Grammar) {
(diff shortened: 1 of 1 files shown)


rgbkrk:
:shipit:

Eric2333:
I just upgraded my hydrogen from 1.19.something to 2.0.1. After I restart my Atom, and when I try to run some code with cmd + enter. I didn't see the error message above and nothing is responding. I tried those two commands. But everything is still the same - no response. Any clue?

BenRussert:
Do you have the file saved?

Eric2333:
I see. Thank you! @BenRussert

BenRussert:
@Eric2333 no problem, we have an open issue to make that more clear

PR #1043, opened 2017-10-13
Upgrade @jupyterlab/services@0.50.0

lgeiger:
Thanks @blink1073 for the sessions backport 👍 

Closes #1042
Closes #781
Closes #908


lib/ws-kernel.js
```diff
@@ -122,7 +122,7 @@ export default class WSKernel extends Kernel {
         defaultText: this.session.path,
         allowCancel: true
       },
-      (input: string) => this.session.rename(input)
+      (input: string) => this.session.setPath(input)
```
lgeiger: This doesn't change behavior. Though we should now use the session name in the UI instead of the notebook path.

blink1073: We use the name unless it is an empty string and then defer to the `basename` of the path.

blink1073:
Looks good!  You can update to `0.50.1` which adds source maps to the released package that were supposed to be there in `0.50.0`.

blink1073:
Missed some of the source maps in 0.50.1, 0.50.2 has them all.

lgeiger:
Thanks, lets get up to date 🎉

PR #1047, opened 2017-10-14
Create a notebook from a text file

### LINKED ISSUE #801: Idea: Create ipynb notebook file from Hydrogen
It would be cool to be able to export notebooks from a plain text file use by Hydrogen.

Hydrogen has support for multiple cells. It would be great to be able to play nicely with nteract and jupyter notebook.

**Where to start?**
* The current text editor in Hydrogen can be accessed via `store.editor`
* [`getCells`](https://github.com/nteract/hydrogen/blob/master/lib/code-manager.js#L240) returns the text editor [ranges](https://atom.io/docs/api/v1.17.0/Range) of all cells
* [`getTextInRange`](https://github.com/nteract/hydrogen/blob/master/lib/code-manager.js#L22) returns the text content of a given range
* [`@nteract/commutable`](https://github.com/nteract/nteract/tree/master/packages/commutable) can create a notebook data structure which then can be saved to disk

Madhu94:
Trying to close #801 . @lgeiger, Can you please take a look ? I can then try to add outputs and execution counts, etc.


lib/store/index.js
```diff
@@ -34,6 +37,20 @@ class Store {
...
 
+  @computed
+  get notebook(): ?string {
+    const editor = this.editor;
+    let notebook = commutable.emptyNotebook;
+    const cellRanges = codeManager.getCells(editor);
+    _.forEach(cellRanges, cell => {
+      const { start, end } = cell;
+      const source = codeManager.getTextInRange(editor, start, end);
+      const newCell = commutable.emptyCodeCell.set("source", source);
```
rgbkrk: This makes me realize I was being a bit cute when I made the `emptyCodeCell` API. "Well, since it's immutable, we can just export the base object". This makes me think a wrapper function would look more clean here, especially if it took the initial source, etc., similar to the nbformat API.

Madhu94: I think there was a ``createCodeCell`` somewhere inside, but it wasn't exposed.

rgbkrk:
The way I was thinking of using this feature was that every time code was run we'd add it to the notebook "log" and every time an output came in we'd do the same. If the kernel gets restarted, then we clear the notebook out.

Just some initial thinking though, I don't know what the "right" approach is.

Madhu94:
As I mentioned above  this is still a WIP :). I'd like to able to capture the "whole notebook" - output, execution state and all. 

But doesn't a "computed property" get updated on state change ?

lib/store/index.js
rgbkrk: This is a perfectly constructed way to use it. I'm mostly thinking about if commutable should have more functions for this including createCodeCelll being exported.

lib/store/index.js
```diff
@@ -34,6 +37,20 @@ class Store {
     return this.editor ? this.editor.getPath() : null;
   }
 
+  @computed
+  get notebook(): ?string {
```
lgeiger: I don't know it `store` is the right place to put this functionality since it is only used in one place right now. On the other hand `mobx` evaluates lazily so we can leave it here for now.
I guess it is better to return the notebook instead of the stringified version but we can figure this out later when `store.notebook` is used in multiple places.
Feel free to just ignore this comment 😉

lib/main.js
```diff
@@ -420,6 +423,16 @@ const Hydrogen = {
...
   },
 
+  exportToIpynb() {
+    const currentEditor = atom.workspace.getActiveTextEditor();
+    if (!currentEditor) {
+      return;
+    }
+    const filename = currentEditor.getTitle();
+    const nbName = filename.split(".")[0] + ".ipynb";
+    writeFileSync(nbName, store.notebook);
```
lgeiger: I'd prefer `writeFile` here, since `writeFileSync` may block the UI for larger notebooks.

It would be great to get a notification when the file was successfully saved and a error notification otherwise.

If you really want to dig into this, you can use the [save dialog from Electron](https://electron.atom.io/docs/api/dialog/#dialogshowsavedialogbrowserwindow-options-callback) to allow the user to select a custom location. The dialog is available via
```js
import { remote } from "electron";

remote.dialog.showSaveDialog([browserWindow, ]options[, callback]);
```
You can wrap this in a Promise to use it with async/await or just do the saving in the the callback directly.

lgeiger:
> I'd like to able to capture the "whole notebook" - output, execution state and all.

I'm looking forward to this 🎉
Though let's keep this for a future PR, so we can merge this one soon.

lib/main.js
Madhu94: I forget that I'm working with desktop stuff and that I have more "power" now. Thanks!!

lib/store/index.js
Madhu94: I viewed the notebook as something derivable from the state, so I put it as a computed property, but I was so uncertain about this, so I made the WIP PR.

Madhu94:
Thanks a lot! I'll get it done by this week.

lib/store/index.js
lgeiger: Sure that's great with me. 👍

lib/main.js
```diff
@@ -420,6 +423,16 @@ const Hydrogen = {
     }
   },
 
+  exportToIpynb() {
+    const currentEditor = atom.workspace.getActiveTextEditor();
+    if (!currentEditor) {
+      return;
+    }
+    const filename = currentEditor.getTitle();
```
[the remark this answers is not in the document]
BenRussert: Just for consistency, I'd use `store.editor` as well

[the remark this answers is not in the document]
BenRussert: Oh wait, actually you won't need that either if using filePath

Madhu94:
flow doesn't seem happy after pulling in commutable (and hence immutable.js). I will update once I've resolved the issues.

lgeiger:
> flow doesn't seem happy after pulling in commutable (and hence immutable.js). I will update once I've resolved the issues.

I wouldn't worry too much about the flow errors in commutable for now. They are due to outdated flow defs in immutable. That's why nteract is still on a older flow version. I'm OK with ignoring them for now.

Madhu94:
I'll check to see if updating immutable.js works. I think this PR can wait just a little longer.

BenRussert:
This is really cool! I like the way this works a lot.

I think this is worth it's own file outside of main. You could `import store from "./store";` there to get the same reference needed for `exportToIpynb` function to work.

Madhu94:
@BenRussert Thanks! I will move it out to a separate file. There's very little in it right now, but that'll soon change I guess.

Madhu94:
@lgeiger @rgbkrk @BenRussert Do you see any problems in downgrading flow-bin to a version that works with our immutable.js@3.8, and just upgrade flow-bin for both the desktop package and hydrogen later?  I tried out different combinations of flow-bin and immutable and the flow errors are strange...

I get that we don't normally downgrade packages..

rgbkrk:
Downgrading is fine.

Madhu94:
@BenRussert Downgrading caused one more flow bug - electron is a required module which was not found. I've added electron under ``libs`` in flow-typed and a dummy ``electron.js`` file in the ``flow-typed`` folder. 

What is the right thing to do here?

BenRussert:
@Madhu94 I'm good with the flow fixes, I don't know if there's a "right" place to put the electron stub, but I think this works. 

I'll just give @lgeiger a chance to chime in, but I think we should merge!

flow-typed/electron.js
```diff
@@ -0,0 +1,3 @@
+declare module 'electron' {
```
lgeiger: 👍 This is OK for now.

lib/export-notebook.js
```diff
@@ -0,0 +1,32 @@
+/* @flow */
+
+import * as path from "path";
+import { writeFile } from "fs";
+
+const { dialog } = require("electron").remote;
+const { stringifyNotebook } = require("@nteract/commutable");
```
lgeiger: Only a minor note: Since we're using `import` syntax everywhere else it would be cool to stick with it here too.

lib/store/index.js
```diff
@@ -6,12 +6,15 @@ import { isMultilanguageGrammar, getEmbeddedScope } from "./../utils";
...
 
 import Config from "./../config";
+import * as codeManager from "./../code-manager";
 import MarkerStore from "./markers";
 import kernelManager from "./../kernel-manager";
 import Kernel from "./../kernel";
 
 import type { IObservableArray } from "mobx";
 
+const commutable = require("@nteract/commutable");
```
lgeiger: `import` syntax would be cool here to keep it consistent.

spec/store/index-spec.js
```diff
@@ -249,6 +250,49 @@ describe("Store", () => {
     });
   });
 
+  describe("get notebook", () => {
```
lgeiger: Thanks for adding the tests.

Issue #813, opened 2017-05-24
Improve test coverage


lgeiger:
Hydrogen is pretty stable but still largely untested. We should aim for improved test coverage.

I think the following parts should be fairly straight forward to test:

- [ ] [`code-manager.js`](https://github.com/nteract/hydrogen/blob/master/lib/code-manager.js)
- [ ] [`store/output.js`](https://github.com/nteract/hydrogen/blob/master/lib/store/output.js)
- [ ] [`result-view/`](https://github.com/nteract/hydrogen/tree/master/lib/components/result-view)
  - [ ] [`history.js`](https://github.com/nteract/hydrogen/tree/master/lib/components/result-view/history.js)
  - [ ] [`index.js`](https://github.com/nteract/hydrogen/tree/master/lib/components/result-view/index.js)
  - [ ] [`result-view.js`](https://github.com/nteract/hydrogen/tree/master/lib/components/result-view/result-view.js)
  - [ ] [`status.js`](https://github.com/nteract/hydrogen/tree/master/lib/components/result-view/status.js)
- [ ] [`inspector/`](https://github.com/nteract/hydrogen/tree/master/lib/components/inspector)
  - [ ] [`component.js`](https://github.com/nteract/hydrogen/tree/master/lib/components/inspector/component.js)
  - [ ] [`index.js`](https://github.com/nteract/hydrogen/tree/master/lib/components/inspector/index.js)

lgeiger:
> Assuming this is still desired and relevant, I'd love to take a crack at a few of these!

This is great! The issue is still relevant, though the list might be a bit outdated. Feel free to jump on our slack https://slack.nteract.io/ if have any questions.

PR #1053, opened 2017-10-17
Add specs for history slider

xanecs:
This is a rudimentary set of tests for the history slider. (Just testing basic rendering and the arrow buttons)
Ref #813

### DIFF
--- a/spec/components/result-view/history-spec.js
+++ b/spec/components/result-view/history-spec.js
@@ -1,0 +1,47 @@
+/* @flow */
+"use babel";
+
+import React from "react";
+import Enzyme, { shallow } from "enzyme";
+import Adapter from "enzyme-adapter-react-16";
+
+Enzyme.configure({ adapter: new Adapter() });
+
+import store from "../../../lib/store";
+import Kernel from "../../../lib/kernel";
+import History from "../../../lib/components/result-view/history";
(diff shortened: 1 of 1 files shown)


spec/components/result-view/history-spec.js
```diff
@@ -0,0 +1,47 @@
+/* @flow */
```
lgeiger: If you remove this line, CI will go green. Otherwise we'd need to configure `flow` properly.

xanecs: I've noticed that and did a second commit on my repo. Do you know why it doesn't show up in the pull request?

spec/components/result-view/history-spec.js
```diff
@@ -0,0 +1,46 @@
+"use babel";
+
+import React from "react";
+import Enzyme, { shallow } from "enzyme";
+import Adapter from "enzyme-adapter-react-16";
+
+Enzyme.configure({ adapter: new Adapter() });
+
+import store from "../../../lib/store";
+import Kernel from "../../../lib/kernel";
```
lgeiger: I think those lines can be removed too:
```js
import store from "../../../lib/store";
import Kernel from "../../../lib/kernel";
```

PR #1054, opened 2017-10-17
Pr/autocomplete

sigbjorngrini:
Added the config value from autocomplete-plus instead of 3 as min length. Not sure if there should be some if statement that provide 3 as an alternative if it cannot find atom.config.get("autocomplete-plus.minimumWordLength").

### DIFF
--- a/lib/autocomplete-provider.js
+++ b/lib/autocomplete-provider.js
@@ -107,10 +107,10 @@ export default function() {
         return null;
       }
 
-      var minimumWordLength = atom.config.get(
+      let minimumWordLength = atom.config.get(
         "autocomplete-plus.minimumWordLength"
       );
-      if (typeof minimumWordLength != "number") {
+      if (typeof minimumWordLength !== "number") {
         minimumWordLength = 3;
       }
 
--- a/lib/autocomplete-provider.js
+++ b/lib/autocomplete-provider.js
@@ -107,10 +107,14 @@ export default function() {
         return null;
       }
 
-      if (
-        prefix.trim().length <
-        atom.config.get("autocomplete-plus.minimumWordLength")
-      ) {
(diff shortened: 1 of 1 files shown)


lgeiger:
Thanks a lot for the PR!

> Not sure if there should be some if statement that provide 3 as an alternative if it cannot find `atom.config.get("autocomplete-plus.minimumWordLength")`.

That's exactly what `flow` on Travis complains about. You'd need to check if `atom.config.get("autocomplete-plus.minimumWordLength")` is a number and use `3` if that's not the case. Then everything should be green 😄

sigbjorngrini:
Not sure how you want this to be written, but I added ``` var minimumWordLength```. It can be done without defining a new variable also.

lib/autocomplete-provider.js
```diff
@@ -107,7 +107,14 @@ export default function() {
         return null;
       }
 
-      if (prefix.trim().length < 3) {
+      var minimumWordLength = atom.config.get(
```
lgeiger: I'd prefer `let` over `var` here. We tend to use `let` and `const` in the codebase, so it would be nice to keep it consistent.

lib/autocomplete-provider.js
```diff
@@ -107,7 +107,14 @@ export default function() {
...
+        "autocomplete-plus.minimumWordLength"
+      );
+      if (typeof minimumWordLength != "number") {
```
lgeiger: I think a strict equality check with `!==` vs `!=` would be better here.

lgeiger:
@sigbjorngrini Are you interested in helping us maintain Hydrogen? I'd be really happy to add you to the nteract organization.

sigbjorngrini:
Thanks, I can try to look at some solutions to issues, but my js skills are very limited.

PR #1058, opened 2017-10-21
Add option to disable status bar component

### LINKED ISSUE #1055: Add option to hide status bar item
There is limited space down there, and this takes up a large chunk of it. No package should be adding stuff to the status bar without a corresponding option to hide it…

BenRussert:
- Closes #1055

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -93,7 +93,7 @@ const Hydrogen = {
     );
 
     store.subscriptions.add(
-      atom.config.onDidChange("Hydrogen.statusBarDisable", ({ newValue }) => {
+      atom.config.observe("Hydrogen.statusBarDisable", newValue => {
         store.setConfigValue("Hydrogen.statusBarDisable", Boolean(newValue));
       })
     );
--- a/lib/main.js
+++ b/lib/main.js
@@ -93,6 +93,12 @@ const Hydrogen = {
     );
 
     store.subscriptions.add(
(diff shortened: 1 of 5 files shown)


BenRussert:
I'll rethink and see if I can simplify. Thanks for the feedback!

The reason I went with the config store was 1. To fit with the current stateless observer component and 2. Because I think it would be good to have a reusable way to keep track of configs that plays well with mobx.

lgeiger:
>  Because I think it would be good to have a reusable way to keep track of configs that plays well with mobx.

Thanks for explaining your train of thoughts. If you think this is generally useful, I'm happy to merge as is.

lib/main.js
```diff
@@ -93,6 +93,12 @@ const Hydrogen = {
     );
 
     store.subscriptions.add(
+      atom.config.onDidChange("Hydrogen.statusBarDisable", ({ newValue }) => {
```
BenRussert: This should be `observe` instead of `onDidChange` so it initializes. I forgot to push that change, will do so later.

BenRussert:
I think the config store will be helpful for a number of things, for example the way we handle language mappings currently. Also anytime we use a config value within an observer component (now or in the future).

BenRussert:
I think most people will want to leave the status bar on.

By the way here is something else we could add using the observable config map as an example:
zoom-output-font mp4

PR #1062, opened 2017-10-24
Add ability to set watch/output pane history with keyboard

BenRussert:
- Focus the rangeslider handle and press left/right to change the index
set-slider-with-key-focus mp4


BenRussert:
> What about adding the subscription to the History component directly? This way we wouldn't need to duplicate the logic.

I tried this initially, but wasn't sure how to keep the correct output store and rangeslider handle element in sync with the stateless component.

PR #1070, opened 2017-10-29
Set up specs with a mock global store

BenRussert:
- The global store export causes issues between specs
- This could be fixed by exporting the Store class, for spec purposes only,
- the spec here is just an example of how we could test modules independently outside of main.


BenRussert:
I was messing around with specs and ran into issues w/ the global store. This was the best way I found to avoid the issue and make it easier to do independent tests.

This is more of a "look at this request", but maybe it's a good idea?

PR #1076, opened 2017-11-01
A couple quick fixes

BenRussert:
To unrelated one line commits I didnt want to forget about.


lib/code-manager.js
```diff
@@ -5,7 +5,6 @@ import { Point, Range } from "atom";
 import escapeStringRegexp from "escape-string-regexp";
 import _ from "lodash";
 
-import store from "./store";
```
lgeiger: We should improve our linting setup, so things like this will be caught by the linter.

PR #1079, opened 2017-11-02
WIP on simplifying remote kernel docs

maxim-lian:
- Some updates (no need for Docker Toolbox, Dec 2016 bug fixed)
- Some of the way to combining the two docker sections, that are mostly repetitive
- Starting jupyter kernel should be part of the docker process, not something people are `exec`ing in to start
- Everyone should run tini rather than go through the long-winded steps of manually shutting processes

One question: what's the difference between running a notebook and connection to a kernel there, and running `kernelgateway`

### DIFF
--- a/docs/Usage/RemoteKernelConnection.md
+++ b/docs/Usage/RemoteKernelConnection.md
@@ -6,7 +6,7 @@ To connect to a server, you must first add the connection information to the Hyd
 
 ```json
 [{
-  "name": "Remote notebook",
+  "name": "Remote kernel",
   "options": {
     "baseUrl": "http://example.com:8888",
     "token": "my_secret_token"
@@ -48,20 +48,12 @@ jupyter notebook --port=8888
 
 TODO: what is the difference between running the kernel gateway and the notebook?
 
-# Running a Kernel gateway on Docker
+# Running a Kernel gateway using Docker
 
 You can use the same technique to create a kernel gateway in a Docker container. That would allow you to develop from Atom but with all the dependencies, autocompletion, environment, etc. of a Docker container.
 
 **Note**: due to the way that the kernel gateway creates sub-processes for each kernel, you have to use it in a special way, you can't run the `jupyter kernelgateway` directly in your `Dockerfile` `CMD` section. You need to call it with an init manager such as [**tini**](https://github.com/krallin/tini) or run it from an interactive console.
 
-If all you need is a Docker Python environment to execute your code, you can read the section [**Example Jupyter Docker Stack kernel gateway**](#example-jupyter-docker-stack-kernel-gateway) (this method uses **tini** under the hood).
-
-
-## Jupyter Docker Stack kernel gateway
-
-Follow this if you only need to have a simple environment to run commands inside a Docker container and nothing more.
-
-If you need to customize a Docker image (e.g. for web development) follow the section below: [**Example Docker kernel gateway**](#example-docker-kernel-gateway).
 
 ### Dockerfile
 
@@ -144,7 +136,7 @@ docker run -it --rm --name hydro -p 8888:8888 hydro
 - If running locally, you can use `localhost` as the host of your `baseUrl`
 - In Atom, open a Python file, e.g. `main.py`
 - Connect to the kernel you just configured: `ctrl-shift-p` and type: `Hydrogen: Connect To Remote Kernel`
-- Select the kernel gateway you configured, e.g. `docker-kernel`
+- Select the kernel gateway you configured, e.g. `Remote kernel`
 - Select the "type of kernel" to run, there will just be the option `Python 2` or `Python 3`
 - Then select the line or block of code that you want to execute inside of your container
 - Run the code with: `ctrl-shift-p` and type: `Hydrogen: Run`
@@ -186,4 +178,4 @@ markdown.version ['2.6.6']
 ## Terminate the connection and container
 
 - To terminate a running kernel gateway you can "kill" it as any Linux process with `ctrl-c`
-- If you're running the kernel gateway as a Docker container, you can stop the Docker container (this assumes you're using an init manager, such as tini. If you're not, this may not close the process)
+- If you're running the kernel gateway as a Docker container, you can stop the Docker container (this assumes you're using an init manager, such as tini. If you're not, this may not close the process.)
--- a/docs/Usage/RemoteKernelConnection.md
+++ b/docs/Usage/RemoteKernelConnection.md
@@ -77,7 +77,7 @@ ENTRYPOINT [/tini, --]
 # --no-browser & --port aren't strictly necessary. presented here for clarity
 CMD [jupyter-notebook, --no-browser, --port=8888]
 # if running as root, you need to explicitly allow this:
-# CMD [jupyter-notebook, --alow-root, --no-browser, --port=8888]
+# CMD [jupyter-notebook, --allow-root, --no-browser, --port=8888]
 
 ```
 
@@ -91,7 +91,7 @@ services:
   hydro:
     build: .
     entrypoint: [/tini, --]
-    command: [jupyter-notebook, --alow-root, --no-browser, --port=8888]
+    command: [jupyter-notebook, --allow-root, --no-browser, --port=8888]
     ports:
       - 8888:8888
     environment:
--- a/docs/Usage/RemoteKernelConnection.md
+++ b/docs/Usage/RemoteKernelConnection.md
@@ -58,6 +58,7 @@ You can use the same technique to create a kernel gateway in a Docker container.
 - Create a `Dockerfile` based on either one of the [Jupyter Docker Stacks](https://github.com/jupyter/docker-stacks) (recommended), or your own
 - Ensure `jupyter_kernel_gateway` and `tini` are installed, either in the image or as an additional command in the `Dockerfile`
 - Expose the gateway port, in this example it will be `8888`
+- Add an environment variable allowing the Kernel Gateway to get a list of the kernels running
 - Set the `CMD` instruction to starting a Kernel Gateway (though run through an init manager such as `tini`):
 
 ```Dockerfile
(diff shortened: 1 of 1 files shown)


MaximilianR:
This eagerly awaits feedback 

Ready to go from my end after this in answered: 
> One question: what's the difference between running a notebook and connection to a kernel there, and running kernelgateway

lgeiger:
Thanks a lot for the documentation overhaul. I can take a look at this tomorrow :tada:

docs/Usage/RemoteKernelConnection.md
```diff
@@ -1,12 +1,12 @@
...
 
-In addition to managing local kernels and connecting to them over ZeroMQ, Hydrogen is also able to connect to Jupyter Notebook (or Jupyter Kernel Gateway) servers. This is most useful for running code remotely (e.g. in the cloud).
+In addition to managing local kernels and connecting to them over ZeroMQ, Hydrogen is also able to connect to Jupyter Notebook (or Jupyter Kernel Gateway) servers. This is most useful for running code remotely (e.g. in the cloud), or in a Docker container running locally.
 
 To connect to a server, you must first add the connection information to the Hydrogen `gateways` setting. An example settings entry might be:
 
 ```json
 [{
-  "name": "Remote notebook",
+  "name": "Remote kernel",
```
nikitakit: The name actually refers to the server (not to a particular kernel), so better to say something like "Remote server".

docs/Usage/RemoteKernelConnection.md
```diff
@@ -46,74 +46,53 @@ jupyter notebook --port=8888
...
-
-Follow this if you only need to have a simple environment to run commands inside a Docker container and nothing more.
-
-If you need to customize a Docker image (e.g. for web development) follow the section below: [**Example Docker kernel gateway**](#example-docker-kernel-gateway).
 
 ### Dockerfile
 
-- Create a `Dockerfile` based on one of the [Jupyter Docker Stacks](https://github.com/jupyter/docker-stacks).
-- Install `jupyter_kernel_gateway` in your `Dockerfile`
+- Create a `Dockerfile` based on either one of the [Jupyter Docker Stacks](https://github.com/jupyter/docker-stacks), or one you supply
```
willingc: I think it would be best to say:
- Create a `Dockerfile` based on either: 1) one of the Jupyter Docker Stacks (we recommend this option) or 2) your own supplied Docker Stack

docs/Usage/RemoteKernelConnection.md
```diff
@@ -46,74 +46,53 @@ jupyter notebook --port=8888
...
+- Ensure `jupyter_kernel_gateway` and `tini` are installed, either in the image or as an additional command in the `Dockerfile`
 - Expose the gateway port, in this example it will be `8888`
-- Make the command to run be the Kernel Gateway:
+- Make the command to run be the Kernel Gateway (though run through an init manager such as `tini`):
```
willingc: Perhaps reword and simplify.

docs/Usage/RemoteKernelConnection.md
```diff
@@ -46,74 +46,53 @@ jupyter notebook --port=8888
...
 
 ```Dockerfile
 FROM jupyter/minimal-notebook
+# this can also be an image of your own, e.g.:
```
willingc: Perhaps:
```
# If using your own Docker Stack image, use the following `FROM` command syntax substituting your image name
```

MaximilianR:
Should we replace all references to `kernel-gateway` with `jupyter notebook`, then?

docs/Usage/RemoteKernelConnection.md
lgeiger: 👍 for "Remote server"

lgeiger:
@MaximilianR Thanks a lot for your work, this makes it a lot cleaner 🎉 👍 

> Should we replace all references to `kernel-gateway` with `jupyter notebook`, then?

I'd be happy to only recommend `jupyter notebooks`. This gives users less choice and probably makes it easier for new people to understand.

maxim-lian:
@lgeiger I tried using the `jupyter notebook` approach. 

How do you handle the `token`? I get a new one each time I start the process, which is fine when copying & pasting into a browser, but less convenient if it requires updating the config. I also have an SSL issue (using an unsigned certificate?) but that's a separate issue.

That's not needed for the `kernel-gateway`, as there's no authentication (which is fine if running locally).

Do you enter it into the config each time?

maxim-lian:
I've made updates based on the prior feedback, thanks for those

nikitakit:
@maxim-lian To handle tokens, you can [set the notebook server config][1] to always use the same token every time the server starts. You can even set it to the empty string to avoid using tokens altogether (however, this is insecure). If editing the config is inconvenient for a docker setup, I think there's an environment variable such as `JUPYTER_TOKEN` that can be used instead.

[1]: https://github.com/nteract/hydrogen/pull/1079/files#diff-3921662d6df5c3e9793164c8a30b8ff5R39

maxim-lian:
@nikitakit thanks - that worked well.

I've updated the documentation and ready to merge / open to more feedback

docs/Usage/RemoteKernelConnection.md
```diff
@@ -46,75 +46,58 @@ jupyter notebook --port=8888
...
-  hydro-kernel-gateway:
+  hydro:
     build: .
+    entrypoint: [/tini, --]
+    command: [jupyter-notebook, --alow-root]
     ports:
-      - "8888:8888"
+      - 8888:8888
+    environment:
+      # if you can also have this var in your environment it will override this
```
nikitakit: This is a bit unclear because you have 2x "this" and 1x "it", and I'm not sure what they refer to.

docs/Usage/RemoteKernelConnection.md
```diff
@@ -46,75 +46,58 @@ jupyter notebook --port=8888
...
-- Run your container mapping the port of the gateway
-- Give your container a name
-
-```bash
-docker run -it --rm --name hydro-kernel-gateway -p 8888:8888 hydro-kernel-gateway
+EXPOSE 8888
+ENTRYPOINT [/tini, --]
+CMD [jupyter-notebook]
+# if running as root, you need to explicitly allow this:
+# CMD [jupyter-notebook, --alow-root]
```
lgeiger: I think we should add the `--no-browser` flag and maybe even the hardcoded port number to make it clearer for people who want to expose a different port.

I don't have experience with docker (I'm unsure about the placement of the `--port` flag) but something similar to this should do the job:

```Dockerfile
CMD [jupyter-notebook, --no-browser, --port 8888]
# if running as root, you need to explicitly allow this:
# CMD [jupyter-notebook, --alow-root, --no-browser, --port 8888]
```

docs/Usage/RemoteKernelConnection.md
```diff
@@ -46,75 +46,58 @@ jupyter notebook --port=8888
...
 - Create a `docker-compose.yml` file with something like:
 
 ```yml
 version: '2'
 services:
-  hydro-kernel-gateway:
+  hydro:
     build: .
+    entrypoint: [/tini, --]
+    command: [jupyter-notebook, --alow-root]
```
lgeiger: Same es above, I think the following would make it a bit clearer what's happening in the background.

```Dockerfile
command: [jupyter-notebook, --alow-root, --no-browser, --port 8888]
```

docs/Usage/RemoteKernelConnection.md
```diff
@@ -263,19 +141,18 @@ You can test that it is actually working by installing a package in your contain
...
-# Remove in production
-RUN pip install jupyter_kernel_gateway
-EXPOSE 8888
-```
+ENV JUPYTER_TOKEN=my_secret_token # you can also pass this at runtime
 
-**Note**: If you followed the [**Example Jupyter Docker Stack kernel gateway**](#example-jupyter-docker-stack-kernel-gateway) section, your `Dockerfile` will look different. Just make sure you add a line with:
+EXPOSE 8888
+ENTRYPOINT [/tini, --]
+CMD [jupyter-notebook]
```
lgeiger: Same as above:

```Dockerfile
CMD [jupyter-notebook, --no-browser, --port 8888]
```

docs/Usage/RemoteKernelConnection.md
nikitakit: Just saw this: is the correct flag `--alow-root` or `--allow-root`?

maxim-lian: 🤦‍♂️

maxim-lian: 🤦‍♂️

maxim-lian:
Updated!

maxim-lian:
Tests failed, but I think unrelated?

docs/Usage/RemoteKernelConnection.md
```diff
@@ -46,75 +46,60 @@ jupyter notebook --port=8888
...
+ENV JUPYTER_TOKEN=my_secret_token  # you can also pass this at runtime
 
-- Run your container mapping the port of the gateway
-- Give your container a name
+EXPOSE 8888
+ENTRYPOINT [/tini, --]
+# --no-browser & --port aren't strictly necessary. presented here for clarity
+CMD [jupyter-notebook, --no-browser, --port=8888]
+# if running as root, you need to explicitly allow this:
+# CMD [jupyter-notebook, --alow-root, --no-browser, --port=8888]
```
lgeiger: Whoops, still `--alow-root` instead of `--allow-root`.

docs/Usage/RemoteKernelConnection.md
```diff
@@ -46,75 +46,60 @@ jupyter notebook --port=8888
...
 - Create a `docker-compose.yml` file with something like:
 
 ```yml
 version: '2'
 services:
-  hydro-kernel-gateway:
+  hydro:
     build: .
+    entrypoint: [/tini, --]
+    command: [jupyter-notebook, --alow-root, --no-browser, --port=8888]
```
lgeiger: `--allow-root`

maxim-lian:
> the typo is still present

🤕 

...but now updated!

lgeiger:
Thanks for those excellent docs 💯 🎉

lgeiger:
I can't believe we haven't invited you to the nteract organization yet.
Here we go: Welcome to nteract 🎉

maxim-lian:
Thank you! Hope I can be helpful in the near future.

willingc:
Welcome and thanks @maxim-lian :tada:

PR #1110, opened 2017-11-18
Prompt to authenticate with token or cookie

nikitakit:
(This is a proof of concept, don't merge)

The cluster I've been using for work has recently gotten a JupyterHub installation, and I've been trying to connect Hydrogen to it instead of having to run my own notebook servers all the time. However, there are two problems that prevent me from using hydrogen as-is:
* The security policy doesn't allow just setting a token once and forgetting about it. Instead there is a custom authentication procedure that creates a new session every time you log in.
* There's no way to request a token for accessing the notebooks/kernels. The only way to get a session handle is to copy the cookie out of the browser. (I think JupyterHub is working on a better token-auth story than that, but the nature of having someone else run a server for you is that you don't get to control what version of software it's running.)

The code changes here are something I've put together to address these issues. The first change is that instead of specifying `options.token` in a gateway config, you can instead specify `prompt: "token"` to have hydrogen ask you to input a token while connecting to a kernel. The second change also implements an analogous `prompt: "cookie"` option, which lets you copy-paste your session cookie into Hydrogen and have it use that to authenticate.

I'm not really sure where to go next with this code, other than using it for my personal use-case.

Another thing to note is that there is code here to generate same-origin headers for the websocket connection. This used to be a lot harder before the introduction of the `wsFactory` API to `@jupyterlab/services`. I wonder if it would make sense to enable this for all remote connections, not just the cookie-based ones as I have here.

Disclaimer: the code has not been cleaned up, and I haven't run any of the dev tooling (e.g. flow types or style check)

### DIFF
--- a/lib/ws-kernel-picker.js
+++ b/lib/ws-kernel-picker.js
@@ -15,7 +15,8 @@ import InputView from "./input-view";
 import store from "./store";
 
 class CustomListView {
-  onConfirmed: Function = () => {};
+  onConfirmed: ?Function = null;
+  onCancelled: ?Function = null;
   previouslyFocusedElement: ?HTMLElement;
   selectListView: SelectListView;
   panel: ?atom$Panel;
@@ -34,7 +35,10 @@ class CustomListView {
       didConfirmSelection: item => {
         if (this.onConfirmed) this.onConfirmed(item);
       },
-      didCancelSelection: () => this.cancel()
+      didCancelSelection: () => {
+        this.cancel();
+        if (this.onCancelled) this.onCancelled();
+      }
     });
   }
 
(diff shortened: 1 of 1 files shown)


lgeiger:
Thanks a lot for this PR!

I haven't used jupyter hub so I don't know if there is a canonical way of authenticating.

From a users perspective I would prefer hydrogen to conditionally promt for a token or cookie if no token is specified in the settings instead of having to add `prompt: "cookie"` to the `json` settings. Do you think this would be possible?

/cc @willingc

nikitakit:
Here's an updated version of the code that automatically prompts you for a token/cookie when it can't connect. In this case, the decision to prompt you is made based on whether an attempt to get the server to list kernelspecs succeeds, rather than based on whether the `options.token` flag is present.

How does this look? The tradeoff here is:
- If you automatically prompt when no token is specified in the settings, you will force people running tokenless servers to manually specify `token: null` or similar to avoid getting constantly nagged. Also, you would have to change the behavior of automatically cancelling the kernel picker if you input an empty string for the token
- If you prompt when a connection fails, you end up giving spurious prompts when the real issue is that you've specified the wrong server or the server isn't running. This also runs the risk of accidentally submitting your token/cookie to the wrong server if you really mess up your config. (Unfortunately, the error type I get for auth failures is ECONNREFUSED, which is unhelpful in differentiating them from other sources of errors)

lgeiger:
Thanks for continuing the effort! I can take a closer look this weekend.

I had hoped there would be more helpful error messages from jupyter lab/services so we could get around the trade-offs.

nikitakit:
I've dropped the POC label from the title to reflect that the feature is ready for code review. Also, this passes CI now that issues with dependencies that affected all of hydrogen are resolved.

lgeiger:
Sorry for the long wait. I tested it and it works great 🎉 
I think the drawbacks are well worth the improvements. Thanks for your great work!

Would you like to add the new behaviour to our docs?

You may want to take a look at the next version of `jupyterlab/services`: https://github.com/jupyterlab/jupyterlab/pull/3294 it might improve the granularity of the error codes.

nikitakit:
Thanks for reviewing this!

I took a look at the next version of `jupyterlab/services`, and it looks like it's quite a big change. The switch from using XHR to `node-fetch` will mean that I'll have to rewrite a portion of the code for this PR, mainly the parts that deal with cookies and spoofing same-origin requests. I probably won't have the time to get to that for a little while.

(Offhand, do you know if `eval('require')('node-fetch')` will violate Atom's security policy? The new version of services is doing that, but I seem to recall that there are some restrictions to using eval in Atom.)

lgeiger:
> I probably won't have the time to get to that for a little while.

No problem. I just wanted to link https://github.com/jupyterlab/jupyterlab/pull/3294 for the record.

> (Offhand, do you know if eval('require')('node-fetch') will violate Atom's security policy? The new version of services is doing that, but I seem to recall that there are some restrictions to using eval in Atom.)

I'm pretty sure the new CSP allows the usage of eval.

Issue #1118, opened 2017-11-22
something wrong with copy to clipboard


wulidongdong:
Hi, dear devs:
   I think there is something wrong with copy to clipboard function in my hydrogen. when I click the output, it shows that 'copied to clipboard', but when I want to paste this output, comes out nothing. It seems like that '' (empty string) has been copied to clipborad.
OS: windows 10
atom: 1.22.1
hydrogen: 2.1.0
it works well in former version.

BenRussert:
Thanks​ for the report! What type of output are you trying to copy?

wulidongdong:
Thanks for response! and sorry about missing this description.
 print('hello world') in python3.6

rafaelgm:
Well, I was able to pinpoint and fix the problem here.

As I mentioned above about having two machines, one working, one not working, I did this:

1. Copied the **%USERPROFILE%\\.atom\\packages\\Hydrogen** folder from my work PC to my home PC, and now the copy/paste feature works (both the single line output and multiline, where the copy button reappeared)

2. I restored the original folder and then compared this folder with the one from work and noted there were many differences (but on both folders, **package.json** had **"version": "2.1.0"**.

3. Through a careful process on replacing/undo replacing the different files (binary search helped here lol), I was able to find the critical difference in this file:

> %USERPROFILE%\\.atom\\packages\\Hydrogen\\node_modules\\@nteract\\display-area\\lib\\display.js

4. On the left is my version, and on the right is the version running on my work PC. If I transfer those yellow lines to my version and save, the problem disappears.

![image](https://user-images.githubusercontent.com/6760593/33215067-8a0f7724-d115-11e7-945a-d9e2aa84ad64.png)

The package **display-area** here was version **3.1.0**, and at work it was version **3.0.2** (looking at **display-area\\package.json**). For some reason those lines were deleted on the new version and this feature stopped working.

The mystery now (for me) is why the versions on my two computers are different. 😁

BenRussert:
This should be resolved for now, though we will need to refactor our use of `@nteract/display-area`. We can track that on #1116

PR #1124, opened 2017-11-25
Pin display area

BenRussert:
- xref #1118


BenRussert:
@rafaelgm looked into the cause of #1118 and narrowed it down to `@nteract/display-area`. Pinning the version at `3.0.2` will be  a quick fix for now.

lgeiger:
Thanks!

PR #1144, opened 2017-12-15
display-area: Add padding for big outputs

lgeiger:
### Before
bildschirmfoto 2017-12-14 um 20 12 57

### After
bildschirmfoto 2017-12-14 um 20 12 17

Note: The flow failure is unrelated but unfortunately I don't know how to fix it.

### DIFF
--- a/lib/components/result-view/result-view.js
+++ b/lib/components/result-view/result-view.js
@@ -130,7 +130,10 @@ class ResultViewComponent extends React.Component<Props> {
         style={
           isPlain
             ? inlineStyle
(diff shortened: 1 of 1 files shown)


rgbkrk:
Do you want to just merge with the flow failure or introduce `$FlowFixMe`. I'll also take a gander here.

rgbkrk:
You can fix the flow error by switching from `import React from "react"` to `import * as React from "react"`, in _all_ cases. The default export no longer exists in the same capacity.

PR #1161, opened 2017-12-23
Add message when startup cancels due to unsaved file

### LINKED ISSUE #1064: Add a message when kernel startup cancelled due to unsaved file
See #1060, when starting a kernel you need to have your file saved in order to keep track of multiple independent kernels.

Currently the startup just skips without info. We should add some instructions.

BenRussert:
- closes #1064

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -335,7 +335,12 @@ const Hydrogen = {
 
   createResultBubble(editor: atom$TextEditor, code: string, row: number) {
     const { grammar, filePath, kernel } = store;
-    if (!grammar || !filePath) return;
+
+    if (!filePath || !grammar) {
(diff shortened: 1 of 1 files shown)


lib/main.js
```diff
@@ -335,7 +335,12 @@ const Hydrogen = {
 
   createResultBubble(editor: atom$TextEditor, code: string, row: number) {
     const { grammar, filePath, kernel } = store;
-    if (!grammar || !filePath) return;
+
+    if (!filePath || !grammar) {
+      return atom.notifications.addInfo(
```
lgeiger: Should we make it a warning or error instead?

BenRussert: Changed this, please squash!

PR #1186, opened 2018-01-11
Add markdown code blocks

reynoldsnlp:


docs/Usage/Examples.md
```diff
@@ -5,6 +5,7 @@ Please share your favorite snippets with us and add them to this page.
 
 ## Interactive plots using [Plotly](https://plot.ly/api/)
 
+```
 {% codetabs name="Python", type="py" -%}
 from plotly import offline
```
rgbkrk: The reason for this formatting here is for the little language switcher in https://nteract.gitbooks.io/hydrogen/docs/Usage/Examples.html, so I'm not sure if this will affect the rendering in the gitbook.

reynoldsnlp: Ah! That makes sense. As you were... :)

Issue #1245, opened 2018-03-02
Feature request: Get ID of current session


kylebarron:
I often have multiple sessions running in separate code files; is there a way to figure out the identifier for the current session in order to connect with that session in a new file?

I.e. It would be convenient if there were a way to figure out from a file that's connected to one of these two sessions what the ID is, so that I can connect to the right one from a new document.

![image](https://user-images.githubusercontent.com/15164633/36921111-1e819ab4-1e31-11e8-8867-31fdf4ad5f57.png)

nikitakit:
Looks like I'm not the only one running into this. This has actually been bothering me a lot ever since I upgraded Hydrogen around September.

I want to point out that from my perspective (as the person who implemented the first version of the remote kernel connections), the screenshot above is a bug. The session list was designed to always use filenames instead of IDs.

If you click on the "Python 3 | idle" text in the status bar, there's an option called "Rename session". The name given there should be the one displayed in the kernel selector -- with the "rename" command providing additional management flexibility. Also if you open up the notebook in a web browser, it will also always list a meaningful name and not "Session ..."

Unfortunately, there's been a lot of rapid change in 3 distinct pieces of software: Jupyter, the Jupyter services library, and Hydrogen. It now seems like the screenshot above happens for the majority of version combinations, even though originally it was originally designed as fallback code that would only trigger in less than 0.1% of situations.

I'm not really sure why this is the case. I probably won't be able to investigate in detail in the near future, since the precise behavior depends on version combinations. I swear there is at least one version of Jupyter notebook out there that will correctly display filepaths.

kylebarron:
It seems like "rename session" doesn't transfer information back to the Jupyter kernel.
When I click on rename session, it does have filled in 
```
filepath-6d45bdb3-d114-49f8-ba04-082d7990fe0b
```
, but that UUID isn't the same as any of the UUID's listed when I try to connect to a remote session. It seems like the names aren't the same. And when I do rename the session, it doesn't persist for when I try to connect from a new file.

Thanks still, for the info.

nikitakit:
Yes, currently session naming is quite broken.

After taking a quick look, it looks like the fix might actually be quite simple (see #1247).

lgeiger:
Closed by #1247

nikitakit:
@kylebarron I know that this bugfix isn't precisely the feature you were requesting, so please let me know if there's some aspect of your workflow that hasn't been addressed with the latest release.

[...]

PR #1247, opened 2018-03-03
Fix session name display in WSKernelPicker

nikitakit:
This is in response to #1245 

`model.notebook.path` does not exist for kernels spawned from hydrogen, but `model.path` does.

In general, the schema for session model objects has undergone some changes in the past year or so. Once Jupyterlab and `@jupyter/services` get to the point of actually having a stable developer API, we can consider additional changes (such as taking advantage of the `name` field instead of pretending that everything is a path)

### DIFF
--- a/lib/ws-kernel-picker.js
+++ b/lib/ws-kernel-picker.js
@@ -282,7 +282,9 @@ export default class WSKernelPicker {
         });
         const items = sessionModels.map(model => {
           let name;
-          if (model.notebook && model.notebook.path) {
+          if (model.path) {
+            name = tildify(model.path);
+          } else if (model.notebook && model.notebook.path) {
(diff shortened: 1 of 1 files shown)


lgeiger:
> In general, the schema for session model objects has undergone some changes in the past year or so. Once Jupyterlab and `@jupyter/services` get to the point of actually having a stable developer API, we can consider additional changes (such as taking advantage of the `name` field instead of pretending that everything is a path)

`@jupyter/services` is stable and following semantic versioning now. Unfortunately upgrading to `@jupyter/services` requires some changes on our side see #1162.
A great alternative is also [`rx-jupyter`](https://github.com/nteract/nteract/tree/master/packages/rx-jupyter) if someone is interested in looking into that.

PR #1248, opened 2018-03-03
Bump jasmine@3

lgeiger:
For some reason the notebook export test are failing since the upgrade to jasmine 2.
This disables them for now.

/cc  @Madhu94

Closes #1209


BenRussert:
Sorry I didn't see this for a while. I'll merge and see if we can get the test fixed later (or decide if we need it.)

PR #1256, opened 2018-03-09
Support vega 3 and vega-lite 2 transforms

nikitakit:
I know there are some other PRs open regarding transforms, but these changes should be orthogonal.

### DIFF
--- a/lib/components/result-view/transforms.js
+++ b/lib/components/result-view/transforms.js
@@ -6,10 +6,16 @@ import {
   registerTransform
 } from "@nteract/transforms";
 import PlotlyTransform from "@nteract/transform-plotly";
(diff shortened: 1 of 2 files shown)


lgeiger:
Are the vega/altair examples still up to date?
 https://nteract.gitbooks.io/hydrogen/docs/Usage/Examples.html#static-plots

nikitakit:
@lgeiger The altair examples are still up-to-date. It doesn't look like altair supports the newer Vega version, so for now I think the only way to get Vega 3 graphs is to write them manually.

Note that JupyterLab has extensions that support the newer version of Vega, and this updates Hydrogen to match.

lgeiger:
Cool :+1: :+1:

PR #1281, opened 2018-04-07
Hide password input

### LINKED ISSUE #1208: Using getpass to input password brings up input box that uses plaintext.
**Description:**

Using the Python library getpass to input passwords uses an input box that displays in plain text instead of *'s.

**Versions:**

MacOS 10.12.6
Atom 1.23.1
Hydrogen 2.2.2

BenRussert:
- closes #1208
- will work for jupyter >= v5.0

### DIFF
--- a/lib/input-view.js
+++ b/lib/input-view.js
@@ -2,7 +2,12 @@
 
 import { TextEditor } from "atom";
 
-type opts = { prompt: string, defaultText?: string, allowCancel?: boolean };
+type opts = {
+  prompt: string,
+  defaultText?: string,
+  allowCancel?: boolean,
+  password?: boolean
+};
 type cb = (s: string) => void;
 
 export default class InputView {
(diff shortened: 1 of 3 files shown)


BenRussert:
I guess they added a password key to [input requests](http://jupyter-client.readthedocs.io/en/stable/messaging.html#messages-on-the-stdin-router-dealer-channel) for jupyter v5.0+. 

If you want to try this out here is a test snippet:

```python
from getpass import getpass

x = input()
print("Regular input: ", x)
p = getpass()
print("Password input: ", p)
```

lib/input-view.js
```diff
@@ -11,11 +16,16 @@ export default class InputView {
...
-  constructor({ prompt, defaultText, allowCancel }: opts, onConfirmed: cb) {
+  constructor(
+    { prompt, defaultText, allowCancel, password }: opts,
+    onConfirmed: cb
+  ) {
     this.onConfirmed = onConfirmed;
 
     this.element = document.createElement("div");
     this.element.classList.add("hydrogen", "input-view");
+    if (password) this.element.classList.add("password");
```
lgeiger: We could also add the CSS directly to the scroll view, but this should work too.

PR #1291, opened 2018-04-12
Update README.md

qzcool:
Small typo.


n-riesco:
:smile: This is the only case where we actually mean the planet!

Issue #1293, opened 2018-04-13
something wrong with tqdm module


wulidongdong:
hi, devs:
After i update my hydrogen to v2.4.0, the tqdm module doesnt work well. The progress bar does not change during the iteration, only suddenly jump to 100% after iteration done. it works well with hydrogen2.3.0, and my atom crashed couple of times after hydrogen update.

OS: windows10
atom:1.25.1
hydrogen:2.4.0
tqdm:4.22.0
python:3.6.4

thanks.

lgeiger:
I guess this is related to our recent MobX upgrade in #1284

BenRussert:
I agree @lgeiger . @wulidongdong do you have an example code snippet i can use to reproduce?

BenRussert:
I modified a simple example from the `tqdm` docs, adding a sleep call to make it take some time and show progress:

```python
import time
from tqdm import tqdm

for i in tqdm(range(10)):
    time.sleep(1)
    pass

```


Still looking into this, but I've noticed it works in the output area dock, but in regular inline result views.

AstrickHarren:
Hey devs:
Somethings happening again like this. 
The example is not working well:
```python
import time

for i in range(10):
    time.sleep(1)
    print(i)

Regular inline result view or dock view are both not behaving well: they all give the first 1 or 2 output then wait another 8 secs and give out all the output at once. Same thing happens with `tqdm`.

OS: MacOS Mojave 10.14.2
atom: 1.38.2
python: 3.7.0
tqdm: 4.28.1
hydrogen: 2.10.0

thx.

**Comment by jacquespeeters** [comment]
Not sure if you are aware that there are some trouble with Hydrogen Output area and tqdm.

I struggled for a long time to make tqdm work with Hydrogen Output Area. I the following problem when using Hydrogen, interupting tqdm, debugging, then running tqdm loop back.

![image](https://user-images.githubusercontent.com/15957925/77262141-161a3f80-6c94-11ea-8aa5-a05d05aad2e9.png)

From my understanding, it can be solve on Jupyter notebooks by using tqdm_notebook but it doens't work on Hydrogen (it relies on HTML widgets).

Was able to solve my problem by following this addice.
https://stackoverflow.com/a/60067890/5247364

However it throws and error if no instances exists. If you still want to use it with scripts and Hydrogen IDE. Here is my code. 

```
try:
    # Avoids problem on notebooks / Hydrogen IDE
    tqdm.tqdm._instances.clear()
[...]

PR #1294, opened 2018-04-13
Fix calculated kernel for multilanguage

### LINKED ISSUE #1292: Hydrogen kernel shuts down after each command (doesn't stay running as idle)
**Description:**

Hydrogen kernel shuts down after each command (doesn't stay running as idle).  This only happened when I switched to 2.4.0 from 2.3.0. 

**Steps to Reproduce:**

1. Imagine the following file (it's markdown):
```Markdown
```python
t=5+10
print(t)
```
2. Now within Atom, use shift-enter to start the kernel and run the first command. In 2.3.0, this runs and leaves a little "Python 3 | idle" flag in the lower left corner to indicate my kernel is idle but still alive. However, in 2.4.0 (and I tested this on Windows and mac), that tag doesn't stay there and the kernel immediately shuts down. Then when running hydrogen on the next line, the kernel starts up again but has cleared its memory, so t is not defined. I've switched multiple times between 2.3.0 and 2.4.0 to confirm there were no confounding variables. 

**Versions:**

2.4.0

Atom    : 1.25.1
Electron: 1.7.11
Chrome  : 58.0.3029.110
Node    : 7.9.0

**Logs:**

The only minor issue I see that comes up in the console (with debug messages enabled) is that I don't have a default kernel defined:

```
'Hydrogen.kernelspec' could not set the default. Attempted default: undefined; Schema: {}
```
and in 2.3.0, I did select the kernel I wanted, but it didn't restart each time I used shift-enter.

BenRussert:
- closes #1292

### DIFF
--- a/lib/components/result-view/list.js
+++ b/lib/components/result-view/list.js
@@ -2,7 +2,7 @@
 
 import React from "react";
 import { observer } from "mobx-react";
-import { values } from "mobx";
+import { toJS } from "mobx";
 import { Display } from "@nteract/display-area";
 import { transforms, displayOrder } from "./transforms";
 
@@ -44,7 +44,8 @@ class ScrollList extends React.Component<Props> {
         }}
       >
         <Display
-          outputs={values(this.props.outputs)}
+          // $FlowFixMe
+          outputs={toJS(this.props.outputs)}
           displayOrder={displayOrder}
           transforms={transforms}
           theme="light"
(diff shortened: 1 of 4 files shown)


BenRussert:
This will fix #1293 as well. Never should have stopped using `toJS` here, but flow told me to!

lgeiger:
Let's ship a patch release!

Issue #1057, opened 2017-10-19
Give posibility to run different file in existing kernel


Datseris:
With the new update when I run a file a Kernel opens. Okay.

I switch to a different file that may or *may not* be in the same directory as the first one.
When I run it a new kernel opens. This happens for every file.

I literally scanned through the options of Hydrogen. The "directory to start kernel in" is not what I am looking for.

There is no option to allow me to run all files in the **existing kernel** without opening a new one. I would very much like that! 

(I am asking for an option to make this available **by default** by user's choice)

Datseris:
Currently I have to press so many buttons:
First Ctrl Shift P, then "Hydrogen Kernel" then Click hydrogen connect to existing Kernel, then **try to guess** which one is the kernel I want, since the popup window is not wide enough to read the full path AND it is not scrollable with Atom Material theme (bug).

MaximilianR:
Is there a reason to have a kernel per file? 

This makes my workflow much tougher - I'm copying and pasting code into a file to run it in the existing kernel!

Datseris:
Same for me.

BenRussert:
I think there is definitely a happy medium to be worked out on this as the 2.0 release is only a couple weeks old. I still think independent kernels are most intuitive and should be the default, but I'm hoping ideas from this thread result in making this better through config or other enhancements. I'm interested in working on that or helping anyone else who wants to contribute.

> Is there a reason to have a kernel per file?

In my opinion (and I think this is a matter or preference), the reason each file should start its own kernel by default is because if you want your file to run outside of hydrogen, say in a script or an app, you'd have to pay attention to namespaces and what you import/require/using etc. It's also how jupyter notebook works unless additional setup is done to connect to existing kernels.

Datseris:
but it is not how Spyder works. For example I switched from spyder to atom.

I am not asking to change the default, but at least giving me the option to
recover the behaviour I need should be really easy...

MaximilianR:
I agree it's not a complete unadulterated slam dunk and there are multiple perspectives...

I think a couple of changes we could make that would help everyone:
- An option to have a global kernel
- A way of setting a default kernel. I use a docker setup, and in order to get a kernel running in a new file, I need to run through 4-5 prompts (rather than Cmd-Enter if using a local kernel). Much of the frustration of the multiple kernels is around getting any kernel set up, not necessarily having the same kernel in that file

lgeiger:
Checkout #40 for more information about this topic.

[...]

PR #1303, opened 2018-05-03
Global kernel mode

BenRussert:
See #1057 , this PR introduces a config to use a single global kernel between files of the same grammar.

For now you have to restart if you change globalMode after hydrogen has been activated.

To try it out, checkout this branch then set global mode with: `atom.config.set("Hydrogen.globalMode", true)` (or in settings view)

### DIFF
--- a/lib/store/index.js
+++ b/lib/store/index.js
@@ -148,6 +148,12 @@ export class Store {
   updateEditor(editor: ?atom$TextEditor) {
     this.editor = editor;
     this.setGrammar(editor);
+
+    if (this.globalMode && this.kernel && editor) {
+      const fileName = editor.getPath();
+      if (!fileName) return;
+      this.kernelMapping.set(fileName, this.kernel);
+    }
   }
 
   @action
--- a/lib/store/index.js
+++ b/lib/store/index.js
@@ -31,9 +31,18 @@ export class Store {
   @observable editor = atom.workspace.getActiveTextEditor();
   @observable grammar: ?atom$Grammar;
   @observable configMapping: Map<string, ?mixed> = new Map();
+  globalMode: boolean = Boolean(atom.config.get("Hydrogen.globalMode"));
 
   @computed
   get kernel(): ?Kernel {
+    if (this.globalMode) {
+      if (!this.grammar) return null;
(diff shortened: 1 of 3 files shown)


lib/store/index.js
```diff
@@ -31,9 +31,18 @@ export class Store {
   @observable editor = atom.workspace.getActiveTextEditor();
   @observable grammar: ?atom$Grammar;
   @observable configMapping: Map<string, ?mixed> = new Map();
+  globalMode: boolean = Boolean(atom.config.get("Hydrogen.globalMode"));
```
BenRussert: This mode won't change while hydrogen is activated. I was afraid live updating would cause too many problems and most people will keep this one or the other most of the time anyway

lib/store/index.js
```diff
@@ -31,9 +31,18 @@ export class Store {
...
 
   @computed
   get kernel(): ?Kernel {
+    if (this.globalMode) {
+      if (!this.grammar) return null;
+      const currentScopeName = this.grammar.scopeName;
+      return this.runningKernels.filter(
+        k => k.grammar.scopeName === currentScopeName
+      )[0];
+    }
```
BenRussert: Jasmine was failing when I originally used simple `k.grammar === this.grammar` this made it work.

lib/store/index.js
```diff
@@ -139,6 +148,12 @@ export class Store {
   updateEditor(editor: ?atom$TextEditor) {
     this.editor = editor;
     this.setGrammar(editor);
+
+    if (this.globalMode && this.kernel && editor) {
+      const fileName = editor.getPath();
+      if (!fileName) return;
+      this.kernelMapping.set(fileName, this.kernel);
+    }
```
BenRussert: Files with an associated kernel still need to be tracked in `kernelMapping` for things like Kernel Monitor view. This ensures all editors with kernels associated are tracked.

lgeiger:
Thanks for tackling this! I'll check it out tomorrow and don't merge it right now since you've added a WIP label.

BenRussert:
Thanks @lgeiger. I removed the WIP tag.

PR #1306, opened 2018-05-04
kernelspecs sort by display_name

arl-o:
To avoid inconsistent ordering of kernels, and allow users to dictate which kernels appear first, sort the kernel specs by their display_name.

I generate a lot of kernels, but current behaviour lists default kernels (eg `Python 3`) last, which is inconvenient.


BenRussert:
Thanks for the PR! Looks good to me.

Issue #1341, opened 2018-06-20
Hyrogen Run not working correctly with if/else statements in python


mlangiu:
Pictures say more than words... 

**When attempting to run** the following scenarios (i.e. execute the selected lines by hitting <kbd>shift</kbd> + <kbd>enter</kbd>) we'l run into some issues:

![grafik](https://user-images.githubusercontent.com/15340151/41656249-a8a0a7bc-748f-11e8-9323-368da0c97553.png) only runs the `if` part: ![grafik](https://user-images.githubusercontent.com/15340151/41656285-bf5368c8-748f-11e8-9618-941c0ee5fdb4.png)
This is unfortunate but could be considered a beauty problem, as
![grafik](https://user-images.githubusercontent.com/15340151/41656354-ed06f19a-748f-11e8-8414-72c8c49696e8.png) circumvents this issue: ![grafik](https://user-images.githubusercontent.com/15340151/41656379-0572171e-7490-11e8-88e4-e4b87e210abb.png)

However, within indented sections (where usually this is possible: 
![grafik](https://user-images.githubusercontent.com/15340151/41656455-53f153d2-7490-11e8-964b-be69a3a28594.png) -> ![grafik](https://user-images.githubusercontent.com/15340151/41656496-7270fccc-7490-11e8-865c-58574c9db73a.png))
This: ![grafik](https://user-images.githubusercontent.com/15340151/41656630-f66ce270-7490-11e8-911f-4ef740533b30.png) will now also fail: ![grafik](https://user-images.githubusercontent.com/15340151/41657039-4425e9e8-7492-11e8-99f8-abfde18dcae2.png)

      File "<tokenize>", line 3
        else:
        ^
    IndentationError: unindent does not match any outer indentation level

**Versions:**

OS: Win 10 Enterprise
Atom    : 1.27.2
Electron: 1.7.15
Chrome  : 58.0.3029.110
Node    : 7.9.0

mbroedl:
You may want to try this package, which takes care of the whitespace problem.
https://atom.io/packages/hydrogen-python

And then you may want to file the if/else behaviour as a feature request to that package, because it's a language specific issue?
#862 #1285

kylebarron:
This specific issue should have been fixed from https://github.com/nteract/hydrogen/pull/1363 when selecting indented if/else blocks. For automatic if/else completion, see hydrogen-python.

Issue #1285, opened 2018-04-09
Unable to run code blocks that are already indented


tobiasbj:
**Description:**

Unable to run code blocks that are already indented.
Example is shown below.

**Steps to Reproduce:**

1. Copy paste this code.
```
for i in range(10):  # This is the line you ignore
    test = "test"
    if test is None:
        print("yes")
    else:
        print("no")
```

2. Mark and run all lines except the initial one.

**Versions:**

Windows 10
Atom 1.25.1
Hydrogen 2.3.0

**Logs:**
Error output
```
File "<ipython-input-1-ec679601011f>", line 2
    if test is None:
    ^
IndentationError: unexpected indent
```

tobiasbj:
Ah, cool. No I haven't tried hydrogen-python yet. I'll try it now 👍

BenRussert:
Looks like this was fixed in #1363, try this again after you update hydrogen. Feel free to reply if you still face the issue after updating.

Issue #1194, opened 2018-01-17
Feature request: Run multiple selections


kylebarron:
If multiple selections are highlighted, run each of them sequentially. I.e. instead of:

![image](https://user-images.githubusercontent.com/15164633/35060634-0c2929de-fb8d-11e7-928e-6256f56a55f4.png)

It would be:
![image](https://user-images.githubusercontent.com/15164633/35060661-1cd7fc2e-fb8d-11e7-8c8c-cb4e6751b229.png)

kylebarron:
If I were to submit a PR for this, would this be considered? Or is this behavior not desired?

BenRussert:
Hi @kylebarron, sorry for the late response.

I just realized it looks like you have multiple cursors in your screenshot, something we don't currently handle. If this is something you would like to improve, I would be interested in a way to handle that if you sent a PR.

Otherwise, my take is that for individual cursor selections the way things work now are consistent with Jupyter platforms that separate output by cells. The intended way to get multiple outputs in hydrogen is to include cell markup in your code file. 

https://nteract.gitbooks.io/hydrogen/docs/Usage/GettingStarted.html#hydrogen-run-cell

kylebarron:
@BenRussert As an example, I often want to debug functions for which I have default values, and my ideal way of initially defining those values would be like this:

![peek 2018-07-16 16-41](https://user-images.githubusercontent.com/15164633/42782519-cf50a82e-890e-11e8-909a-58c21879762a.gif)

This isn't really a situation where `run-cell` would be useful, because only occasionally do I want to run the multiple selections, and it would seem that `run-cell` would run the entire function, not the localized selections.

I'm guessing that the best way to implement this would be by changing `codeManager.findCodeBlock()`?
https://github.com/nteract/hydrogen/blob/9d595258c9619b3b106b796c7b1305e3109bcf9e/lib/code-manager.js#L301-L312

PR #1363, opened 2018-07-30
Don't trim code selection before sending to kernel

### LINKED ISSUE #862: Request: strip _common_ leading whitespace from input before passing to kernel

kylebarron:
### Description of the Change

Don't remove whitespace from start and end of string before sending to kernel. Would fix #862.

I spent several hours trying to figure out enough JavaScript to remove common leading whitespace (that's on [this branch](https://github.com/kylebarron/hydrogen/tree/strip_common_whitespace), and I'd submit that as a PR if preferred) when I realized that the IPython kernel can deal with common whitespace just fine:

image

The issue is that Hydrogen trims strings by default before sending them to kernels.
https://github.com/nteract/hydrogen/blob/ad6217f61131a5a8e8bf3134f10a764b168d0184/lib/code-manager.js#L15-L20

So what IPython actually receives is:
image

### Alternate Designs

I wrote code that actually removes common leading whitespace. That's on [this branch](https://github.com/kylebarron/hydrogen/tree/strip_common_whitespace). 

The hydrogen-python extension package tries to deal with this but has some [limitations](https://github.com/nikitakit/hydrogen-python#limitations) that prevent its use.

### Benefits

Selection-based Python code would work regardless of the amount of common indentation.

### Possible Drawbacks

Other kernels might be expecting strings without any leading or trailing whitespace. I would imagine that kernels would account for extra whitespace if necessary and that it wouldn't be an issue.

Otherwise, to preserve true backwards compatibility we could have only the Python kernel not trim whitespace. Or have a user-defined config setting. Maintainers probably have the best idea for what is best.

### Applicable Issues

Would fix #862, half of #1341, #1285.

### Examples

image
image

(This is the same behavior as the IPython kernel)
image

image

### DIFF
--- a/lib/code-manager.js
+++ b/lib/code-manager.js
@@ -14,7 +14,7 @@ import {
 
 export function normalizeString(code: ?string) {
   if (code) {
-    return code.replace(/\r\n|\r/g, "\n").trim();
+    return code.replace(/\r\n|\r/g, "\n");
   }
   return null;
 }


kylebarron:
@nikitakit

kylebarron:
Example to show concordance with original comment on 862:

> Note that in my example if I were to start my selection at the I character in Imax then the behavior should be to still throw an error because the first line in my selection doesn't have leading whitespace in common with all other lines.

image

nikitakit:
@kylebarron It's great that you're taking a look at this!

Regarding the patch here:

1. Is there any situation where it would still make sense to strip whitespace from the *end* of the string? I can't think of one off the top of my head.
2. It might still be useful to strip whitespace when there is only one line of non-whitespace that's being run. I'm not sure how other kernels would react to single-line statements that are indented. That said, I may not have the best idea of what other kernels do because I mostly use Python.
3. I'm OK with failing to run incomplete selections missing the indentation on the first line. I also wouldn't be opposed to adding a workaround that implicitly expands the selection to the start of the line, if that would add whitespace characters only.
4. Have you tested the regular "run" command, without any selections active? Ideally if you place the cursor on an "if condition:" line and hit "run", it should run the entire if statement. (Issues regarding "else" clauses are TBD)

I'd be happy merging this, assuming there aren't known regressions for other languages.

I see this as a high-priority issue because it's quite annoying and affects a very popular language. If problems come up with other languages we can make this behavior grammar-dependent. My attempts to solve this exclusively in a plugin have all run into insurmountable difficulties, so I'm very much in favor of addressing this in hydrogen itself.

kylebarron:
1. I don't know. There are no situations that I can think of, but it would be just as easy to change `.trim()` to `.trimRight()`.

2. I don't know how other kernels would react either. I did some very basic testing with the ijavascript, R, and Julia kernels. The results were the same no matter if the text had leading indents or not.

3. I think it might be difficult to try to figure out how much indentation to add to the first line. And it might be confusing for users. I would hope that if they didn't include initial whitespace in their selection and received an `IndentationError`, that they would conclude that they need to include the original whitespace. As it stands, Hydrogen sends the kernel the same thing that would be sent with a copy-paste into IPython in the terminal, and I think that's a plus for consistency in what Hydrogen runs.

4. Yes. It seems that the current behavior of Hydrogen is to include any code that is more indented on following lines, and that hasn't changed. Pressing <kbd>cmd</kbd>-<kbd>enter</kbd>:
image

mbroedl:
That's a good one @kylebarron!

Looking quite good! Only thing I could find right now is that a mix of hard tabs and spaces causes Indentation Errors, but that's not supposed to be working in python anyway:
image
These two lines should help in case you feel like implementing that:
```
const tabSub = ' '.repeat(e.getTabLength());
code = code.replace(/\t/g, tabSub);
```

Another point: I would suggest moving the implementation to `normalizeString` as every function seems to call that before returning. That'd avoid unforeseen consequences in case getTextInRange, getRow, and getRows are called. (I don't know what the circumstances for this would be.)

RE: 3. I'm in favour of expanding the selection to the front of the line, assuming that there is only whitespace at the beginning of the line.


Only problem is that from the code now passed to `hydrogen-python` is that you cannot anymore expand the selection because you don't know the indentation level. This would then indeed require another API function in hydrogen. Since the whitespace is now stripped even _before_ the loop whether or not there is a selection, I'm not sure where to put the API anymore... unless you pass the indentation-level as well?

kylebarron:
@mbroedl 

> Only thing I could find right now is that a mix of hard tabs and spaces causes Indentation Errors, but that's not supposed to be working in python anyway

If the mix of hard tabs and spaces confuses Python anyways, I don't think we should fix that on Hydrogen's end. Otherwise the file will work with hydrogen but not with `python file.py`, right?

> I would suggest moving the implementation to `normalizeString`

[The implementation _is_ in `normalizeString`.](https://github.com/nteract/hydrogen/pull/1363/files)

> Only problem is that from the code now passed to hydrogen-python is that you cannot anymore expand the selection because you don't know the indentation level.
>
> Since the whitespace is now stripped even before the loop whether or not there is a selection

Maybe I'm confused... The point is that _no whitespace_ is being stripped with this change. If you select both lines and run this:
```py
    print('hello world')
    print('foobar')
```

the text that Hydrogen spits out is exactly:
```
    print('hello world')\n    print('foobar')\n
```

This should make it _easier_ to find the indentation level, since you can simply get the indentation before the first non-whitespace character.

mbroedl:
You're absolutely right @kylebarron . Sorry, I looked at the wrong fork, as you pointed out in the other thread!
Then I'm game with the PR, looking good!

kylebarron:
@mbroedl Sorry! I was just trying to juxtapose them!

nikitakit:
Re my point 3: Let's leave this for now to try to get this fix out faster. I want to clarify, however, that the original suggestion was not to infer the leading whitespace, but rather to read it directly from the text editor (and decide to include it or not based on whether there are any non-whitespace characters in same line as the start-of-selection)

Re mixing tabs and spaces: If the whitespace in the original file is a mess, I think the correct action is for the programmer to fix it (Atom has some whitespace normalization commands, I think). I don't see a reason to do this in Hydrogen.

Re my point 2: this is still the main one I'm worried about in terms of possible regressions. I know it won't be a problem for Python, but I have no idea how other kernels handle things. But that doesn't mean this PR has to be changed, unless an actual issue is discovered.

It looks like you already tested a few of the most popular kernels with this change, and I don't have any others in mind that might be problematic. Perhaps it makes sense to sanity check on multi-syntax documents though, since those use a different code path.

Does anyone want to see a particular kernel tested, or have other comments on this PR? Personally I'm in favor of merging now, letting it get tested on master for a little while, and then releasing.

n-riesco:
@kylebarron @nikitakit The use of `.trim()` was an unintended change during Hydrogen's transition from Coffeescript to ES6 (see #503). It's not there to fix any issues with a kernel (this is unlike the transformation to use unix line-endings (see #451) that was introduced to make IRKernel happy).

My view is that, as white-space has semantic meaning for some kernels,  white-space transformations should live in language-specific plugins (like [hydrogen-python](https://github.com/nikitakit/hydrogen-python)).

My :+1: to merging this PR as is.

BenRussert:
Thanks for the PR @kylebarron! Would you like to help us maintain hydrogen?

kylebarron:
I'm interested, though at this point my js knowledge is still very weak, and I know nothing about React.

In the next couple weeks I may try to create a PR for multiple cursors https://github.com/nteract/hydrogen/issues/1194, which is an enhancement I'd be happy to have.

PR #1378, opened 2018-08-10
disable the status bar & config test for now

rgbkrk:
I'm thinking that this test is not a red herring -- something is up with either:

* The Hydrogen Store
* An Atom API
* Enzyme (they just had a new release)


BenRussert:
Thanks!

PR #1379, opened 2018-08-10
Fix status-bar test

BenRussert:
This is actually how I originally meant to write this test :grimacing:


rgbkrk:
That makes way more sense!

PR #1380, opened 2018-08-12
Fix kernel restart issue

BenRussert:
- this PR fixes a bug that prevents kernel restart from working properly.
- spawnteract will remove the connection file on process exit by default as of v5.0.0
- we need to handle this manually right now since we recycle the connection file between restarts


rgbkrk:
:tada: 👾

PR #1384, opened 2018-08-13
Check text for undefined

abendafb:
With: @lsaddan

This happened when we tried connecting hydrogen to a tweaked Jupiter server.
It worked nicely with this change.

### DIFF
--- a/lib/store/output.js
+++ b/lib/store/output.js
@@ -57,7 +57,7 @@ export function reduceOutputs(
 export function isSingeLine(text: string, availableSpace: number) {
   // If it turns out escapeCarriageReturn is a bottleneck, we should remove it.
   return (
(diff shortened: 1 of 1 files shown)


lib/store/output.js
```diff
@@ -57,7 +57,7 @@ export function reduceOutputs(
 export function isSingeLine(text: string, availableSpace: number) {
```
BenRussert: Since you have discovered a bug that allows `undefined` values to sneak in, we should change this `text: string` to a  [flow maybe type](https://flow.org/en/docs/types/maybe/), which looks like this: `text: ?string`. 

The  `?string` type will tip `flow` off that it is possible for the argument `text` to be `undefined`. Then `flow` will warn us that we need to handle the case where `text === undefined`.

abendafb:
Fixing null case as well

abendafb:
BTW, is it is**Singe**Line on purpose?

PR #1385, opened 2018-08-15
Upgrade flow

BenRussert:


BenRussert:
I took a crack at the $FlowFixMe's around the kernel types, but it felt like I was making things worse. Hoping you'll take a look @nikitakit!

PR #1389, opened 2018-08-21
Add class to status bar item

danielbayley:
Add .hydrogen class to distinguish it and prevent clashing with other items in the status bar.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -255,7 +255,7 @@ const Hydrogen = {
 
   consumeStatusBar(statusBar: atom$StatusBar) {
     const statusBarElement = document.createElement("div");
-    statusBarElement.className = "inline-block";
+    statusBarElement.classList.add("inline-block", "hydrogen");
 
     statusBar.addLeftTile({
       item: statusBarElement,


BenRussert:
Hi, sorry to leave you hanging here. It's harder for me to review styling PRs because I don't know what side effects there might be without trying it out.

Will this change anything for hydrogen? Or just make it more friendly to other status bar elements?

danielbayley:
> I don't know what side effects there might be

@BenRussert All it does is add an extra class to the selector already under `.status-bar`; so I can't see how there could be any side effects… unless there is another package called hydrogen that puts things in the status bar with that class selector (and apm packages are namespaced…)

> Or just make it more friendly to other status bar elements?

It changes absolutely nothing for Hydrogen… this is just so it will play nice with other packages that put things in the status bar. Everyone seems to use it as a dumping ground without any thought to what else is in there! The main reason for this PR is so Hydrogen doesn't mess up the order of thing in my status bar (using https://atom.io/packages/move-status-items).

BenRussert:
Cool, thanks!

Issue #1406, opened 2018-09-11
Uncaught Error: Cannot find module '@babel/runtime-corejs2/helpers/interopRequireWildcard'


yburatti:
**Description:**
Hi there,
I installed Hydrogen on Atom using the command line (apm install hydrogen) as it throws an error directly from the package manager. After installation I open Atom to run some test code (eg, print("hello")). Even before I try to "shift-Enter", the error pops up. Unable to run any code with Hydrogen "shift-enter", running the file as a script works fine though.
Spend a bit of time looking around ( I rarely write help posts), and the closest I found was the same type of error message from meteor: [Dates from April 2018]
https://forums.meteor.com/t/solved-cannot-find-module-babel-runtime-helpers-builtin-objectspread-after-update-meteor-to-1-6-1-1/43034

The problem and solution there was:

"The meteor-babel npm package has been updated to version 7.0.0-beta.42, which may require updating any custom Babel plugins you’ve enabled in a .babelrc file, and/or running the following command to update @babel/runtime
meteor npm install @babel/runtime@latest"

I wonder if it's the same issue with Hydrogen, and if there is a fix for it or if I'm doing something wrong.

Help :)

Thanks

PS: In troubleshooting guide, it says to add debug messages by ticking enable debug message.
I don't have that tick in the settings (picture attached).
![image](https://user-images.githubusercontent.com/4509592/45332076-dc9e4980-b5b1-11e8-944c-8ab0d1607593.png)

**Steps to Reproduce:**
(Hard to know since it happens when I open Atom after installation)
1. Install Hydrogen with command line
2. Open Atom
3. Error will pop-up

**Versions:**

**Atom**: 1.30.0 x64
**Electron**: 2.0.5
**OS**: Microsoft Windows 10 Enterprise
**python**: 3.6

**Logs:**
### Stack Trace

Uncaught Error: Cannot find module '@babel/runtime-corejs2/helpers/interopRequireWildcard'

```
At module.js:545

Error: Cannot find module '@babel/runtime-corejs2/helpers/interopRequireWildcard'
    at Module._resolveFilename (module.js:543:15)
    at Module._resolveFilename (~/AppData/Local/atom/app-1.30.0/resources/electron.asar/common/reset-search-paths.js:35:12)
    at Function.get_Module._resolveFilename (~/AppData/Local/atom/app-1.30.0/resources/app/src/module-cache.js:354:58)
    at Module.require (/app.asar/static/index.js:40:43)
[...]

PR #1407, opened 2018-09-11
Add link to data-explorer plugin readme

BenRussert:


BenRussert:
Looks like a dependency upgrade broke the build, see #1406. I'm seeing if there is a quick fix now.

rgbkrk:
Since #1410 fixes things up, I went ahead and merged this. I'm also going to put out a patch release for all our packages that includes the babel core dep explicitly.

Issue #1366, opened 2018-08-01
Python - OLS Regresson results summary - Text Alignment issues on hydrogen output window


sjvaidya:
**Description:**
Text Alignment issues with OLS Regression result summary in hydrogen-output. Things get messy to interpret without alignment. Please refer two screenshot for more details.

Note:- This works fine when the results are displayed inline

Code:

import pandas as pd
import numpy as np
import statsmodels.api as sm

data_train = pd.read_csv('train.csv')
data_train.columns.values

y = data_train[['SalePrice']]
X = data_train[['GrLivArea'] + ['TotalBsmtSF']]

X1 = sm.add_constant(X)

print(sm.OLS(y,X1).fit().summary())

**Steps to Reproduce:**

1. Run python code to Display OLS Regression results summary inline
2. From settings, enable hydrogen output window
2. Run python code to Display OLS Regression results summary again
3. See the text/data alignments in both cases.

**Versions:**

Which OS and which version of Hydrogen and Atom are you running?
Win10, Atom 1.28.1 x64
![hydrogen-inline-results](https://user-images.githubusercontent.com/29955169/43519558-8d482186-95ad-11e8-8442-b1c52dace5d1.png)
![hydrogen-output-text-misaligned](https://user-images.githubusercontent.com/29955169/43519559-8d73d68c-95ad-11e8-84ef-3bdd6cb01470.png)

bigheadming:
I have the same problem when using show command in pyspark. Can you please fix it, or allow the user to choose the font-family in output area? Thanks.

peopleDF.show(truncate=False)

+-----------------------+-------+------------------------------------+
|cities                 |name   |schools                             |
+-----------------------+-------+------------------------------------+
|[palo alto, menlo park]|Michael|[[stanford, 2010], [berkeley, 2012]]|
|[santa cruz]           |Andy   |[[ucsb, 2011]]                      |
|[portland]             |Justin |[[berkeley, 2014]]                  |
+-----------------------+-------+------------------------------------+

BenRussert:
This needs a stylesheet tweak. This is a great first issue and if one of you decide to look into it you probably will not even need an understanding of javascript to take this on.

I will answer any questions I can and would be so grateful if you took a crack at fixing this :bowing_man: 

[...]

Issue #1288, opened 2018-04-11
Q: Fixed font in hydrogen output area 


dsanalytics:
Anyone knows how to change/config hydrogen output area font? Being able to specifying font size as well would be great. Thanks.

BenRussert:
There is actually an "Output area font size" in the hydrogen config. Try that out and see if it's what you're looking for.

dsanalytics:
I'm looking to change font 1st and foremost - e.g. to fixed font so that formatted numbers line up nicely across rows.

P.S. I could not find that font size option - screenshot would be appreciated.

BenRussert:
![screenshot from 2018-05-03 05-41-46](https://user-images.githubusercontent.com/10860657/39574534-d82328b4-4e94-11e8-86b2-28c68d5fe7d1.png)

dsanalytics:
Thanks. Handy option, but I need an option to change font too.

Issue #894, opened 2017-06-24
Insufficient contrast when running code with error


domainedma:
**Description:**

Insufficient contrast with dark color schemes when running code with error.

<img width="763" alt="color_scheme" src="https://user-images.githubusercontent.com/11057863/27504174-b9b16bf6-58c8-11e7-99d1-7896ca3b087a.png">

Error messages should use the Atom syntax scheme or use a white background or other background color to increase contrast.

**Steps to Reproduce:**

1. Set Atom UI theme to One Dark and Atom Syntax theme to Atom Dark (or other dark theme like in https://nteract.gitbooks.io/hydrogen/docs/Usage/GettingStarted.html)
2. Run Python code with error in it

```
def permute(string_so_far, string_rest):
    if len(string_rest) <= 0:
        print(string_so_far)

    for i in range(len(string_rest)):
        print(i)
        remaining_string = string_rest[0, i] + string_rest[i + 1]
        permute(string_so_far + string_rest[i], remaining_string)

permute('', 'hello')
```

<img width="1021" src="https://user-images.githubusercontent.com/11057863/27504209-0c438b4c-58c9-11e7-981b-be933e697942.png">

**Versions:**

Which OS and which version of Hydrogen and Atom are you running?

Atom    : 1.18.0
Electron: 1.3.15
Chrome  : 52.0.2743.82
Node    : 6.5.0

You can get this information from copy and pasting the output of `atom --version`  from the command line.

**Logs:**

N / A

Please post any error logs and the output of the developer tools as described in our [Debugging Guide](https://nteract.gitbooks.io/hydrogen/docs/Troubleshooting.html).

n-riesco:
IPython has an option to configure the colour scheme, see `InteractiveShell.colors` in http://ipython.readthedocs.io/en/stable/config/options/kernel.html .

rgbkrk:
I'd like to have a way to override these for the sake of nteract as well, similar to how it's done in iTerm2. We currently just interpret ANSI codes as is for `ansi-to-react`.

lgeiger:
Part of the problem here is that we cannot easily get the colors used by the current Atom theme (at least I don't know of any good way).

We only have access to the [official syntax colors](https://github.com/atom/one-dark-syntax/blob/master/styles/syntax-variables.less).

BenRussert:
Could/does `ansi-to-react` optionally return w/ class names instead of color style? 

[...]

PR #1424, opened 2018-09-24
Use monospace font in output area

kylebarron:
Fixes https://github.com/nteract/hydrogen/issues/1366.
Fixes https://github.com/nteract/hydrogen/issues/1288.

This submits the fix found by @bigheadming in https://github.com/nteract/hydrogen/issues/1366. He also raised the issue of bad contrast with output colors, but that is a separate issue (https://github.com/nteract/hydrogen/issues/894).

### DIFF
--- a/styles/hydrogen.less
+++ b/styles/hydrogen.less
@@ -251,6 +251,7 @@ svg:first-child {
     flex: 0 0 auto;
   }
   .multiline-container {
+    font-family: monospace;
     flex: 0 1 auto;
   }
   .scroll-list {


BenRussert:
Cool, thanks!

joolius:
monospace is ugly. now every time I update hydrogen I have to outcomment that line... can we make it an option?

aviatesk:
If you can accept something bad described in #1366, you can put something like below into your stylesheet:

```less
.hydrogen {
  .output-area .multiline-container {
    font-family: var(--editor-font-family);  // Or the font name you want to use in output pane
  }
}
```

joolius:
Beautiful solution to an ugly problem, thanks!

PR #1440, opened 2018-10-06
Add a spec for ResultViewComponent

hpierce1102:
#813

This adds a basic test for ResultViewComponent. I was hoping to get farther and write more useful tests, but ran into issues with mocking some internals of the class.

For instance, I wanted to demonstrate the clicking on the topmost `<div>` actually ran `openInEditor`, but struggled to mock `openInEditor`. I also wanted to demonstrate that clicking `icon-clippy` actually copied stuff to the clipboard, but figuring out how to make that actually button appear for a test was outside of my React skills (I got to a point where it looked like it had something to do with`ref = { ref => {]}`) 😕.

I'd be more than happy if someone had some advice to offer in terms of adding more value to these tests. Or if someone thinks this is useful enough to merge without more tests - I suppose that'd be fine too.

### DIFF
--- a/spec/components/result-view/result-view-spec.js
+++ b/spec/components/result-view/result-view-spec.js
@@ -1,0 +1,72 @@
+"use babel";
+
+import React from "react";
+import { shallow } from "enzyme";
+
+import ResultViewComponent from "../../../lib/components/result-view/result-view";
+
+describe("ResultViewComponent", () => {
+  it("should render", () => {
+    const mockStore = {
+      index: 0,
+      outputs: [{}, {}],
+      status: "",
+      executionCount: 0,
+      position: {
+        lineHeight: 0,
+        lineLength: 0,
+        editorWidth: 0,
+        charWidth: 0
+      }
+    };
+
+    const mockKernel = {};
+
+    const destroy = () => {};
+
+    const component = shallow(
+      <ResultViewComponent
+        store={mockStore}
+        kernel={mockKernel}
+        destroy={destroy}
+        showResult={true}
+      />
+    );
+
+    expect(component.type()).not.toBeNull();
+  });
+
+  it("should destroy itself when x is clicked.", () => {
+    const mockStore = {
+      index: 0,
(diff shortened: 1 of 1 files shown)


spec/components/result-view/result-view-spec.js
```diff
@@ -0,0 +1,72 @@
...
+        editorWidth: 0,
+        charWidth: 0
+      }
+    };
+
+    const mockKernel = {};
+
+    let destroyInvoked = false;
+    const destroy = () => {
+      destroyInvoked = true;
```
lgeiger: This could be simplified using [jasmine spies](https://jasmine.github.io/2.0/introduction.html):

```js
const destroy = jasmine.createSpy("destroy");
const component = shallow(
  <ResultViewComponent
    store={mockStore}
    kernel={{}}
    destroy={destroy}
    showResult={true}
  />
);

component.find(".icon-x").simulate("click");
expect(destroy).toHaveBeenCalled();
```

lgeiger:
Side note: It would be great to be able to use jest snapshots for this kind of component testing. It would make things a lot easier for sure.

BenRussert:
> Just guessing from the the code, I would say the button is not rendered because there's no text to be copied in the output.

This is correct. [Here is the relevant line](https://github.com/nteract/hydrogen/blob/master/lib/components/result-view/result-view.js#L182). Basically `this.getAllText().length > 0` checks to see if there is any _text_ to copy, there won't be any text to copy if the output is an image, for example.

spec/components/result-view/result-view-spec.js
BenRussert: @hpierce1102 did you want to take a crack at using Jasmine spies here?

BenRussert:
Thanks for putting this together! I know you had some more goals you mentioned above, but I think it makes sense to merge now and please feel free to send over more PRs if you keep working on that (or anything else 😀)

Welcome to the nteract party!

PR #1450, opened 2018-10-14
Clean up mobx usage

BenRussert:
Makes decorator usage more consistent and fixes some flow issues

### DIFF
--- a/lib/store/index.js
+++ b/lib/store/index.js
@@ -18,14 +18,13 @@ import MarkerStore from "./markers";
 import kernelManager from "./../kernel-manager";
 import Kernel from "./../kernel";
 
-import type { IObservableArray } from "mobx";
-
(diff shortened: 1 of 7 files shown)


lib/store/index.js
```diff
@@ -18,14 +18,12 @@ import MarkerStore from "./markers";
...
 
-import type { IObservableArray } from "mobx";
-
 const commutable = require("@nteract/commutable");
 
 export class Store {
   subscriptions = new CompositeDisposable();
   markers = new MarkerStore();
-  runningKernels: IObservableArray<Kernel> = observable([]);
+  runningKernels: Array<Kernel> = [];
```
lgeiger: Shouldn't this still be an observable?

BenRussert: Whoops, that would have been bad.

PR #1458, opened 2018-10-24
Add code cell folding

### LINKED ISSUE #1001: Feature request: codecell folding
It would be convenient if the codecell functionality were such that I could hide/fold codecells I'm not currently working on in the source. Something like

```
# <codecell> Boring Code
print 'Hello World'
boringCode.done = True

# <codecell> Fun Code
print 'Here's my current working area...
funCode.done = False
```

would become:

```
# <codecell> Boring Code
# <codecell> Fun Code
print 'Here's my current working area...
funCode.done = False
```

I'm referring to the other codecell varients as well, of course, not just ```<codecell>```. Hope that makes sense... :smile:

kylebarron:
Fixes #1001.

This adds two new top-level functions: `hydrogen:fold-current-cell` and `hydrogen:fold-all-but-current-cell`. This uses `codeManager.getCurrentCell` and `codeManager.getCells` to find code cell boundaries, then modifies those boundaries slightly so that the text on the initial line of the cell is still showing.

A couple things to do still:

- [x] Flow types. I'm getting errors from flow IDE that the `selections` property of `TextEditor` is missing. Do I have to add that to `atom.js.flow`?
- [x] Tests. I don't really know how to add these. I'll try to decipher the current code in `code-manager-spec.js`, but might need some guidance on that too.
- [x] Javascript style. This is my first non-trivial change in the Javascript, so I figure I should make sure the general style is ok. Like if I'm creating an array to be filled in with a for loop, should it be initialized with `var` or `let`?

codecell_folding

### DIFF
--- a/lib/code-manager.js
+++ b/lib/code-manager.js
@@ -336,7 +336,7 @@ export function foldCurrentCell(editor: atom$TextEditor) {
   const cellRange = getCurrentCell(editor);
   const newRange = adjustCellFoldRange(editor, cellRange);
   editor.setSelectedBufferRange(newRange);
-  const selection = editor.selections[0];
+  const selection = editor.getSelections()[0];
   selection.fold();
 }
 
@@ -351,9 +351,10 @@ export function foldAllButCurrentCell(editor: atom$TextEditor) {
     newRanges.push(adjustCellFoldRange(editor, allButCurrentCell[i]));
   }
 
+  const selections = editor.getSelections()
   editor.setSelectedBufferRanges(newRanges);
-  for (var i = 0; i < editor.selections.length; i++) {
-    editor.selections[i].fold()
+  for (var i = 0; i < selections.length; i++) {
+    selections[i].fold()
   }
 
   // Restore selections
--- a/lib/code-manager.js
+++ b/lib/code-manager.js
@@ -336,8 +336,7 @@ export function foldCurrentCell(editor: atom$TextEditor) {
   const cellRange = getCurrentCell(editor);
   const newRange = adjustCellFoldRange(editor, cellRange);
   editor.setSelectedBufferRange(newRange);
-  const selection = editor.getSelections()[0];
-  selection.fold();
+  editor.getSelections()[0].fold();
 }
 
 export function foldAllButCurrentCell(editor: atom$TextEditor) {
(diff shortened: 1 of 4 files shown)


lib/code-manager.js
```diff
@@ -331,3 +331,40 @@ export function findCodeBlock(editor: atom$TextEditor) {
...
+export function foldAllButCurrentCell(editor: atom$TextEditor) {
+  const initialSelections = editor.getSelectedBufferRanges()
+
+  const allCellRanges = getCells(editor).slice(1);
+  const currentCellRange = getCurrentCell(editor);
+  const allButCurrentCell = allCellRanges.filter(item => !item.isEqual(currentCellRange))
+  var newRanges = [];
+  for (var i = 0; i < allButCurrentCell.length; i++) {
+    newRanges.push(adjustCellFoldRange(editor, allButCurrentCell[i]));
+  }
```
lgeiger: This is totally valid javascript, but since you asked for some code style advise this could be rewritten as:

```js
const newRanges = allCellRanges
  .filter(cellRange => !cellRange.isEqual(currentCellRange))
  .map(cellRange => adjustCellFoldRange(editor, cellRange));
```

This is more concise and (arguable) more readable.

lib/code-manager.js
```diff
@@ -331,3 +331,40 @@ export function findCodeBlock(editor: atom$TextEditor) {
...
+
+  editor.setSelectedBufferRanges(newRanges);
+  for (var i = 0; i < editor.selections.length; i++) {
+    editor.selections[i].fold()
+  }
```
lgeiger: This can be also rewritten as
```js
editor.selections.forEach(selection => selection.fold())
```

lib/code-manager.js
kylebarron: Thanks! That does look better. I'll update the PR later.

kylebarron:
> You can probably just replace the typings with the ones used by the Nuclide team

Should that be a separate PR to update the typings or added to this?

> For the array example you can actually use it too since you're not reassigning it, you only mutate the content by pushing values.

Makes sense, thanks.

kylebarron:
I think I fixed the flow errors and the code style. Haven't added any tests yet.

kylebarron:
I think it would still be good to add tests. I just haven't looked into that yet.

kylebarron:
While writing tests, I found and fixed a bug, so that was helpful. I like Jasmine, and I really like the automatic linting during the Git commit.

I think it's good to merge now.

kylebarron:
Maybe would be good to add a note to the documentation. Should that go in this PR as well?

BenRussert:
> Maybe would be good to add a note to the documentation. Should that go in this PR as well?

Up to you how to add. Definitely good to document features as we've already got a few out there undocumented.

PR #1461, opened 2018-10-24
Update Atom flow type definitions

kylebarron:
This PR includes updated Atom flow type definitions from Nuclide: https://github.com/facebook/nuclide/blob/master/flow-libs/electron-v2.0.9.js.flow

#1458 uses a part of the selections API that didn't exist in the previous set of type definitions.


kylebarron:
Apparently just copy-pasting the type definitions creates more flow errors. Should I just copy the specific part that's needed for #1458?

Issue #1455, opened 2018-10-21
Allow Hydrogen's autocompletions to be listed first


kylebarron:
It can be pretty annoying to me that my snippets are always listed before the completions from the running kernel, since the latter are usually more personalized to what I'm trying to do. 

Adding a `suggestionPriority` value in the `autocompleteProvider` would allow Hydrogen's autocompletions to be sorted first (depending on the value). The autocomplete-snippets package uses a `suggestionPriority` [value of 2](https://github.com/atom/autocomplete-snippets/blob/ade72c2d4402cdf6292217ca64401164f5cc192f/lib/snippets-provider.js#L6). What about adding a user setting to Hydrogen for what `suggestionPriority` to use? If I set the `suggestionPriority` to 3, then those autocompletions will be listed first.

PR #1465, opened 2018-11-01
Add option to show Hydrogen's autocomplete suggestions first

kylebarron:
Fixes https://github.com/nteract/hydrogen/issues/1455.

To me, I usually find autocomplete suggestions from Hydrogen to be more accurate than other providers or snippets. This adds a setting in the configuration (currently on by default) to set the `suggestionPriority` higher than that of the Snippets autocomplete provider.


kylebarron:
Screenshot with the Bash kernel:
image

Issue #1296, opened 2018-04-22
Export Notebook - markdown cells not correctly exported to ipynb


dynobo:
**Description:**

When exporting a py-File to ipynb-File the Markdown-Cells are exported as Code-Cells, not as Markdown cells. 

**Steps to Reproduce:**

1. Copy the following code to a new py-File:
```
#%%
print('hello world')

#%%markdown
# # Headline1
# - bullet point
# **bold text**

# %%
print('bye world')
```

2. Run "Hydrogen - Export Notebook"

3. Open the created ipynb-File in Jupyter Notebook, you'll see:

![screen1](https://user-images.githubusercontent.com/11071876/39092766-1d380196-4615-11e8-992d-a2fb34f0ffa0.png)

Expected would be a markdown cell:

![screen2](https://user-images.githubusercontent.com/11071876/39092775-3e334f7c-4615-11e8-87a6-ff69921ba870.png)

**Versions:**
Hydrogen 2.4.1
Atom 1.25.1
Running on Solus Linux
Which OS and which version of Hydrogen and Atom are you running?

**Logs:**
*not relevant*

Madhu94:
Picking this up, as I made the original PR for exporting to notebook.

Madhu94:
@dynobo Is the "%%markdown" syntax expected to be like the ipython's "%%markdown" cell magic ? I think in that case, the expected output would be this ?

![image](https://user-images.githubusercontent.com/9015820/42131700-fb1f08a0-7d25-11e8-9287-95d374d01dcf.png)

dynobo:
@Madhu94 
No, not really. I would expect it to me converted into a [cell of type "markdown"](https://ipython.org/ipython-doc/3/notebook/nbformat.html#markdown-cells). After running a cell of this type, the source code vanishs and only the rendered markdown stays visible. 

For an example, see the text in this shot of a notebook. The first two cells ("Jupyter Notebook Slides.." and "Overview" are those markdown cells: 
![notebook](https://cdn-images-1.medium.com/max/1600/1*kt_bEri7QAArBhP1qPJgMg.png)

Madhu94:
My mistake, I thought ``%%markdown`` was a magic like ``%%javascript``. I just wondered where the "%%markdown" syntax came from; I couldn't find it in the Hydrogen docs either.

dynobo:
[...]

PR #1466, opened 2018-11-02
Require cell regex to be at beginning of line

kylebarron:
Currently, the regular expression used to delimit cells matches the comment symbol plus `%%` or other symbols _anywhere_ on the line. That means that using `export-notebook` on the following splits the string:
```py
# %%

import pandas as pd
import numpy as np

print('# %%')
print('this should be one cell')
```

Before:
image

After:
image

Presumably, `run-cell` would also fail on this.

### DIFF
--- a/lib/code-manager.js
+++ b/lib/code-manager.js
@@ -135,7 +135,7 @@ export function getRegexString(editor: atom$TextEditor) {
     commentStartString.trimRight()
   );
 
-  const regexString = `${escapedCommentStartString}(%%| %%| <codecell>| In\[[0-9 ]*\]:?)`;
+  const regexString = `^\s*${escapedCommentStartString}(%%| %%| <codecell>| In\[[0-9 ]*\]:?)`;
 
   return regexString;
 }
--- a/lib/code-manager.js
+++ b/lib/code-manager.js
@@ -135,7 +135,7 @@ export function getRegexString(editor: atom$TextEditor) {
     commentStartString.trimRight()
   );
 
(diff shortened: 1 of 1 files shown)


kylebarron:
Do I need to use `\\s` to represent a space?

n-riesco:
On a regexp, no, but that's a string!

kylebarron:
So I do need to use `\\s` then?

n-riesco:
yes, we need to escape the backslash.

kylebarron:
Thanks. That does make sense.

n-riesco:
I can't look at this right now, but I don't think changing the regexp is the right approach.

I think we should update [this part of the code](https://github.com/nteract/hydrogen/blob/aa55c090dee2721a1c313c442269c429d0de2c46/lib/code-manager.js#L180-L192) to skip regexp matches for which `isComment` is false.

kylebarron:
That makes sense. I might try to find some time to look at that this weekend.

I'm also trying to figure out a good way to refactor some of this code that searches for code cell boundaries in order to carry some metadata about the cell, namely if it's a Markdown cell. Essentially all that's needed to fix #1296 is knowledge about which cells are Markdown and which are code.

BenRussert:
Good catch. I think the approach is fine, my only hesitation is @n-riesco's comment.

BenRussert:
Either way, it would be good to add a test to cover this.

kylebarron:
See #1520

PR #1487, opened 2018-12-02
Add new grammars to list of markupGrammars

kylebarron:
This PR just adds two new grammar names to the list of `markupGrammars` that Hydrogen will run code inside of. This should have no side effects. Fixes https://github.com/kylebarron/stata_kernel/issues/186


kylebarron:
Yeah... I wanted to get around to adding tests to https://github.com/nteract/hydrogen/pull/1458 today but I didn't make it...

PR #1490, opened 2018-12-10
Remove duplicated data explorer docs example

kylebarron:
The data explorer example is listed twice
image

### DIFF
--- a/docs/Usage/Examples.md
+++ b/docs/Usage/Examples.md
@@ -196,17 +196,3 @@ df1 = pd.read_csv(iris_url)
 df1
 {%- endcodetabs %}
 (https://blog.nteract.io/designing-the-nteract-data-explorer-f4476d53f897)
-
-{% codetabs name="Python", type="py" -%}
-import pandas as pd
-
-pd.options.display.html.table_schema = True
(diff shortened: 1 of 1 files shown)


BenRussert:
Oh woops. Actually, I think we should just remove the example entirely. We never shipped the data-explorer as a built-in to hydrogen because it is in alpha and I didn't want to indicate that it was fully supported yet. 

I created a [data-explorer plugin](https://github.com/BenRussert/data-explorer/) where I am working through the various issues with using the data-explorer within atom. Now that I mention it, I have some work to do to get an updated version of this plugin shipped as the data-explorer itself is rapidly developing.

kylebarron:
Right, that makes sense. As well as there's documentation in the plugin itself, we probably don't need to include those examples.

BenRussert:
Did you want to cut the rest of the example out before we merge?

lgeiger:
> On second thought, I don't have a problem with leaving that at the bottom of the examples, if only to advertise the data explorer more.

I agree, we should add a link to the data explorer plugin.

lgeiger:
We can add links to the plugin in a future PR.

Issue #1486, opened 2018-12-01
Toggle output area only works in one direction


fredcallaway:
Atom: 1.32.2
Hydrogen: 2.6.0

The "Toggle output area" command seems to only enable the output area. It doesn't bring back inline results when you run it again.

kylebarron:
The issue is here:

https://github.com/nteract/hydrogen/blob/762852e90137da44eadd4ba50b432875cc02b2b5/lib/main.js#L375-L381

Once the `OutputPane` is created with `toggle-output-area`, the pane exists, and it will continue to exist until the `destroy()` method is called and deletes its entire state. So even when the output area is hidden, `atom.workspace.getPaneItems().find(item => item instanceof OutputPane)` resolves to `true`, so `globalOutputStore` is not `null`, and so it opens the dock again.

I don't know yet how to fix this. The ideal way is to figure out if the output pane is hidden or not, but I don't know how to do that.

fredcallaway:
I'm not sure that the "check if hidden" approach would generate the correct behavior. By default, when you switch to a different tab, the output pane is hidden and it remains hidden when you switch back to the original tab (until you run some code and generate more output). So, I think if you used the "check if hidden" approach, you would effectively automatically toggle back to inline input every time you switched tabs, which I don't think is desirable.

Forgive my naiveté (I've never developed an atom plugin) but would it be possible to simply attach a flag variable to `atom.workspace` indicating whether the output area is active?

BenRussert:
The part that is probably confusing people is that the word "toggle" is actually refering to the visibility of the dock holding the output area, not the "mode" that hydrogen is using to render output (inline vs in the dock). 

We could either add a new command that actually toggles the output mode, or we could consider modifying the existing command to do this.

fredcallaway:
I think it is most intuitive to have one command that toggles the visibility of the dock and also the direction of output, ie output should go to the dock if and only if the output area is visible.

BenRussert:
I think I agree @fredcallaway. I messed around with this some yesterday, we could change this behavior fairly easily.

Anyone else want to weigh in on this? What do you think @lgeiger?

If the user has the output area open, should running the "Toggle Output Area" command destroy the output pane and send the next output back inline? Currently it is only hidden with this command and reopened on output.

kylebarron:
Should be fixed via #1494

PR #1494, opened 2018-12-16
Toggle hydrogen output mode

BenRussert:
There is a discussion on this in #1486. 

In short: the command `hydrogen:toggle-output-area` only toggles visibility of the dock. Some users find this confusing because the command is the only way to enable the output area, but it cannot be used to switch back to inline output mode.

### DIFF
--- a/lib/commands.js
+++ b/lib/commands.js
@@ -1,7 +1,14 @@
 /* @flow */
 
-import { log, reactFactory, INSPECTOR_URI } from "./utils";
+import {
+  log,
+  reactFactory,
+  INSPECTOR_URI,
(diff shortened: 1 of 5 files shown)


BenRussert:
> What happens when the output pane is hidden and toggle is called? Will the output pane receive focus or will the output switch to inline mode?

Calling the `hydrogen:toggle-output-area` command in this case would flip output back to inline. You could call `window:toggle-left|right|bottom-dock` if you just wanted to show a dock that is hidden.

kylebarron:
@BenRussert 	Think this is good to merge?

BenRussert:
@kylebarron, I do, go for it!

Issue #1134, opened 2017-12-02
Try out Atoms new code folding


lgeiger:
For determining which code to run we should take a look at the new code folding methods soon to come to Atom:
https://github.com/atom/atom/pull/16299

lgeiger:
It's now on behind a feature flag on stable: http://blog.atom.io/2018/03/15/atom-1-25.html

lgeiger:
Tree sitter finally fixes the issues we're having due to incorrect indentation based code folding.

Unfortunately it doesn't support highlighting of code cells anymore. I created a Atom issue for this: https://github.com/atom/atom/issues/18196

PR #1498, opened 2018-12-25
Support exporting markdown cells to notebook file

### LINKED ISSUE #1296: Export Notebook - markdown cells not correctly exported to ipynb

kylebarron:
This adds support for Markdown cells in the `export-notebook` command (closes #1296).

- This requires the substring `md` or `markdown` to be present on the cell marker line, i.e. any of the following would work correctly:
	```py
    # %% md
    # %% markdown
    # %% [md]
    # %% [markdown]
    # In[1] md
    # In[1] markdown
    # In[1] [md]
    # In[1] [markdown]
    # <codecell> md
    # <codecell> markdown
    # <codecell> [md]
    # <codecell> [markdown]
	```

	Right now, `# <md>` and `# <markdown>` are **not** supported. It wouldn't be hard to support them; I'd just have to change this regex list:
	https://github.com/nteract/hydrogen/blob/ea0004326a666c679119fbec582530123c4e7cca/lib/code-manager.js#L138

- If there's a comment symbol as the first characters on a line, it and any left whitespace will be removed.
- There's always a cell created before the first `# %%` marker, even if it's on the first line. To me that's a bug, but that should be solved in another PR.

Any comments/code improvement suggestions are welcome. Hopefully this JS is a little more idiomatic than my last PR.

```py
# %%
print('hello world')

# %% markdown
# # Headline1
# - bullet point
# **bold text**

# %%
print('bye world')
```

In Jupyter Notebook:
image

cc: @mwouts, @Madhu94

### DIFF
--- a/lib/code-manager.js
+++ b/lib/code-manager.js
@@ -3,6 +3,7 @@
 import { Point, Range } from "atom";
 
 import escapeStringRegexp from "escape-string-regexp";
+import stripIndent from "strip-indent";
 import _ from "lodash";
 
 import {
@@ -83,33 +84,7 @@ export function removeCommentsMarkdownCell(
       editedLines.push(line);
     }
   });
-  return dedentCommonLeadingWhitespace(editedLines).join("\n");
-}
-
-function dedentCommonLeadingWhitespace(lines: Array<string>): Array<string> {
-  const leadingWhitespaceRegex = /^([ \t]*)(?=[^ \t\n])/;
-  const onlyWhitespace = /^[\s]*$/;
-  const leadingSpaces = _.compact(
-    _.map(lines, line => {
-      if (!line.match(onlyWhitespace)) {
-        const match = line.match(leadingWhitespaceRegex);
-        return match ? match[1] : null;
-      }
-    })
-  );
-  const leadingSubstring = getLongestCommonStartingSubstring(leadingSpaces);
-  return _.map(lines, line => line.replace(leadingSubstring, ""));
-}
-
-// https://stackoverflow.com/questions/1916218/find-the-longest-common-starting-substring-in-a-set-of-strings
-function getLongestCommonStartingSubstring(array: Array<string>): string {
-  const A = array.concat().sort();
-  const a1 = A[0];
-  const a2 = A[A.length - 1];
-  const L = a1.length;
-  let i = 0;
-  while (i < L && a1.charAt(i) === a2.charAt(i)) i++;
-  return a1.substring(0, i);
+  return stripIndent(editedLines.join("\n"));
 }
 
 export function getSelectedText(editor: atom$TextEditor) {
--- a/lib/code-manager.js
+++ b/lib/code-manager.js
@@ -50,6 +50,42 @@ export function getRows(
   return normalizeString(code);
 }
 
+export function getMetadataForRow(
+  editor: atom$TextEditor,
+  start: atom$Point
+): string {
+  // `start` is a Point on the first line of the cell. The cell marker is on the
+  // previous line
+  if (start.row === 0) {
+    return "code";
+  }
+  var rowText = getRow(editor, start.row - 1);
+  if (_.includes(rowText, "md") || _.includes(rowText, "markdown")) {
+    return "markdown";
+  } else {
+    return "code";
+  }
+}
+
+export function removeCommentsMarkdownCell(
+  editor: atom$TextEditor,
+  text: string
+): string {
+  const commentStartString = getCommentStartString(editor).trimRight();
+
+  const lines = text.split("\n");
+  const editedLines = [];
+  _.forEach(lines, line => {
+    if (line.startsWith(commentStartString)) {
+      // Remove comment from start of line
+      editedLines.push(line.slice(commentStartString.length).trimLeft());
+    } else {
+      editedLines.push(line);
+    }
+  });
+  return editedLines.join("\n");
+}
+
 export function getSelectedText(editor: atom$TextEditor) {
   return normalizeString(editor.getSelectedText());
 }
@@ -118,13 +154,15 @@ export function getCodeToInspect(editor: atom$TextEditor) {
   return [code, cursorPosition];
 }
 
-export function getRegexString(editor: atom$TextEditor) {
-  const {
-    commentStartString
-    // $FlowFixMe: This is an unofficial API
-  } = editor.tokenizedBuffer.commentStringsForPosition(
+function getCommentStartString(editor: atom$TextEditor): string {
+  // $FlowFixMe: This is an unofficial API
+  return editor.tokenizedBuffer.commentStringsForPosition(
     editor.getCursorBufferPosition()
-  );
+  ).commentStartString;
+}
+
+export function getRegexString(editor: atom$TextEditor) {
+  const commentStartString = getCommentStartString(editor);
 
   if (!commentStartString) {
     log("CellManager: No comment string defined in root scope");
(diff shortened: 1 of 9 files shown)


mwouts:
Hello @kylebarron , thanks for the developement, this feature was wanted for a long time!

Is it correct that `%% [markdown]` cells from `jupytext` will also be recognized as markdown? Like in this [example](https://github.com/mwouts/jupytext/blob/master/demo/World%20population.pct.py)? As a quick reminder: we have choosen square brackets and not just the work `markdown` in order to support cell titles from Spyder, cf this [thread](https://github.com/spyder-ide/spyder/issues/7933).

kylebarron:
`# %% [markdown]` would be supported fine. As long as the substring `markdown` (or `md`) exists somewhere after `# %%`, it'll be caught. I think that's better than us imposing another format.

I'm also working on a PR to import Notebook files _into_ Hydrogen, and hopefully it'll be possible to import, modify, and export without any data loss. I'd love to explore a `run all cells and export notebook` command that adds cell output to the created notebook file.

kylebarron:
> That sounds great! That will make Hydrogen the first text editor where users can act on the ipynb file without having to run the notebook in full a second time.

Well, not necessarily, at least not at first. I'm working on a simple PR to import _just the source text_. That's quite straightforward. In the short run, no outputs will be saved in import or export. 

I don't know enough about Hydrogen internals yet to work with output in import or export.

lib/code-manager.js
```diff
@@ -50,6 +50,42 @@ export function getRows(
...
+  editor: atom$TextEditor,
+  start: atom$Point
+): string {
+  // `start` is a Point on the first line of the cell. The cell marker is on the
+  // previous line
+  if (start.row === 0) {
+    return "code";
+  }
+  var rowText = getRow(editor, start.row - 1);
+  if (_.includes(rowText, "md") || _.includes(rowText, "markdown")) {
```
lgeiger: I think that's OK for now.

lib/code-manager.js
```diff
@@ -50,6 +50,42 @@ export function getRows(
...
+  text: string
+): string {
+  const commentStartString = getCommentStartString(editor).trimRight();
+
+  const lines = text.split("\n");
+  const editedLines = [];
+  _.forEach(lines, line => {
+    if (line.startsWith(commentStartString)) {
+      // Remove comment from start of line
+      editedLines.push(line.slice(commentStartString.length).trimLeft());
```
lgeiger: Does this still allow for indentation in the markdown cell?

lib/store/index.js
```diff
@@ -76,7 +76,13 @@ export class Store {
...
       let source = codeManager.getTextInRange(editor, start, end);
       source = source ? source : "";
-      const newCell = commutable.emptyCodeCell.set("source", source);
+      const cellType = codeManager.getMetadataForRow(editor, start);
+      if (cellType === "code") {
+        var newCell = commutable.emptyCodeCell.set("source", source);
+      } else if (cellType === "markdown") {
+        source = codeManager.removeCommentsMarkdownCell(editor, source);
+        var newCell = commutable.emptyMarkdownCell.set("source", source);
+      }
```
lgeiger: I'd prefer the use of let here, but that's personal preference:
```js
const cellType = codeManager.getMetadataForRow(editor, start);
let newCell;
if (cellType === "code") {
  newCell = commutable.emptyCodeCell.set("source", source);
} else if (cellType === "markdown") {
  source = codeManager.removeCommentsMarkdownCell(editor, source);
  newCell = commutable.emptyMarkdownCell.set("source", source);
}
```

lib/code-manager.js
kylebarron: You're right, I don't think it does.

lib/store/index.js
kylebarron: Fixed

lib/code-manager.js
kylebarron: Just removing `.trimLeft()` isn't my favorite solution because then if a markdown cell looks like
```py
# %% markdown
# # Header
# **bold**
# - list
#     - indented list
```
then the text in the markdown cell in the Jupyter notebook would still have a leading space in each row, i.e.:
```
 # Header
 **bold**
 - list
     - indented list
```
I think writing with a space after the `#` is clearer in the `.py` file, while having a leading space on every line in the exported markdown cell is annoying.

I added code to dedent common leading whitespace for Markdown blocks, so that the above cell would be converted to
```
# Header
**bold**
- list
    - indented list
```

spec/store/index-spec.js
```diff
@@ -307,13 +307,15 @@ describe("Store", () => {
       expect(store.notebook).toEqual(commutable.toJS(nb));
     });
 
-    xit("should return a fully-fledged notebook when the file isn't empty", () => {
```
kylebarron: Previously the test had used `xit()`, so I think the test was never run, especially because it would have failed.

kylebarron: The test still doesn't pass, but I'm not sure why because it seems to produce the correct output when I paste the test commands into the dev console.

kylebarron:
@lgeiger 	It took a while but I figured out why the test wasn't passing. I think I had to activate `language-python`. I borrowed `waitAsync` from `code-manager-spec.js`. If you want me to create a separate file instead of duplicating the `waitAsync` wrapper, I can do that.

I also added one simple test for testing that code under `%% markdown` should be exported to a Markdown cell. Is that ok by itself?

BenRussert:
> Maybe tomorrow I'll add some documentation.

I'll add the WIP tag for now, but remove it when you are ready and we can merge.

kylebarron:
@BenRussert 	I added a new page in the documentation for exporting a file to a notebook. Hopefully in the future that page will also be updated for _notebook import_ https://github.com/nteract/hydrogen/pull/1501

kylebarron:
I also updated the breakpoints grammar so that markdown cell markers also get highlighted, at least if not using tree-sitter (https://github.com/nteract/hydrogen/issues/1134#issuecomment-427305103)

kylebarron:
@BenRussert 	@lgeiger 	
If I have an approving review and some tests and they pass, is it ok for me to merge my own pull requests? I don't know what the acceptable norms are around that, and I don't want to commit any changes that aren't accepted.

BenRussert:
Thanks @kylebarron!

spec/store/index-spec.js
BenRussert: This was marked as expect fail intentionally because the error didn't make sense and we didn't want to hold up ci.

kylebarron: Ohh that makes sense. Is the updated test ok now?

BenRussert: You seem to have discovered why it was failing and made it pass, though It looks like we are testing that there is an empty cell at the top now, which seems like a bug. I know you pointed this out before, is it something to fix?

kylebarron: I updated that in #1505 https://github.com/nteract/hydrogen/pull/1505/files

BenRussert:
Nice, looks like the test is correct there. Thanks for digging into that!

PR #1505, opened 2018-12-31
Remove initial empty cell from codeManager.getCells()

kylebarron:
Currently, if you have the following text
```py
# %%
print('hello world')
# %%
print('foo bar')
```
and call `codeManager.getCells`, you get _three_ cells, not two. Calling `codeManager.getBreakpoints` returns two breakpoints: `Point(0, 0)` and `Point(2, 0)`. Then calling `codeManager.getCellsForBreakPoints` creates an extra empty cell corresponding to the range from `Point(0, 0)` to `Point(0, 0)`. 

This also means that exporting the above code to a notebook gives _three_ cells, not two. Here's a similar example from #1498 (note the empty first cell despite `print('hello world'` being on the first line of the editor):
image

This PR fixes this behavior by setting the first cell to start with the first non-empty line. So all the following code excerpts would give the same output to `codeManager.getCells`
```py
print('hello world')
# %%
print('foo bar')
```

```py
(extra spaces that don't show up in Github Markdown by default)


print('hello world')
# %%
print('foo bar')
```

```py
(extra spaces that don't show up in Github Markdown by default)


# %%
print('hello world')
# %%
print('foo bar')
```

### DIFF
--- a/lib/code-manager.js
+++ b/lib/code-manager.js
@@ -253,7 +253,7 @@ export function getCellsForBreakPoints(
   let start = new Point(0, 0);
   // Let start be earliest row with text
   editor.scan(/\S/, match => {
-    start = match.range.start;
+    start = new Point(match.range.start.row, 0);
     match.stop();
   });
   return _.compact(
--- a/lib/code-manager.js
+++ b/lib/code-manager.js
@@ -243,18 +243,26 @@ export function getCells(
   } else {
     breakpoints = getBreakpoints(editor);
   }
-  return getCellsForBreakPoints(breakpoints);
+  return getCellsForBreakPoints(editor, breakpoints);
 }
 
 export function getCellsForBreakPoints(
+  editor: atom$TextEditor,
   breakpoints: Array<atom$Point>
 ): Array<atom$Range> {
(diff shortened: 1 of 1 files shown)


kylebarron:
Is it a problem to change the function signature of `getCellsForBreakpoints`? I add `editor` to the inputs.

BenRussert:
> Is it a problem to change the function signature of getCellsForBreakpoints? I add editor to the inputs. (Especially in reference to the plugin architecture)

I don't think there is a problem. We don't include this function as part of the plugin api. In fact, I don't think we even needed to export this from CM as it is only used in this file.

kylebarron:
I saw this PR: https://github.com/nteract/hydrogen/pull/957 and figured that maybe `getCellsForBreakpoints` was also exported out of `CodeManager` as part of the plugin API.

kylebarron:
I added tests for the three examples above.

kylebarron:
Oops; I thought the tests passed...

kylebarron:
I guess I merged before Travis was finished running the second time for the second commit; my bad. The master Travis passed, so we should be good, but I'll wait longer before merging in the future.

PR #1515, opened 2019-01-23
Add import-notebook as opener for .ipynb files

kylebarron:
When a `.ipynb` file is opened, it calls `loadNotebook` instead of loading the JSON into a TextEditor.

Should this be automatic or should it be a config setting?

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -130,6 +130,13 @@ const Config = {
         "If enabled, all files of the same grammar will share a single global kernel (requires atom restart)",
       type: "boolean",
       default: false
+    },
+    importNotebookURI: {
+      title: "Enable Notebook Auto-import",
+      description:
+        "If enabled, opening a Jupyter Notebook file in Atom (with extension `.ipynb`) will instead import the file's source into a new tab.",
+      type: "boolean",
+      default: true
     }
   }
 };
--- a/lib/config.js
+++ b/lib/config.js
@@ -134,7 +134,7 @@ const Config = {
     importNotebookURI: {
       title: "Enable Notebook Auto-import",
       description:
-        "If enabled, opening a Jupyter Notebook file in Atom (with extension `.ipynb`) will instead import the file's source into a new tab.",
+        "If enabled, opening a file with extension `.ipynb` will [import the notebook](https://nteract.gitbooks.io/hydrogen/docs/Usage/NotebookFiles.html#notebook-import) file's source into a new tab. If disabled, or if the Hydrogen package is not activated, the raw file will open in Atom as normal.",
       type: "boolean",
       default: true
     }
--- a/lib/main.js
+++ b/lib/main.js
@@ -11,7 +11,6 @@ import {
 import _ from "lodash";
 import { autorun } from "mobx";
 import React from "react";
-import * as path from "path";
 
 import KernelPicker from "./kernel-picker";
 import WSKernelPicker from "./ws-kernel-picker";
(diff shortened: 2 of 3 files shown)


BenRussert:
> Should this be automatic or should it be a config setting?

I think you are on to something with the config setting. I'm thinking the user should probably be able to open up the raw ipynb somehow if they want.

kylebarron:
Ok, I tried to do that before but got stuck. I gather that the output of `addOpener` is a disposable that you can call when you want to disable the feature, I.e. when the setting is turned off. What I couldn't figure out is how to store this separately from the other disposables... Like you can you find this specific instance to dispose of later?

BenRussert:
I was thinking you'd add some conditional inside the opener function. So if it's ipynb and the config setting, then loadNotebook. Else default.

I think most people will just ignore the config and import the ipynb anyway. I could merge this now and think about it later if you want.

kylebarron:
Oh that makes so much more sense... I'll have to test

kylebarron:
Adding the config setting was so much more straightforward. I think this is ready to merge

lib/config.js
```diff
@@ -130,6 +130,13 @@ const Config = {
         "If enabled, all files of the same grammar will share a single global kernel (requires atom restart)",
       type: "boolean",
       default: false
+    },
+    importNotebookURI: {
+      title: "Enable Notebook Auto-import",
+      description:
+        "If enabled, opening a Jupyter Notebook file in Atom (with extension `.ipynb`) will instead import the file's source into a new tab.",
```
BenRussert: What about linking to the docs and mentioning that Hydrogen has to be active. Something like:

"If enabled, opening a file with extension `.ipynb` will [import the notebook](https://nteract.gitbooks.io/hydrogen/docs/Usage/NotebookFiles.html#notebook-import) file's source into a new tab. If disabled, or if the Hydrogen package is not activated, the raw file will open in Atom as normal."

lib/main.js
```diff
@@ -222,6 +223,17 @@ const Hydrogen = {
...
 
+    store.subscriptions.add(
+      atom.workspace.addOpener(uri => {
+        if (
+          path.extname(uri).toLowerCase() === ".ipynb" &&
+          atom.config.get("Hydrogen.importNotebookURI") === true
+        ) {
+          return loadNotebook(uri);
+        }
+      })
```
BenRussert: What do you think about moving this to `.\import-notebook.js` and importing? That way `main.js` stays cleaner, import notebook logic is isolated ( e.g. `loadNotebook` stays "private" to `importNotebook` to avoid confusion/ambiguity) 

This will make it easier to enhance the opener or import logic down the line if we want and it can be unit tested independently of main.
```js
// import up top becomes:
import { importNotebook, ipynbOpener } from "./import-notebook";
// ...
// then this chunk becomes
store.subscriptions.add(atom.workspace.addOpener(ipynbOpener));
```

Side note: `loadNotebook` is `export`ed only to unit test easier (by skipping over the electron open dialog). We probably should have used the `_` prefix convention and named it `_loadNotebook` initially to indicate that.

lib/main.js
```diff
@@ -11,6 +11,7 @@ import {
 import _ from "lodash";
 import { autorun } from "mobx";
 import React from "react";
+import * as path from "path";
```
BenRussert: we can remove this import as well (see below);

lib/config.js
kylebarron: You're right. I assume that we don't activate Hydrogen at Atom startup to save time? I was hoping the URI would trigger a Hydrogen activation command, but it doesn't look like it does

lib/main.js
kylebarron: All good points.

kylebarron:
I think that should be good now

lib/config.js
BenRussert: I think there is a way to do something minor with a package like adding an opener without activating the package. We can look into it and add in another PR if you want that behavior, I'd ask in the atom slack or http://discuss.atom.io.

Issue #1405, opened 2018-09-07
Hydrogen breaks with a Python 3.7 env but works fine with a 3.6 one


Fercho120:
Hi everyone, I stumbled across an error that confused me a lot:

I have two Python environments managed with conda, one with 3.6 and the other with 3.7, when I try to run on Atom some basic code (for example a print command) selecting the 3.7 kernel I got an error saying that Atom was unable to load some DLL file, however with 3.6 all works fine.
Also if I execute Atom from anaconda's prompt with the 3.7 env active, I got no error
What could be the problem here?

This is a screenshot of the error:

![captura de pantalla 2018-09-07 14 28 09](https://user-images.githubusercontent.com/18184315/45239487-110cbf80-b2ab-11e8-9302-32e84b74dbcc.png)

I'm using:
Windows 10 64 Bits
Atom 1.3
Hydrogen 2.5.1

Error log:
[...]

PR #1519, opened 2019-01-27
install activated Python and R kernels in Anaconda/Miniconda

kiwi0fruit:
I've tested it in the batch and bash on Windows, in Ubuntu 16.04.

May be add code of the instructions here: https://nteract.io/kernels ?

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -25,7 +25,7 @@ Checkout [nteract.io/kernels](https://nteract.io/kernels) for instructions on ho
 Tested and works with:
 
 - [IPython](http://ipython.org/)
-- [IRkernel](https://github.com/IRkernel/IRkernel) `0.4+` requires [`language-r`](https://atom.io/packages/language-r) or similar
+- [IRkernel](https://github.com/IRkernel/IRkernel) `0.4+` requires [`language-r`](https://atom.io/packages/language-r) or similar. See [here](https://github.com/kiwi0fruit/pandoctools/blob/master/docs/tips.md#install-r) how to install R kernel in Anaconda/Miniconda.
 - [IJulia](https://github.com/JuliaLang/IJulia.jl)
 - [iTorch](https://github.com/facebook/iTorch)
 - [IJavascript](https://github.com/n-riesco/ijavascript)
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -20,12 +20,12 @@ we recommend you to install `Hydrogen` via console command: `apm install hydroge
 
 ## Kernels
 
-Checkout [nteract.io/kernels](https://nteract.io/kernels) for instructions on how to install the most popular kernels.
+Checkout [nteract.io/kernels](https://nteract.io/kernels) for instructions on how to install the most popular kernels. Additionally see how to install **activated** Python and R kernel in Anaconda/Miniconda: [Python](https://github.com/kiwi0fruit/pandoctools/blob/master/docs/tips.md#install-python-kernel), [R](https://github.com/kiwi0fruit/pandoctools/blob/master/docs/tips.md#install-r).
 
 Tested and works with:
 
 - [IPython](http://ipython.org/)
-- [IRkernel](https://github.com/IRkernel/IRkernel) `0.4+` requires [`language-r`](https://atom.io/packages/language-r) or similar. See [here](https://github.com/kiwi0fruit/pandoctools/blob/master/docs/tips.md#install-r) how to install R kernel in Anaconda/Miniconda.
+- [IRkernel](https://github.com/IRkernel/IRkernel) `0.4+` requires [`language-r`](https://atom.io/packages/language-r) or similar.
 - [IJulia](https://github.com/JuliaLang/IJulia.jl)
 - [iTorch](https://github.com/facebook/iTorch)
 - [IJavascript](https://github.com/n-riesco/ijavascript)
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -20,7 +20,7 @@ we recommend you to install `Hydrogen` via console command: `apm install hydroge
 
 ## Kernels
 
-Checkout [nteract.io/kernels](https://nteract.io/kernels) for instructions on how to install the most popular kernels. Additionally see how to install **activated** Python and R kernels in Anaconda/Miniconda: [Python](https://github.com/kiwi0fruit/pandoctools/blob/master/docs/tips.md#install-python-kernel), [R](https://github.com/kiwi0fruit/pandoctools/blob/master/docs/tips.md#install-r).
+Checkout [nteract.io/kernels](https://nteract.io/kernels) for instructions on how to install the most popular kernels. Additionally see how to install **activated** Python and R kernels in Anaconda/Miniconda: [Python](https://github.com/kiwi0fruit/pandoctools/blob/master/docs/tips.md#install-python-kernel), [R](https://github.com/kiwi0fruit/pandoctools/blob/master/docs/tips.md#install-r-kernel).
 
 Tested and works with:
 
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -20,12 +20,12 @@ we recommend you to install `Hydrogen` via console command: `apm install hydroge
 
 ## Kernels
 
-Checkout [nteract.io/kernels](https://nteract.io/kernels) for instructions on how to install the most popular kernels. Additionally see how to install **activated** Python and R kernel in Anaconda/Miniconda: [Python](https://github.com/kiwi0fruit/pandoctools/blob/master/docs/tips.md#install-python-kernel), [R](https://github.com/kiwi0fruit/pandoctools/blob/master/docs/tips.md#install-r).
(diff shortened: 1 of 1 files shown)


kylebarron:
Why is any of that necessary?

If I do 
```
conda create --name r-testing r r-essentials
source activate r-testing
R -e "IRkernel::installspec()"
atom .
```
Hydrogen uses the correct R executable

kiwi0fruit:
with the proposed instruction there is no need to start atom from activated
environment.

otherwise R kernel from Anaconda may give dll not found error.

>

kylebarron:
> with the proposed instruction there is no need to start atom from activated environment.

It might be helpful to state this more clearly, but we always require Hydrogen to be started from the activated Conda environment for any language if that conda environment uses a different path. I don't think it's wise to advocate a convoluted workaround when we can just say "activate the conda environment then open Atom"

kiwi0fruit:
I never start Atom from activated environment as it's not convenient. I did
so before I read nteract and Hydrogen manual though. I register the kernel
for the User and it works.

kiwi0fruit:
@kiwi0fruit By the way, where did you get the idea that pyhon environments need activating before using them with Hydrogen or nteract? https://nteract.io/kernels says nothing about it.

nikitakit:
FWIW I have kernels set up for all of my conda environments. This seems to be the whole point behind having the modern kernelspec system, though I essentially have to do this because I only use remote kernels and don't want to start a separate notebook server for each environment. I always have trouble remembering how I set everything up, so a link in the documentation would be good.

However, I'm at a loss why the scripts linked have to be so complicated. I certainly didn't use "exec-wrappers" to set up my kernels, and it's unclear why that would be necessary. My kernelspecs are as simple as:
```json
{
 "argv": [
  "<path-to-anaconda>/envs/<env-name>/bin/python",
  "-m",
  "ipykernel_launcher",
  "-f",
  "{connection_file}"
 ],
 "display_name": "Python 3 (<env-name>)",
 "language": "python"
}
```

If environment dependencies are necessary, the kernelspec json file supports an `env` key for just this purpose.

kiwi0fruit:
Sounds unclear. But googling showed this:

https://ipython.readthedocs.io/en/stable/install/kernel_install.html

> Using virtualenv or conda envs, you can make your IPython kernel in one env available to Jupyter in a different env. To do so, run ipykernel install from the kernel’s env, with –prefix pointing to the Jupyter env:

    /path/to/kernel/env/bin/python -m ipykernel install --prefix=/path/to/jupyter/env --name 'python-my-env'

I hope this would work...

kiwi0fruit:
The proposed instruction didn't work. I still get: 

> Traceback (most recent call last): File "D:\User\Python\Miniconda3\envs\python3\lib\runpy.py", line 193, in _run_module_as_main "main", mod_spec) File "D:\User\Python\Miniconda3\envs\python3\lib\runpy.py", line 85, in _run_code exec(code, run_globals) File "D:\User\Python\Miniconda3\envs\python3\lib\site-packages\ipykernel_launcher.py", line 15, in from ipykernel import kernelapp as app File "D:\User\Python\Miniconda3\envs\python3\lib\site-packages\ipykernel__init__.py", line 2, in from .connect import File "D:\User\Python\Miniconda3\envs\python3\lib\site-packages\ipykernel\connect.py", line 18, in import jupyter_client File "D:\User\Python\Miniconda3\envs\python3\lib\site-packages\jupyter_client__init__.py", line 4, in from .connect import File "D:\User\Python\Miniconda3\envs\python3\lib\site-packages\jupyter_client\connect.py", line 23, in import zmq File "D:\User\Python\Miniconda3\envs\python3\lib\site-packages\zmq__init__.py", line 47, in from zmq import backend File "D:\User\Python\Miniconda3\envs\python3\lib\site-packages\zmq\backend__init__.py", line 40, in reraise(*exc_info) File "D:\User\Python\Miniconda3\envs\python3\lib\site-packages\zmq\utils\sixcerpt.py", line 34, in reraise raise value File "D:\User\Python\Miniconda3\envs\python3\lib\site-packages\zmq\backend__init__.py", line 27, in _ns = select_backend(first) File "D:\User\Python\Miniconda3\envs\python3\lib\site-packages\zmq\backend\select.py", line 27, in select_backend mod = import(name, fromlist=public_api) File "D:\User\Python\Miniconda3\envs\python3\lib\site-packages\zmq\backend\cython__init__.py", line 6, in from . import (constants, error, message, context, ImportError: DLL load failed: �� ������ ��������� ������.

in Hydrogen. Though I changed `/path/to/kernel/env/bin/python` to `/path/to/myenv/bin/python` because I don't have jupyter in the kernel env (and I'm not going to install anything additional to the root env as when it screws conda it's very nasty).

So the only way I know to get working Python and R environments in Hydrogen is the script proposed in the PR.

kiwi0fruit:
@nikitakit I actually also had a working simple jupyter spec as you quote **before** I updated to fresh pandas and Python 3.7

BenRussert:
Are these changes entirely hydrogen specific? Otherwise I think you should open an issue on [the nteract.io site repo](https://github.com/nteract/nteract.io/issues). I'd really prefer to link there for general kernel instructions/advice and if it is specific to hydrogen to just add a paragraph or a page to our [gitbook docs](https://nteract.gitbooks.io/hydrogen/).

kiwi0fruit:
I guess you are right. I'll open PR there. But maybe (?) I would create a dedicated conda packlage that installs activated R and Python kernels...

wadethestealth:
@BenRussert Shouldn't this be closed?

wadethestealth:
Closing because it seems like this PR needs to be modified to be approved on hydrogen, but also seems like the PR has been abandoned.

kiwi0fruit:
But closing this you also agree that this bug should be ignored too :)

wadethestealth:
@kiwi0fruit This is not the case, and you were advised to open a PR on nteract.io for this change and you agreed.

kiwi0fruit:
But I still don't have motivation to reopen it. And who knows when I would
have it (may be I would never have it again as everything works with
hackish instructions in this PR). So closing this PR means removing the
only reminder about the bug.

If you care about the bug that is part of this PR then the right way would
be to close this PR only after opening an issue about this bug (or making
sure that another PR was opened by me - which would be ideal but
unfortunately that is not the case).

>

wadethestealth:
Well it seems like the "bug" was only caused by not having an active environment, which is believed to be a requirement of hydrogen. This PR still exists in its closed state, so it will still be searchable. If you have the motivation to eventually and you believe the bug is important, we strongly encourage you to submit the PR to nteract.io instead of hydrogen. Just because of the nature of the PR not because we don't care about the bug. Also the solution given works, so I believe its best not to open an issue that already has been given a working solution.

kiwi0fruit:
Ah. I see your point about the bug. There is a working solution here so
it's not ideal but OK. But I thought that it's better to fix bugs than
store solutions for them... But that's assuming it's a bug.

Where does the idea about environment activation requirement come from?
Nteract docs never state something like that and I suppose Hydrogen should
work the same way. I also haven't seen something like that in Hydrogen docs.

As far as I know Nteract is actually intended to work without additional
extra activations. If I'm wrong then it would be interesting to know why.

So I see it as a bug.

wadethestealth:
@kiwi0fruit I have no idea where the activated env thing came from, but I believe it's related to how atom works. Since running it from the desktop or similar and opening a project doesn't assign it a directory the same way `atom .` does inside a directory in the terminal. There are many issues that mention it so that how I know. See this issue for example #1405.

kiwi0fruit:
Starting Atom from terminal of activated environment looks like even more
hackish workaround than mine. And yes that's the very same bug as here. And
the issue was also closed with provided workaround but without proper
resolution.

I believe that intended Nteract and Atom-Hydrogen user experience is to
install kernels once then use Nteract and Atom as normal GUI apps without
annoying activations.

wadethestealth:
I am not sure if that's true, but I would love for you to open an issue here on hydrogen explaining your concerns and desires and I will label it for you and have it reviewed. But it seems like the proposed instructions may be too lengthy instead of the "work around" which I believe to be the solution

kiwi0fruit:
@lgeiger Could you please elaborate on what is the intended Hydrogen/Nteract user experience. Please see previous post for details.

wadethestealth:
@BenRussert post above as well

kiwi0fruit:
@wadethestealth You are right that my solution is half-baked really. Though it works with simply copy-pasting. But it should properly be implemented somewhere: either is should be conda package to install activated kernels or the environment activation functionality should be added to Hydrogen itself.

kiwi0fruit:
@wadethestealth Oh. the right place to fix it is a special conda package that patches **any** already installed kernel: takes path to executable from there, creates wrapper script, replaces json. Then add instruction to use it to Nteract docs.

But this patching thing can also be implemented in Hydrogen. So it's debatable.

wadethestealth:
Either route works, but it's probably better to implement it as a plugin for hydrogen that you maintain

kiwi0fruit:
That definitely won't happen :) At the moment I write js only if I **must**. Conda package solution would also work for VSCode so it's a general solution.

PR #1520, opened 2019-01-28
Fix cell boundaries when regex matches outside comment

kylebarron:
Uses `isComment()` per @n-riesco's suggestion. Replaces #1466


n-riesco:
Haven't tested it, but code LGTM.

PR #1529, opened 2019-02-04
Remove trailing newline for each cell in notebook export if it exists.

### LINKED ISSUE #1512: Extra newlines in notebook export
```py
# %%
print('hello world')
# %%
print('hello world')
# %%
print('hello world')
```
gets exported to 

image

which in the upcoming `import-notebook` code, then gets converted back to
```py
# %%
print('hello world')

# %%
print('hello world')

# %%
print('hello world')

```

The source of the problem is the export, not the import.

kylebarron:
Fixes #1512 

```py
# %%
print('hello world')
# %%
print('foo bar')
# %%
print('baz')
```

exports to:
image

and then imports back to the exact original text. 

Note that
```py
# %%
print('hello world')# %%
print('foo bar')# %%
print('baz')
```
is exported to the same as in the screenshot, though the import always places each cell marker on its own line.

### DIFF
--- a/lib/store/index.js
+++ b/lib/store/index.js
@@ -66,9 +66,7 @@ export class Store {
   @computed
   get notebook() {
     const editor = this.editor;
-    if (!editor) {
-      return null;
-    }
+    if (!editor) return null;
     let notebook = commutable.emptyNotebook;
     if (this.kernel) {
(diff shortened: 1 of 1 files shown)


kylebarron:
Let me know if you'd like me to add more tests.

PR #1530, opened 2019-02-05
Provide ordering to configuration settings

kylebarron:
Hydrogen has grown to the point where there are 14 configuration settings. Currently, no ordering is provided, so I think they're ordered alphabetically in terms of the config key. I think this is more confusing than it has to be for users.

This PR attempts to group settings logically and then order them with the `order` value in the dictionary. The idea is "general settings" then "output settings" then "kernel settings", and within each ordered by most used to least used. 

The order is:
- `autocomplete`
- `showAutocompleteFirst`
- `statusBarDisable`
- `debug`
- `autoScroll`
- `outputAreaDefault`
- `outputAreaDock`
- `outputAreaFontSize`
- `globalMode`
- `kernelNotifications`
- `startDir`
- `languageMappings`
- `startupCode`
- `gateways`

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -20,7 +20,8 @@ const Config = {
       description:
         "If enabled, use autocomplete options provided by the current kernel.",
       type: "boolean",
-      default: true
+      default: true,
+      order: 0
     },
     showAutocompleteFirst: {
       title: "Show Hydrogen's autocomplete suggestions first",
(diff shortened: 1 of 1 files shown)


BenRussert:
If you are still up ping me on slack real quick @kylebarron, I can help you get this rebased cleanly. We could do that tomorrow if you are out for the night.

PR #1534, opened 2019-02-06
Allow creating kernel in unsaved file with language grammar set

### LINKED ISSUE #1532: Support kernels in unsaved/untitled editors if grammar is set
Back in v1, I used to love running throwaway kernels in untitled editors by `ctrl+n` then `ctrl+shift+L`to set the grammar for the kernel I wanted. We lost this functionality as a design choice when we shipped [`H20`](https://github.com/nteract/hydrogen/releases/tag/v2.0.0)

I used this almost like a "super console" for debugging/calculations/learning/tryingstuff. If I cared about what I wrote (< 10% of the time :laughing:) id save it, otherwise close without saving to get rid of the clutter.

@kylebarron and I talked about this on the hydrogen channel on the nteract slack (btw __everyone__ is welcome). I'll post below to remember and for those who missed it.

kylebarron:
Fixes #1532 

> For example the kernel monitor component displays a real filename.

For now, I have it set to `Unsaved Editor ${editor.id}`:
image

> Also, the store uses a filename as the key for a file-to-kernel map. An unsaved file can eventually be saved, so we need to handle the this also.

When saved, the key is changed from editor ID to file path.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -347,7 +347,7 @@ const Hydrogen = {
 
     if (!filePath || !grammar) {
       return atom.notifications.addError(
-        "Your file must be saved in order to start a kernel"
+        "The language grammar must be set in order to start a kernel. The easiest way to do this is to save the file."
       );
     }
 
--- a/lib/store/index.js
+++ b/lib/store/index.js
@@ -38,6 +38,12 @@ export class Store {
   globalMode: boolean = Boolean(atom.config.get("Hydrogen.globalMode"));
 
   @computed
+  get untitled(): ?boolean {
+    if (!this.editor) return null;
+    return !this.editor.getPath();
+  }
+
+  @computed
   get kernel(): ?Kernel {
     if (this.globalMode) {
       if (!this.grammar) return null;
@@ -55,7 +61,10 @@ export class Store {
 
   @computed
   get filePath(): ?string {
-    return this.editor ? this.editor.getPath() : null;
+    if (!this.editor) return null;
+    return this.untitled
+      ? `Unsaved Editor ${this.editor.id}`
+      : this.editor.getPath();
   }
 
   @computed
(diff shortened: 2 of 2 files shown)


BenRussert:
Thanks @kbarron, I'll take a close look at this very soon. I'd be so happy if the diff is this light (looks like we need to come up with some tests though).

Side note probably for another PR, I see the `forceEditorUpdate` workaround function I wrote what feels like ages ago. I wrote this hack to make `this.filename` update "imperatively" instead of using mobx properly (of course I didnt know that at the time). I probably should have just refactored the store setup a bit.

kylebarron:
> Thanks @kbarron, I'll take a close look at this very soon. I'd be so happy if the diff is this light (looks like we need to come up with some tests though).

I was aiming for a light diff :)

Yes, I didn't touch the tests yet. I'm not confident enough yet that my solution is the best solution that I feel that it's worth it to write tests before feedback!

Specifically, that area of the tests would need to be refactored a little to test this, because you'd probably want to set Python as the language grammar (which uses the async workaround) only for this test and not the others.

> Side note probably for another PR, I see the `forceEditorUpdate` workaround function I wrote what feels like ages ago. I wrote this hack to make `this.filename` update "imperatively" instead of using mobx properly (of course I didnt know that at the time). I probably should have just refactored the store setup a bit.

Well I don't understand mobx too well... so feel free to refactor some. My intuitive guess to apply `@computed` to `this.untitled` just happened to work haha

kbarron:
Ben, i was not your intended Keith Barron.

BenRussert:
> Yes, I didn't touch the tests yet. I'm generally not confident enough yet that my solutions are the optimal solutions that I feel it's worth it to write tests before feedback!

You might want to try doing the opposite, write tests you think should probably pass then code until they pass locally. This way you save time reloading window or starting up hydrogen every time you want to check. 

It can be hard to keep to this habit, but I can't recommend it enough. If you want I'll look into writing a test or two for this to help you get started.

PR #1537, opened 2019-02-08
Remove force update hack

BenRussert:
I wrote this hack a while back because the mobx calculated value for `store.filePath` would only update if the editor changed completely. This is because `store.filePath` relied on `editor.getPath()`, but this is not tracked by mobx. Changing `store.filePath` to an observable like below is much simpler.

WIP until #1534

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -197,7 +197,10 @@ const Hydrogen = {
         );
 
         editorSubscriptions.add(
-          editor.onDidChangeTitle(newTitle => store.forceEditorUpdate())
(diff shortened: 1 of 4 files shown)


BenRussert:
I had originally planned this to be two PRs to keep things simple. Our store and it's tests are due for some cleanup.

The main thing I've done to our specs here is removed the use of the global store across different unit tests. Now each test initializes a `new Store()` and sets up state for that test only so that it is easier to write tests independently.

BenRussert:
I used most of this work for #1540 and will come back to the rest later if needed

PR #1540, opened 2019-02-10
Support kernels in unsaved files

BenRussert:
@kylebarron and I have been collaborating on this. 

The branch is ready for a spin if you want to try it. My plan is just to let this bake a few days in case of comments/thoughts/ideas/issues we think of.

### DIFF
--- a/lib/store/index.js
+++ b/lib/store/index.js
@@ -39,23 +39,30 @@ export class Store {
 
   @computed
   get kernel(): ?Kernel {
-    if (this.globalMode) {
-      if (!this.grammar) return null;
-      const currentScopeName = this.grammar.scopeName;
-      return this.runningKernels.filter(
-        k => k.grammar.scopeName === currentScopeName
-      )[0];
-    }
+    if (!this.grammar || !this.editor) return null;
 
-    if (!this.filePath) return null;
+    if (this.globalMode) {
+      const currentScopeName = this.grammar.scopeName;
+      return this.runningKernels.find(
+        k => k.grammar.scopeName === currentScopeName
+      );
+    }
+    //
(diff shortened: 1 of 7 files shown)


lib/store/index.js
```diff
@@ -39,23 +39,30 @@ export class Store {
...
       const currentScopeName = this.grammar.scopeName;
-      return this.runningKernels.filter(
+      return this.runningKernels.find(
         k => k.grammar.scopeName === currentScopeName
-      )[0];
+      );
     }
-
-    if (!this.filePath) return null;
+    //
```
kylebarron: Empty comment?

lib/store/index.js
```diff
@@ -174,28 +181,32 @@ export class Store {
...
+
+    const embeddedScope = getEmbeddedScope(
+      editor,
+      editor.getCursorBufferPosition()
+    );
+
+    if (!embeddedScope) return grammar;
+    const scope = embeddedScope.replace(".embedded", "");
+    const grammarForScopeName = atom.grammars.grammarForScopeName(scope);
+    return atom.grammars.grammarForScopeName(scope);
```
kylebarron: Is `grammarForScopeName` used?

lib/store/index.js
BenRussert: Invisible comment 🧙‍♂️. We can remove it with magic.

lib/store/index.js
BenRussert: Nope, good catch. We could drop the line above or use the variable name.

kylebarron: either is fine

BenRussert:
@kylebarron you can push those changes if you want and merge.

BenRussert:
Great to have this back, thanks Kyle!

PR #1547, opened 2019-02-15
Associate MarkerStore with specific TextEditor

### LINKED ISSUE #1496: `hydrogen:clear-results` removes results from all panes
Edit: I originally discovered this behavior from starting a remote kernel, but it also exists when clearing results.

**Description:**

When starting a new remote kernel, any results in an existing remote kernel are deleted.

peek 2018-12-18 16-48

**Steps to Reproduce:**

1. Connect to remote kernel in first file; create some sort of results.
2. Connect to remote kernel in second file; results in first file will be gone.

**Versions:**

Both host and remote computers are running Linux.
```
Atom    : 1.33.0
Electron: 2.0.11
Chrome  : 61.0.3163.100
Node    : 8.9.3
```

**Plugins:**

Do you have any Hydrogen plugins installed and active?
- [x] hydrogen-python
- [ ] Hydrogen Launcher
- [ ] Data Explorer

### LINKED ISSUE #1542: [Bug] Hydrogen long running command loses output stream
Having a long-running command in one tab that produces some output, and shutting down a kernel in another tab will ease all output in all tabs running hydrogen.

**Steps to Reproduce:**

1. Run a Hydrogen tab with say, python [Tab 1]. Run the following code:
```
from time import sleep
print("Starting")  # I will soon disappear
for i in tqdm(range(1000)):
    sleep(1)
```
2. While this is running and providing output. Open a different python file in a new tab [Tab 2].
3. Inside [Tab 2] run: 
```
import os
os.getcwd()
```
4. Confirm there is output
5. Shut down kernel running in [Tab 2]
6. Switch over to [Tab 1] and observe that the previously printed string and output from the tqdm command no longer show. However, the kernel in this tab is still executing the sleep loop.

**Versions:**

OS: Archlinux
Hydrogen: 2.7.0

Atom    : 1.34.0
Electron: 3.1.3
Chrome  : 66.0.3359.181
Node    : 10.2.0

**Plugins:**

- [ ] hydrogen-python
- [ ] Hydrogen Launcher
- [ ] Data Explorer

**Logs:**

No logs required. No logs printed :(

kylebarron:
Fixes #1496, fixes #1542.

As discussed in #1496, there are a number of problems with the current implementation of the `MarkerStore`, which all revolve around using a single global store for all editors and kernels. Clearing results in one pane clears results in all others as well. Running a line of code in one pane clears the result from the same line in the other pane as well!

Since each pane is a separate `TextEditor`, it seemed that the easiest way forward was to use the current marker machinery but use a different `MarkerStore` for each editor.

This is a relatively simple PR that creates a new object in the `Store` named `markersMapping`, which is a `Map` of type `<number, MarkerStore>`, that uses `editor.id` as the key and the given `MarkerStore` as the value.

All issues in #1496 seem to be fixed with this. Open to suggestions!

The last (?) revision of `MarkerStore` may also be of interest: #888

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -149,9 +149,10 @@ const Hydrogen = {
     store.subscriptions.add(
       atom.commands.add("atom-workspace", {
         "hydrogen:clear-results": () => {
-          store.markers.clear();
-          if (!store.kernel) return;
-          store.kernel.outputStore.clear();
+          const { kernel, markers } = store;
+          if (markers) markers.clear();
+          if (!kernel) return;
+          kernel.outputStore.clear();
         },
         "hydrogen:import-notebook": () => importNotebook()
       })
@@ -309,7 +310,7 @@ const Hydrogen = {
   }) {
     log("handleKernelCommand:", arguments);
 
-    const { kernel, grammar } = store;
+    const { kernel, grammar, markers } = store;
 
     if (!grammar) {
       atom.notifications.addError("Undefined grammar");
@@ -327,7 +328,7 @@ const Hydrogen = {
     } else if (command === "restart-kernel") {
       kernel.restart();
     } else if (command === "shutdown-kernel") {
-      store.markers.clear();
+      if (markers) markers.clear();
       // Note that destroy alone does not shut down a WSKernel
       kernel.shutdown();
       kernel.destroy();
@@ -337,7 +338,7 @@ const Hydrogen = {
     ) {
       kernel.transport.promptRename();
     } else if (command === "disconnect-kernel") {
-      store.markers.clear();
+      if (markers) markers.clear();
       kernel.destroy();
     }
   },
@@ -383,9 +384,11 @@ const Hydrogen = {
         : null;
 
     if (globalOutputStore) openOrShowDock(OUTPUT_AREA_URI);
+    const { markers } = store;
+    if (!markers) return;
 
     const { outputStore } = new ResultView(
-      store.markers,
+      markers,
       kernel,
       editor,
       row,
@@ -400,6 +403,7 @@ const Hydrogen = {
 
   restartKernelAndReEvaluateBubbles() {
     const { editor, kernel, markers } = store;
+    if (!editor || !kernel || !markers) return;
 
     let breakpoints = [];
     markers.markers.forEach((bubble: ResultView) => {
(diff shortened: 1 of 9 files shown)


lib/main.js
```diff
@@ -149,7 +149,7 @@ const Hydrogen = {
     store.subscriptions.add(
       atom.commands.add("atom-workspace", {
         "hydrogen:clear-results": () => {
-          store.markers.clear();
+          store.markersMapping.get(store.editor.id).clear();
```
BenRussert: Is flow ok with this? What if editor is null? 💥

I think the solution is to hide the markermap detail I'm the store like we do kernelmapping. Do a null check there once in the getter, then `store.marker` is fine

kylebarron: Flow didn't complain but yes I think your solution is definitely cleaner

kylebarron:
@BenRussert 	can you check again?

kylebarron:
Sorry I didn't see the flow errors before; I think Flow IDE is only set up on one of my computers and not the other

kylebarron:
@BenRussert 	Can you look at this again? I think the flow errors are fixed, but there are a couple tests that fail

spec/store/index-spec.js
```diff
@@ -293,17 +294,16 @@ describe("Store", () => {
...
       const kernel2 = jasmine.createSpyObj("kernel2", ["destroy"]);
       store.runningKernels = [kernel1, kernel2];
       store.dispose();
       expect(store.runningKernels.length).toEqual(0);
+      expect(store.markersMapping.size).toEqual(0);
       expect(store.kernelMapping.size).toBe(0);
       expect(kernel1.destroy).toHaveBeenCalled();
       expect(kernel2.destroy).toHaveBeenCalled();
       expect(store.subscriptions.dispose).toHaveBeenCalled();
-      expect(store.markers.clear).toHaveBeenCalled();
```
kylebarron: @BenRussert 	I couldn't figure out a way to `expect` the `clear` on the values of the `Map`...

BenRussert: I think you would just want to change this to `store.markersMapping.clear`, and then above to ` spyOn(store.markersMapping, "clear");`

lgeiger:
Looks like this needs a rebase.

kylebarron:
Git rebase is scary! :fearful: 	I seem to somehow always mess up my rebases...

camall3n:
My strategy for especially scary rebases is to make a duplicate branch before I rebase. That way there's always an easy-to-reach backup even if something goes wrong.

kylebarron:
That's a good idea!

BenRussert:
No problem, I wanted to get this merged. I figured you could make changes
later if I gave you any ideas you liked.

PR #1566, opened 2019-03-07
Add right-click context menu for import-notebook

BenRussert:
When called as an atom-command from the tree-view context menu, we are passed
back an event containing, among other things, the selected filepath

This is useful when hydrogen isn't activated yet,
but you want to import-notebook from tree-view instead of a dialog


BenRussert:
right click to import demo

Issue #1527, opened 2019-02-02
A style customization example seems incompatible with the current version


aviatesk:
First of all, thanks for your great package !

**Description:**

I found the documentation for "Style Customization of Multiline Result View" does not seem to be compatible with the current release.

The example code chunk in problem is:
> https://nteract.gitbooks.io/hydrogen/docs/StyleCustomization.html
```
.hydrogen {
  .multiline-container {
    .cell_display {
      background-color: @background-color-highlight;
    }
  }
}
```
But `.cell_display` should be replaced by `.hydrogen_cell_display` in order to get the expected result attached within the documentation, like:
```
.hydrogen {
  .multiline-container {
    .hydrogen_cell_display {
      background-color: @background-color-highlight;
    }
  }
}
```

This is kind of just a typo, but I'm welcome to create a PR if you approve it.

**Versions:**

Atom    : 1.34.0
Electron: 2.0.16
Chrome  : 61.0.3163.100
Node    : 8.9.3

**Plugins:**

No Hydrogen plugin is installed.

BenRussert:
> This is kind of just a typo, but I'm welcome to create a PR if you approve it.

I'm so sorry I didn't see this earlier. I'd approve and merge this PR so fast @aviatesk!

aviatesk:
Ok, crafted my very first PR !

BenRussert:
Closing now that your PR is merged, thanks again!

PR #1593, opened 2019-04-13
Fix typos in style customization doc (issue: #1527)

aviatesk:
This PR fixes typos in [the style customization doc](docs/StyleCustomization.md#multiline-result-view).
To be more specific:
- ~~`.cell_display`~~ => `.hydrogen_cell_display`: The class name of cells in a multiline result is now defined as `.multiline-container .hydrogen_cell_display`, but the example still uses `.multiline-container .cell_display` (, about which I created #1527)
- Typos: ~~availible~~ => available, ~~puting~~ => putting

(I found 1 package spec error and 5 flow type-check errors, but they are definitely unrelated to this commit)

### DIFF
--- a/docs/StyleCustomization.md
+++ b/docs/StyleCustomization.md
@@ -2,11 +2,11 @@
 
 By default Hydrogen leans towards plain styling so that its elements can be customized easily to your liking.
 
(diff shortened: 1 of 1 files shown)


aviatesk:
Thank you Ben for merging !

Issue #1569, opened 2019-03-13
hydrogen output doesn't auto wrap


Jiawei-Navican:
**Description:**
The hydrogen output doesn't auto wrap long strings in atom block decoration or output panel.
<img width="800" alt="Screen Shot 2019-03-13 at 11 14 19 AM" src="https://user-images.githubusercontent.com/42557646/54313271-90eb4d80-459e-11e9-86d7-c8d7a1b5bb8d.png">
<img width="801" alt="Screen Shot 2019-03-13 at 2 46 09 PM" src="https://user-images.githubusercontent.com/42557646/54313358-cbed8100-459e-11e9-9b06-48e4cbfedba9.png">

**Versions:**

Which OS and which version of Hydrogen and Atom are you running?
MacOS High Sierra 10.13.6

Atom    : 1.35.1
Electron: 2.0.18
Chrome  : 61.0.3163.100
Node    : 8.9.3

**Logs:**
No error logs and output of the developer tools after following instructions on [Debugging Guide](https://nteract.gitbooks.io/hydrogen/docs/Troubleshooting.html).

**atom `config.json`:**
I am using atom both on windows and macOS, the `config.json` is synced between these two machine using `sync-settings`. So you can find windows pathes in the json file( I don't think it's related to this issue).
```
"*":
  Hydrogen:
    autoScroll: false
    debug: true
    outputAreaDock: true
  "atom-beautify":
    executables:
      php:
        path: "D:\\MAMP\\bin\\php\\php7.1.5\\php.exe"
    python:
      beautify_on_save: true
      formatter: "yapf"
      ignore: [
        "E24"
        "E731"
      ]
  "atom-ide-ui":
    "atom-ide-diagnostics-ui":
      showDiagnosticTraces: true
    "debugger-exceptionBreakpointsCollapsed": true
  "autocomplete-python": {}
  core:
    disabledPackages: [
      "markdown-preview"
      "atom-ide-debugger-python"
      "linter"
      "linter-ui-default"
      "language-python"
    ]
    projectHome: "C:\\Users\\jiawku\\git"
    telemetryConsent: "no"
    themes: [
      "atom-material-ui"
      "one-dark-syntax"
    ]
    uriHandlerRegistration: "always"
    useTreeSitterParsers: false
  editor:
    fontSize: 16
    preferredLineLength: 110
    softWrap: true
    softWrapAtPreferredLineLength: true
    tabLength: 4
  "exception-reporting":
[...]

PR #1594, opened 2019-04-14
Fix wrapping result issue #1569

aviatesk:
This PR fixes [hydrogen.less](styles/hydrogen.less) so that results will be wrapped correctly.

Why `pre-wrap` ?:
- The values that wrap texts other than `pre-wrap` *don't* preserve a sequence of whitespaces/tabs, which seems inappropriate in code results. See https://developer.mozilla.org/en-US/docs/Web/CSS/white-space#Values.

### DIFF
--- a/styles/hydrogen.less
+++ b/styles/hydrogen.less
@@ -27,7 +27,7 @@ code, pre {
   font-size: inherit;
   font-family: inherit;
   padding: 0;
-  white-space: pre;
+  white-space: pre-wrap;
 }
 
 img {


BenRussert:
I'll tell you a cool trick @aviatesk:

You referenced the issue # in the commit message like a pro, but if you use the phrase "closes #xxx"  or "fixes #xxx" github will automatically close the referenced issue when the PR is merged!

aviatesk:
@BenRussert 
Oh, I'll definitely use that in the next PR !
Thanks for telling me that 😆

PR #1601, opened 2019-04-25
Enhance kernel monitor

aviatesk:
This PR enhances various features within kernel monitor, which were proposed in #1514.

## What I'm trying to do

- [x] Beautify monitor pane using [react-table](https://github.com/tannerlinsley/react-table/tree/v6#column-header-groups)
- [x] Allow user to click on a filename to activate the editor
    * [x] Not create link to unsaved editors because there is no URI for them
- [x] Add a column for `execution_count`
    * [x] Implement `execution_count` watcher both for ZMQ/WS kernel
- [x] Add a column for last execution time
- [ ] Add a column for kernel spec
- [x] Update usage/style customization documentation

## What I've done so far

Looks like this: 
image

## Remaining work

Want to hear general feedbacks and your ideas about the topics below:
- What/how to show kernel specs:
    * I think opening the kernel.json file is more useful for users than showing the current kernel spec (, which easily can be obtained from `kernel.kernelSpec`).
- The _last execution time_ info:
    * I could implement a watcher for the last execution time as I did so for execution count, but I'm not so sure that this is useful or just verbose that I've not included this enhancement in this PR yet.


BenRussert:
This is looking great @aviatesk! I'll look closer into this after you remove WIP from the title. Here are just a couple thoughts for now:

- For the kernelspec/execution time questions, would you want to split those changes into a separate PR (or discuss in an  issue) after this one is merged rather than holding up the rest of your work?

- For unsaved editors, we may be able to use the editor id. Though I am curious if you think this would be more difficult than imagine off the top of my head. (possibly another PR anyway)

lib/components/kernel-monitor.js
```diff
@@ -66,22 +54,101 @@ const KernelMonitor = observer(({ store }: { store: store }) => {
...
+        const kernel = props.value;
+        return [
+          <a
+            className="icon icon-playback-pause"
+            onClick={interrupt.bind(this, kernel)}
+            title="Interrupt kernel"
+            key={kernel.displayName + "int"}
+          />,
+          <a
+            className="icon icon-playback-rewind"
```
BenRussert: What about something that looks like arrows pointing around, :recycle: . Maybe `icon-sync` for this?

aviatesk: Ha, it's just because I couldn't find the icon when I once looked through the styleguide, but now I've finally found it and changed the restart icon to `icon-sync` !

aviatesk:
Thanks for your feedback, @BenRussert !

My thoughts for your comments are below:

***

> For the kernelspec/execution time questions, would you want to split those changes into a separate PR (or discuss in an issue) after this one is merged rather than holding up the rest of your work?

The information of kernel specs which are currently stored within Hydrogen's kernel representation is like:
```
{argv: Array(5), display_name: "Python 3", language: "python", name: "python3"}argv: (5) ["C:\Users\aviat\Anaconda3\python.exe", "-m", "ipykernel_launcher", "-f", "{connection_file}"]display_name: "Python 3"language: "python"name: "python3"__proto__: Object
```
and we can easily show kernel spec information using this which would look like:
image

But I think those information is less useful than jumping to _the original kernel.json_ file since those kernel specs are read from the file.
Then the problem here is that we've not stored the files information within kernel object and it needs some refactoring in a package in node_modules...

_So in conclusion, I think this enhancement should be in another PR if you go with my idea, otherwise it can be easily included in this PR ._

***

> For unsaved editors, we may be able to use the editor id. Though I am curious if you think this would be more difficult than imagine off the top of my head. (possibly another PR anyway)

Actually I implemented "jumping to file" feature just using `kernel.getFilesForKernel()` and `atom.workspace.open(filePath)` and it totally doesn't work for unsaved editors.

I don't have any idea to get unsaved texteditor's URI and thus I left it unjumpable.
Do you know me a way to get an URI for them ?

***

And I've included last execution time enhancement within the last commit.
Now the monitor looks like:
image

Do you think this information is verbose or useful ??

lib/components/kernel-monitor.js
```diff
@@ -1,62 +1,50 @@
...
+  await atom.workspace
+    .open(file, {
+      searchAllPanes: true
+    })
+    .catch(err => {
+      atom.notifications.addError("Hydrogen", {
+        description: err
+      });
+    });
+}
```
BenRussert: > I implemented "jump to file" feature just using kernel.getFilesForKernel() and atom.workspace.open(filePath) and it totally doesn't work for unsaved editors.

What I had in mind was something like 
```js
const editor = atom.workspace
  .getTextEditors()
  .find(nexEditor => nextEditor.id == relevantEditorId);

atom.workspace.open(editor);

```

The only problem is that our store is set up to use human readable/friendly filenames instead of id numbers. So you would have to parse out the id from `Unsaved Editor x` to get `relevantEditorId` for this example. I think it would be better to just move toward using ids in our store in another unrelated PR, then easily add jump to for unsaved editors.

lib/components/kernel-monitor.js
```diff
@@ -66,22 +54,108 @@ const KernelMonitor = observer(({ store }: { store: store }) => {
...
-          group={group}
-          key={group}
-        />
-      ))}
-    </div>
-  );
+
+  const data = _.map(store.runningKernels, kernel => {
+    return {
+      gateway: kernel.gatewayName || "Local",
```
BenRussert: This line needs fixing to support `WSKernel`. Maybe `kernel.transport.gatewayName` if flow allows it.

lib/components/kernel-monitor.js
```diff
@@ -66,22 +54,108 @@ const KernelMonitor = observer(({ store }: { store: store }) => {
...
+      style: { textAlign: "right " }
+    },
+    {
+      Header: "Managements",
+      accessor: "kernel",
+      Cell: props => {
+        const kernel = props.value;
+        return [
+          <a
+            className="icon icon-playback-pause"
```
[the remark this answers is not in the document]
aviatesk: Okay I will follow you !

lib/components/kernel-monitor.js
aviatesk: Thanks for pointing this out ! I totally forget to look back to this part.

Flow doesn't allow `kernel.transport.gatewayName` and only `Kernel` & `WSKernel` has that property, which seems really weird.

 Now I find the _property_ `gatewayName` is only used in existing-kernel-picker.js and there is no need for `Kernel` object to have this property but only `WSKernel` need to have it.

lib/components/kernel-monitor.js
```diff
@@ -66,22 +54,108 @@ const KernelMonitor = observer(({ store }: { store: store }) => {
...
+            className="icon icon-sync"
+            onClick={restart.bind(this, kernel)}
+            title="Restart kernel"
+            key={kernel.displayName + "restart"}
+          />,
+          <a
+            className="icon icon-trashcan"
+            onClick={shutdown.bind(this, kernel)}
+            title="Shutdown kernel"
+            key={kernel.displayName + "shutdown"}
```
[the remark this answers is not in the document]
aviatesk: I didn't encounter any problem or error message maybe because within react-table, the row itself may have an unique key ?

Anyway, I changed it to use index as a key like: `map(runningKernels, (kernel, key: number) => {...}`, which must be unique across the row & cells

aviatesk:
I'm welcome to write tests ! even though it takes some time to finish because I'm an just newbie to JavaScript & unfamiliar with Jusmine.

First I will try to write and I get stuck I will reach you out.
I may strip `WIP` before I write all the tests up.

BenRussert:
> First I will try to write and I get stuck I will reach you out.
I may strip WIP before I write all the tests up.

Sounds good! Tests can be hard to write here at first because of the need to mock atom and kernel related stuff. I wrote a simple one for this last night, so we can think about tests in another PR after merging this.

lib/components/kernel-monitor.js
aviatesk: Phew, I implemented an easy workaround with `Unsaved Editor`'s ID to open it in [this commit](https://github.com/nteract/hydrogen/pull/1601/commits/fdaee59d469f4de428f28ad109c213bd7cc00d09) with this approach.

I also included `@TODO` comment for further implementation improvements in the future.

In anyway, now we can jump to _undestroyed_ unsaved text editors from the kernel monitor !

aviatesk:
Okay, I've almost done what I wanted to do in my first mind !
Thanks so much for your help @BenRussert !

Throughout the implementation, I did many (of course biased 🤣) manual tests and I want to strip WIP for now.
Of course I know writing tests would be much valuable & important, so I will write them in another PR or include them in this PR if I have time before merging !

aviatesk:
Ah, I will update the corresponding documentation first !

styles/hydrogen.less
```diff
@@ -4,6 +4,7 @@
 // for a full listing of what's available.
 @import "ui-variables";
 @import "syntax-variables";
+@import (less) "../node_modules/react-table/react-table.css";
```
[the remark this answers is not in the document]
aviatesk: Oh really, it would be so annoying.

Do you think is it better to copy the .css file from the original react-table module and paste it in `styles` directory in order to ensure there is the file ??

I first did so but changed to import it since I thought it would more robust to the package update.

lib/components/kernel-monitor.js
```diff
@@ -66,22 +81,114 @@ const KernelMonitor = observer(({ store }: { store: store }) => {
...
+              onClick={openEditor.bind(this, filePath)}
+              title="Jump to file"
+              key={filePath + "jump"}
+            >
+              {tildify(filePath)}
+            </a>
+          );
+          return (
+            <div
+              className="test"
```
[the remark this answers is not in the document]
aviatesk: I just used it when I tweaked styles of that part and totally forget to remove it 😆 
Let's remove in the test PR !

package.json
```diff
@@ -80,6 +80,7 @@
     "react-dom": "^16.3.2",
     "react-hot-loader": "^4.3.3",
     "react-rangeslider": "^2.1.0",
+    "react-table": "^6.10.0",
```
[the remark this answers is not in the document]
aviatesk: Currently I didn't anything with input element focused, so I think we can overlook this bug.

aviatesk:
@BenRussert 

Thanks for merging !
I will join to write tests with you in that PR as well !

PR #1621, opened 2019-05-16
Update regexString

wadethestealth:
This is a small but hopefully quality of life update.

Current Effects:
The change to the regex string allows for any amounts of spaces rather than one or zero between the comment characters and any of the characters that indicate the start of a cell heading. It also removed an unnecessary duplicate search with a space for the `%%` and ` %%` headings.

Future Effects:
Their is also a second change to the regex in which before their were no capturing groups for different cell types. The new regex adds those capturing groups allowing for easy determination of cell types, which currently are only code and markdown. This means it will make it easier to make hydrogen behave differently for multiple cell types in the future.

This change should not affect any current or legacy code. Unless another branch is using the current regexString for capturing the heading line without the comment in front.

### DIFF
--- a/lib/code-manager.js
+++ b/lib/code-manager.js
@@ -177,7 +177,7 @@ export function getRegexString(editor: atom$TextEditor) {
 
   const escapedCommentStartString = escapeStringRegexp(commentStartString);
 
-  const regexString = `${escapedCommentStartString}(%%| %%| <(?:codecell|md|markdown)>| In\[[0-9 ]*\]:?)`;
+  const regexString = `${escapedCommentStartString} *%% *(md|markdown)?| *<(codecell|md|markdown)>| *(In\[[0-9 ]*\])`;
 
   return regexString;
 }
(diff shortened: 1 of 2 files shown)


BenRussert:
Thanks for the PR!

How do you feel about adding a test to the `./spec/code-manager-spec.js` file? If you add a test that fails before the change and passes after it will ensure we don't break this in the future.

wadethestealth:
> How do you feel about adding a test to the `./spec/code-manager-spec.js` file? If you add a test that fails before the change and passes after it will ensure we don't break this in the future.

I will brainstorm some good tests and add them sometime soon.

BenRussert:
That would be awesome.

The test you add could be very similar to other tests we have in codemanager now, so you could copy boilerplate from there if that helps. Happy to help if you have any questions!

wadethestealth:
> 
> 
> That would be awesome.
> 
> The test you add could be very similar to other tests we have in codemanager now, so you could copy boilerplate from there if that helps. Happy to help if you have any questions!

I just altered a pre-existing test that is supposed to test the old regex string. This is ready for review or merge @BenRussert

BenRussert:
Thanks for the updates!

Issue #1632, opened 2019-05-29
Code block text from markdown cells gets default hydrogen styling


wadethestealth:
1. Create Python file with the following code:
```
#%%md
# `Shift-Enter`
#Shift-Enter
```
2. Run line 2 and line 3 in any way you would like
3. The results will appear visually the same

aviatesk:
Close this for now since this issue is _sort of_ solved in #1670.

Issue #1630, opened 2019-05-29
Custom executables for display.Markdown [feature request]


kiwi0fruit:
It would be nice to be able to set custom Markdown to HTML converters for `display.Markdown(...)` in Atom settings. Like path to executable with arguments that reads from stdin and writes to stdout (like `pandoc --from gfm` or `pandoctools --profile Jupyter`).

With this it would be easier to write [Stitch/Knotr](https://github.com/pystitch/stitch) and [Knitty](https://github.com/kiwi0fruit/knitty)/[Pandoctools](https://github.com/kiwi0fruit/pandoctools) documents that put Markdown code inside Python code (all these programs use Pandoc filter that runs code blocks via Jupyter kernels, grabs outputs, parses and inserts them into Pandoc abstract syntax tree for future processing).

Putting all Markdown into Python code would allow using f-strings for variables substitution and would utilize existing Hydrogen capabilities without additional modifications (like running Markdown cells in Hydrogen too - but that's a separate feature request).

kiwi0fruit:
I would appreciate pointing to the right place in Hydrogen code to make changes so that IO could make a PR (if there is no objections to it).

wadethestealth:
@kiwi0fruit I am unsure what you mean by `display.Markdown()` but with IPython.display you can already put Markdown in python code and output it

kiwi0fruit:
I mean `IPython.display.Markdown()`. This creates special Markdown object that is then parsed and displayed by Hydrogen (other Jupyter clients also do this). So I'm talking about overriding client behavior. Not server behavior. Server sends Markdown object and is OK with that.

wadethestealth:
@kiwi0fruit The kernel returns a data stream which we copy to an output component called output store. This output store is a prop of a Display component inside ResultView from inline results and is a prop of an Output component inside of History. Both History and ResultView are exported from files in hydrogen/lib/components/result-view, and Output and Display are imported from @nteract/display-area, which is outdated and there has been inside talk of updating this. Also markdown cells are now runnable as of today, if you are using this repo inside of the released version from apm.

kiwi0fruit:
Thanks! I'll dig into this. And nice to hear that Markdown cells work now.

kiwi0fruit:
> So technically Hydrogen doesn't do any parsing, but its dependency does

[...]

PR #1641, opened 2019-05-31
Update to @nteract/outputs

wadethestealth:
## Intentions
The intention of this PR is to remove `@nteract/display-area` and `@nteract/transforms` which are both outdated, archived, and no longer being maintained. Both of these packages have been replaced by `@nteract/outputs`, which is the new package that accomplishes the same tasks, but adds more flexibility for rendering.

## Effects
- This should have no effects on any code not based inside or on top of the edited view components.
- This should have no immediate visual effects.
- This decreases our number of dependencies (outdated ones at that).
- This allows for better rendering control and customization.
   - For example, fixing #1632.
   - Or, for adding custom markdown flavors #1630 

### This PR is ready for review.


aviatesk:
Good work :)

Could you kindly explain or give us a link to understand that the your statements below are really true ?

- > This should have no effects on any code not based inside or on top of the edited view components.
- > For example, fixing #1632.

wadethestealth:
Sure thing @aviatesk. The new react components take the exact same input meaning no change to previous code that interacted with the previous components. It also produces the same elements with the same classes because `outputs` is meant to replace `display-area` and `transforms` which means no style changes.

BenRussert:
I will have a solid review for this tonight. I'm so glad you took the lead on this @aviatesk!

lib/components/result-view/display.js
```diff
@@ -0,0 +1,27 @@
...
+
+class Display extends React.PureComponent<Props> {
+  render() {
+    return (
+      <Output {...this.props}>
+        <ExecuteResult expanded>{MediaComponents()}</ExecuteResult>
+        <DisplayData expanded>{MediaComponents()}</DisplayData>
+        <StreamText expanded />
+        <KernelOutputError expanded />
+      </Output>
```
BenRussert: Check out [this example](https://github.com/nteract/nteract/blob/a44db1c20aa6d30c78711958a55f0ddf6fff33cc/packages/outputs/__tests__/output.spec.tsx#L112-L128). 

We would support more output media types, what used to be called "transforms". Below we are listing explicitly each media type we support, and `RichMedia` will automatically select the "richest" one based on the order they are declared below:

```jsx
<RichMedia>
    <Vega3/>
    <Vega2/>
        ...
    <Media.Json />
    <Media.HTML />
    <Media.Plain />
</RichMedia>
```

This may eliminate the need for the `transforms.js` file. Unless the list of components would need to be reused outside this function.

lib/components/result-view/transforms.js
```diff
@@ -1,26 +1,56 @@
...
-  registerTransform
-} from "@nteract/transforms";
+import React from "react";
+import { Media } from "@nteract/outputs";
 import PlotlyTransform from "@nteract/transform-plotly";
 import { VegaLite1, VegaLite2, Vega2, Vega3 } from "@nteract/transform-vega";
 
+PlotlyTransform.defaultProps = {
+  ...PlotlyTransform.defaultProps,
+  mediaType: PlotlyTransform.MIMETYPE
```
BenRussert: If `mediaType` is not already a default prop on these transforms, I wonder if it should be. Might be a quick PR to the monorepo.

lib/components/result-view/result-view.js
```diff
@@ -159,14 +158,9 @@ class ResultViewComponent extends React.Component<Props> {
...
-          <Display
-            outputs={toJS(outputs)}
-            displayOrder={displayOrder}
-            transforms={transforms}
-            theme="light"
-            models={{}}
-            expanded
-          />
+          {toJS(outputs).map((output, index) => (
+            <Display output={output} key={index} />
```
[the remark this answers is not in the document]
BenRussert: It may only need to be unique within each `map` function. I am not sure about this though. [Here are the relevant docs](https://reactjs.org/docs/lists-and-keys.html#keys) for more info.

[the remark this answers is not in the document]
aviatesk: > [Keys Must Only Be Unique Among Siblings](https://reactjs.org/docs/lists-and-keys.html#keys-must-only-be-unique-among-siblings)

My fault 😅 I misunderstood they should be globally unique.

lib/components/result-view/transforms.js
wadethestealth: Yeah wouldn't be hard to do. I just don't know how to propose things to them on the mono @BenRussert

lib/components/result-view/display.js
wadethestealth: @BenRussert I feel like the `transforms.js` simplifies each component by calling that function. This function is also used in another file too. However, I also keep a `MediaArray` in case we want easier access to their properties. Instead, I would rather name the file to `media-types.js` since transforms is no longer the correct naming convention. 

Also I don't understand if this was meant for me. 
>Check out this example.
>
>We would support more output media types, what used to be called "transforms". Below we are listing explicitly each media type we support, and RichMedia will automatically select the "richest" one based on the order they are declared below:

I just don't know what you mean here since where you provided your example Output is required and ExecuteResult and DisplayData are wrappers for RichMedia

BenRussert:
Thanks for the great work @wadethestealth! 

I'm going to merge this into another branch on the main repo before we merge into master later on. Some of the review comments are difficult to explain so I will make another PR to show you what I mean.

PR #1643, opened 2019-06-01
Update Troubleshooting.md

LuizGyro:
Accounting for when "No kernel for grammar Null Grammar found" message is displayed.
This was an issue that came up when I was setting up, one which had a very simple solution,
but I had to look on a separate forum post to find. I think it should be present in the
documentation.

### DIFF
--- a/docs/Troubleshooting.md
+++ b/docs/Troubleshooting.md
@@ -30,6 +30,12 @@ Use your distribution's package manager to install.
 
 If your default `python` is 3.x, you need to run instead `PYTHON=python2.7 apm install hydrogen` or change the default version for `apm` with `apm config set python $(which python2.7)` beforehand. You can still use 3.x versions of Python in Hydrogen, but it will only build with 2.x due to a [longstanding issue with `gyp`](https://bugs.chromium.org/p/gyp/issues/detail?id=36).
 
+### No kernel for grammar Null Grammar found
+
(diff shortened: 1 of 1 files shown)


wadethestealth:
IMO: This doesn't seem like a hydrogen issue. If you have no grammar, then you obviously wont be able to open a kernel which requires a grammar. However the file extension is denoted in grammars, and you can also point any kernel to any grammar in Hydrogen package settings, which means that the 'correct' file extension actually is *not* required.

LuizGyro:
I understand. Maybe it would be better to instead add what you said to the troubleshooting page, then?

wadethestealth:
Could you explain why someone wouldn't use a normal file type that isn't caught by atom grammars, which use hundreds of different file types?
And if they are using a non normal file type why wouldn't they use a grammar for atom to color their keywords correctly??

LuizGyro:
The issue does not reside in either of the situations you described.
It is when a user is eager to test the new package, so they create a new file, but fail to save it before using hydrogen. Then, if they are inexperienced, they might be frustrated, because there is no direct reference to the problem they are having in the Troubleshooting section.
I know it may sound like a stretch, but it can happen.

wadethestealth:
@LuizGyro could you review #1643 and see if this makes sense to you and would be helpful for a beginner?

PR #1646, opened 2019-06-03
Update Troubleshooting Docs

wadethestealth:
## The Issue
The troubleshooting doc had an outdated error name, which I have changed to the current day error. I also added a tidbit that I believe should help people troubleshooting beginner issues like #1643. I also believes this fills a gap on kernel information, which we were trying to let the nteract website deal with, but they have no useful info on that site for debugging hydrogen issues.

## Notes
- Please review this with the mindset of a beginner as it needs to be as clear as possible what they need to do to fix their issue.
- If you are new to Hydrogen check out the files changed, and comment whether you believe this is helpful or not.


BenRussert:
> If you are new to Hydrogen check out the files changed, and comment whether you believe this is helpful or not.

I've been waiting to merge to see if anyone chimes in :smile:

wadethestealth:
Me too @BenRussert

PR #1649, opened 2019-06-04
New Error For Import Notebook

### LINKED ISSUE #1648: Import Notebook opens two untitled documents
**Description:**

When I create a new .ipynb file in atom and ask hydrogen to import the notebook nothing happens.
When importing existing notebooks (created in Jupyter Lab), atom opens two untitled notebooks with the desired imported content, but once one of those two files is saved it is not importable again. 

**Steps to Reproduce:**

1. [create new .ipynb file]
2. [try to import notebook]
3. [import existing .ipynb]
4. [try to save one of the two untitled documents created on step 3.]
**Versions:**

Windows 10 with Atom 1.37

**Plugins:**

Do you have any Hydrogen plugins installed and active?
- [ .] hydrogen-python
- [.] Hydrogen Launcher
- [.] Data Explorer

**Logs:**

Please post any error logs and the output of the developer tools as described in our [Debugging Guide](https://nteract.gitbooks.io/hydrogen/docs/Troubleshooting.html).

wadethestealth:
## The Issue
When supplying `parseNotebook` with a bad JSON format (this includes empty files), we didn't tell the user that an error occurred. Nothing would happen visually, but an experienced user could see an error in the console. This PR now catches this error and throws an atom error with the stack trace, so that users know they don't had a valid file.

## Notes
- `Import Notebook` also has always thrown an error immediately after creating a `.ipynb` in atom, since it is not a valid notebook and atom tries to open it immediately.

---

- Closes #1648

### DIFF
--- a/lib/import-notebook.js
+++ b/lib/import-notebook.js
@@ -54,15 +54,22 @@ export function importNotebook(event?: atom$CustomEvent) {
 
 export async function _loadNotebook(filename: string) {
   let data;
+  let nb;
   try {
     data = await readFileP(filename);
+    nb = parseNotebook(data);
   } catch (err) {
-    atom.notifications.addError("Error reading file", {
-      detail: err.message
-    });
+    if (err.name === "SyntaxError") {
+      atom.notifications.addError("Error not a valid notebook", {
+        detail: err.stack
+      });
+    } else {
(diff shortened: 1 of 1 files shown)


BenRussert:
Cool thanks and good catch!

PR #1655, opened 2019-06-07
Fix up outputs upgrade

BenRussert:
- Use latest plotly and vega packages, closes #1525 and #1524
- Add tests
- Isolate all toJS(output) calls to the Display component
- Use a hack to determine if outputs are text in `isTextOutputOnly`
- Remove need for transforms.js file
- All supported display_data and execute_results mediaTypes are in Display component


lib/components/result-view/display.js
```diff
@@ -1,27 +1,61 @@
...
+    <Media.Json />
+    <Media.JavaScript />
+    <Media.HTML />
+    <Media.Markdown />
+    <Media.LaTeX />
+    <Media.SVG />
+    <Media.Image />
+    <Media.Plain />
+  </RichMedia>
+);
```
wadethestealth: I don't think that it is necessary to wrap them in a rich media, since react can interpret arrays as a group. The only place this would be necessary is if you tried to used this const by itself, but that could be seen as unclear inside a render function. If you desired use by itself, you could wrap the array with a RichMedia, and then it becomes much more clear whats going on without having to switch files to this declaration. It also technically uses a slight bit less memory with the array approach of course we are talking bytes here though.

lib/components/result-view/display.js
```diff
@@ -1,27 +1,61 @@
...
+    mediaComponent => mediaComponent.props.mediaType
+  );
+  const bundleMediaTypes = [...Object.keys(data)].filter(mediaType =>
+    supported.includes(mediaType)
+  );
+
+  return bundleMediaTypes.length === 1 && bundleMediaTypes[0] === "text/plain"
+    ? true
+    : false;
+}
```
wadethestealth: This seems like a good hack for now

lib/components/result-view/display.js
```diff
@@ -1,27 +1,61 @@
...
+
+export default function Display(props: Props) {
+  return (
+    <Output output={toJS(props.output)}>
+      <ExecuteResult expanded>{supportedMediaTypes}</ExecuteResult>
+      <DisplayData expanded>{supportedMediaTypes}</DisplayData>
+      <StreamText expanded />
+      <KernelOutputError expanded />
+    </Output>
+  );
```
wadethestealth: I am unsure of how exactly this works like a react component, but it works and nicely done

spec/components/display-spec.js
```diff
@@ -0,0 +1,44 @@
...
+  };
+  let richerBundle = {
+    "text/plain": "I'm very plain",
+    "text/html": "<div>I am a little <b>richer</b>!</div>"
+  };
+  it("should return true if text is the richest supported output", () => {
+    expect(isTextOutputOnly(plainBundle)).toEqual(true);
+    expect(isTextOutputOnly(richerBundle)).toEqual(false);
+  });
+});
```
wadethestealth: Glad we got some test cases

lib/components/result-view/display.js
BenRussert: Yeah, this might seem a little weird right now since the idea of theis that this would all be one component (no array or instance like above).

The idea of the  `@nteract/outputs` api  is you just declare all the supported mediaTypes right in your component, in the order you want them prioritized, and the `Output` and `RichMedia` internals handle the rest. 

I don't think performance will come into play for this, it is more about keeping "transforms" in one place and _hopefully_ being easy to read. 

Since we have `ExecuteResult` and `DisplayData` we need to use this at least twice. I am about to open a PR on the monorepo to change that and get something like below:

```jsx
 function Display(props) {
  return (
    <Output output={toJS(props.output)}>
      <DisplayOrExecuteResult expanded>
        <Vega3 />
        <Vega2 />
        <PlotlyTransform />
        <VegaLite2 />
        <VegaLite1 />
        <Media.Json />
        <Media.JavaScript />
        <Media.HTML />
        <Media.Markdown />
        <Media.LaTeX />
        <Media.SVG />
        <Media.Image />
        <Media.Plain />
      </DisplayOrExecuteResult>
      <StreamText expanded />
      <KernelOutputError expanded />
    </Output>
  );
}

```

BenRussert:
Feel free to merge this @wadethestealth when you are ready.

PR #1658, opened 2019-06-12
Fix to #807

### LINKED ISSUE #807: incorrect cursor_pos for non-BMP characters in Jupyter protocol
It looks like you need an adjustment identical to jupyter/notebook#2509 to correctly report cursor positions for non-BMP Unicode characters (surrogate pairs) in completion and tooltip requests.

wadethestealth:
Please see #807 for full explanation on the fix.

Using the test case below the cursor position should be 5 at the end of the line for both code introspection and autocompletion.
```javascript
𝐚𝐚𝐚𝐚𝐚
```

### Notes:
- Fixes #807

### DIFF
--- a/lib/autocomplete-provider.js
+++ b/lib/autocomplete-provider.js
@@ -1,7 +1,7 @@
 /* @flow */
 
 import _ from "lodash";
-import { log } from "./utils";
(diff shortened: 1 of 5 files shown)


BenRussert:
Thanks @wadethestealth!

PR #1659, opened 2019-06-13
Fix folding of cells

### LINKED ISSUE #1508: Bug with folding current cell
```py
# %%
print('hello world')
# %%

```
The cell folds incorrectly when the next cell marker is on the following line without an empty line:
image
image

wadethestealth:
Closes #1508

### DIFF
--- a/lib/code-manager.js
+++ b/lib/code-manager.js
@@ -465,9 +465,10 @@ function adjustCellFoldRange(editor: atom$TextEditor, range: atom$Range) {
     range.end.row == editor.getLastBufferRow()
       ? range.end.row
       : range.end.row - 1;
+  const endWidth = editor.lineTextForBufferRow(endRow).length;
 
   return new Range(
     new Point(startRow, startWidth),
-    new Point(endRow, range.end.column)
+    new Point(endRow, endWidth)
   );
 }


aviatesk:
@wadethestealth 
Thanks for addressing this issue ! Your fix works.

As a further fix, I wonder you feel like you can fix this too ?
`fold-current-cell` for multiple grammar files (like markdown) can fold multiple cells if invoked on _header of cell_. Please refer to the example below:

Before folding:
image

After folding (dispatch on ` ```python`):
image

wadethestealth:
@aviatesk Currently this is impossible. We currently detect for embedded scopes, which the header is not apart of. The other way to detect would be a markdown code block, but this is specific to markdown and if two code blocks have no blank line between them they will read as the same block. So for now its worth noting that the header is not a part of the cell

aviatesk:
@wadethestealth 
Oh so if I dispatch `fold-current-cell` on an header, the detected scope would be just an grammar of the body text (like markdown, pweave markdown or so) ?

Then I wonder we should disable the command on a body text of multi-grammar files including header.

wadethestealth:
You are totally right since markdown isn't considered a cell but more of a wrapper @aviatesk

aviatesk:
@wadethestealth 
Okay. Anyway, let's put this to another PR/issue.

PR #1661, opened 2019-06-14
Don't focus on kernel monitor after toggling

aviatesk:
## Rationale

Currently `toggle-kernel-monitor` uses just `atom.workspace.toggle` API and after invoking the command our focus will be on the kernel-monitor pane and IMO most of the time we have to manually switch back to the previous editor (or pane in general).

This is because, within kernel monitor, we just _monitor_ it or click to do some action, and there is nothing to do with keybindings, thus there is no need to stay focus on it after toggling it.

## Approach

This PR includes the workaround for `atom.workspace.toggle` using `prevPane` stuff, cause there seems no option for `toggle` API unlike `open`...

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -115,7 +115,7 @@ const Hydrogen = {
           const lastItem = atom.workspace.getActivePaneItem();
           const lastPane = atom.workspace.paneForItem(lastItem);
           await atom.workspace.toggle(KERNEL_MONITOR_URI);
-          lastPane.activate();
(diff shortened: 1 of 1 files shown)


wadethestealth:
Once you fix flow I approve

aviatesk:
ah I totally forget to run `npm run flow` ha

wadethestealth:
Im going to test this before merging, give me some time

PR #1662, opened 2019-06-15
Add a callback for file delete event:

### LINKED ISSUE #1618: We can't restore an unsaved editor once it get destroyed, so it seems better to ...
https://github.com/nteract/hydrogen/blob/99f2c1fb4ea9fc5e9ab108618ff9c9c26b4c7bcf/lib/components/kernel-monitor.js#L40-L45

---

###### This issue was generated by [todo](https://todo.jasonet.co) based on a `TODO` comment in 99f2c1fb4ea9fc5e9ab108618ff9c9c26b4c7bcf when #1601 was merged. cc @aviatesk.

aviatesk:
## The purpose of this PR:

Plz refer to #1618.

## Approach

If an unsaved editor to which a kernel is already attached is destroyed(deleted), we no longer restore the editor and of course use _jump to file_ functionality within kernel-monitor neither.

I addresses this issue by adding a callback to unsaved editors that called when it destroyed (& saved) and changes the file name and disables _jump to file_ functionality.

## Discussion ?

In the current implementation, the kernel itself wouldn't be shutdown. This is because by doing so, we can reconnect to the kernel after deleting files attached to it.

But that case might be very advanced and it might be more friendly if we just shutdown the kernel when any other editor isn't connected to the kernel. (Jupyter kernel can eat some memory in general)

Thus I'm very welcome to discussion.

## Note

Actually the same thing does happen for saved files. Thus I included the fix for them as well in this PR.

***

Closes #1618

### DIFF
--- a/lib/store/index.js
+++ b/lib/store/index.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-import { CompositeDisposable } from "atom";
+import { CompositeDisposable, File } from "atom";
 import {
   observable,
   computed,
@@ -9,7 +9,11 @@ import {
   keys,
   values
 } from "mobx";
-import { isMultilanguageGrammar, getEmbeddedScope } from "./../utils";
+import {
+  isMultilanguageGrammar,
+  getEmbeddedScope,
+  isUnsavedFilePath
+} from "./../utils";
 import _ from "lodash";
 
 import Config from "./../config";
@@ -49,7 +53,6 @@ export class Store {
       );
     }
 
-    const grammar = this.getEmbeddedGrammar(this.editor);
     const kernelOrMap = this.kernelMapping.get(this.filePath);
     if (!kernelOrMap) return null;
     if (kernelOrMap instanceof Kernel) return kernelOrMap;
@@ -124,6 +127,35 @@ export class Store {
     this.startingKernels.set(kernelDisplayName, true);
   }
 
+  addFileDisposer(editor: atom$TextEditor, filePath: string) {
+    const fileDisposer = new CompositeDisposable();
+
+    if (isUnsavedFilePath(filePath)) {
+      fileDisposer.add(
+        editor.onDidSave(event => {
+          fileDisposer.dispose();
+          this.addFileDisposer(editor, event.path); // Add another `fileDisposer` once it's saved
+        })
+      );
+      fileDisposer.add(
+        editor.onDidDestroy(() => {
+          this.kernelMapping.delete(filePath);
+          fileDisposer.dispose();
+        })
+      );
+    } else {
+      const file: atom$File = new File(filePath);
+      fileDisposer.add(
+        file.onDidDelete(() => {
+          this.kernelMapping.delete(filePath);
+          fileDisposer.dispose();
+        })
+      );
+    }
+
+    this.subscriptions.add(fileDisposer);
+  }
+
   @action
   newKernel(
     kernel: Kernel,
@@ -140,6 +172,7 @@ export class Store {
     } else {
       this.kernelMapping.set(filePath, kernel);
     }
+    this.addFileDisposer(editor, filePath);
     const index = this.runningKernels.findIndex(k => k === kernel);
     if (index === -1) {
       this.runningKernels.push(kernel);
--- a/lib/components/kernel-monitor.js
+++ b/lib/components/kernel-monitor.js
@@ -9,6 +9,7 @@ import tildify from "tildify";
 
 import typeof store from "../store";
 import Kernel from "../kernel";
+import { isUnsavedFilePath } from "../utils";
 
 const showKernelSpec = (kernelSpec: {}) => {
   atom.notifications.addInfo("Hydrogen: Kernel Spec", {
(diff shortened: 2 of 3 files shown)


wadethestealth:
Rather than saying "Unsaved Editor 7 (Destroyed)" why not just remove the file name instead. I consider the Files column to be only actively connected files. Also just a thought, but probably better for another PR what if we could name the kernels.

aviatesk:
@wadethestealth 
I kept the name as `Unsaved Editor 7 (Destroyed)` because the kernel itself is still alive and thus we might want to  re-connect the kernel.
I mean if we totally remove the file name, and repeat the process of "attach kernel to unsaved editor -> delete unsaved editor` for _multiple times_, then we can't tell the difference between the kernels.

I know this is very advanced cases, and thus it may sound like _too much_. But if we don't shutdown the kernel itself, this idea seems more safer IMO.

wadethestealth:
Yes thats why I think there should be names for them instead because if you connect a kernel to a bunch of files but them delete them they are just clutter at that point, and the names will distinguish multiple from each other

aviatesk:
Ah I got it. Good idea.

Maybe giving unique id numbers to each unique kernel isn't difficult.
Not sure it would appear sane or non-weird tho.

wadethestealth:
What about that existing thing that gets generated that's basically an I'd right?

aviatesk:
If you mean `connectionFile`, I'm negative, since the name of it usually very arbitrary like `C:\Users\aviat\AppData\Roaming\jupyter\runtime\kernel-9377d5f1-730b-4112-9056-4ea84609bba5.json`.
And even if you meant the kernelspec json file, it won't change anything cause we can spawn multiple kernels from a single kernelspec file. (i.e.: `displayName` is defined in that file)

Thus I think it's more better to attach unique indice to `displayName`s, like:
```
1: Python 3.7 - the names of files
2: Python 3.7 - the names of files
3. Julia 1.1.1 - the names of files
...
```

BenRussert:
I'm worried about adding complexity here. What if we just add 
```js
if (!editor.getPath()) {
   // so that the KernelManager, etc, updates:
  store.kernelMapping.delete(`Unsaved ${editor.id}`);
}
``` 
to [the existing `editor.onDidDestroy` hooks](https://github.com/nteract/hydrogen/blob/3e20663882c4fe8808ef5646511c8d9442e7e6f5/lib/main.js#L196-L200).

Then just let the actual kernel processes get cleaned up naturally [here](https://github.com/nteract/hydrogen/blob/master/lib/store/index.js#L184)

aviatesk:
Yeah I begin to think that way is better here.
After all it's very rare that an user does `launch a kernel on an unsaved editor -> launch another kernel on another unsaved editor -> destroy both editors -> try to reconnect the kernels from other editors` (where we may want `destroyed` or index stuff) is very, very rare.

Thomas also recommended that at first, thus I'll try to change to follow your idea.

aviatesk:
@wadethestealth @BenRussert 
I modified this PR so that it would just delete a file from `kernelMapping` when it gets deleted.

I tried to add specs to this like below, but it's not worked so far. Maybe because `fs.unlink` doesn't actually delete a file ?

> Under [`describe("getFilesForKernel", () => {`] (https://github.com/nteract/hydrogen/blob/master/spec/store/index-spec.js#L205)
```js
it("should not return deleted files previously related to kernel", async () => {
  const path1 = "foo.jl";
  const path2 = "bar.jl";
  const path3 = "baz.py";
  const editor1 = createMockEditor(grammar1);
  const editor2 = createMockEditor(grammar1);
  const editor3 = createMockEditor(grammar2);
  await editor1.saveAs(path1);
  await editor2.saveAs(path2);
  await editor3.saveAs(path3);

  store.newKernel(kernel1, path1, editor1, grammar1);
  store.newKernel(kernel1, path2, editor2, grammar1);
  store.newKernel(kernel2, path3, editor3, grammar2);
  fs.unlink(path2, () => {
    expect(store.getFilesForKernel(kernel1)).toEqual([path1]);
  });
  fs.unlink(path3, () => {
    expect(store.getFilesForKernel(kernel2)).toEqual([]);
  });
  fs.unlinkSync(path1);
});
```

wadethestealth:
unlink is supposed to delete a file, so I am not too sure what went wrong

lib/store/index.js
```diff
@@ -124,6 +127,35 @@ export class Store {
...
+        })
+      );
+      fileDisposer.add(
+        editor.onDidDestroy(() => {
+          this.kernelMapping.delete(filePath);
+          fileDisposer.dispose();
+        })
+      );
+    } else {
+      const file: atom$File = new File(filePath);
```
BenRussert: Is this the atom `File` or the [ browser API ](https://developer.mozilla.org/en-US/docs/Web/API/File)

If it's atom's does it need to be imported, or are they the same class?

BenRussert: I see the import now 😄

aviatesk: Yeah, I'm using Atom API and it's imported.

aviatesk:
I may be able to set a test case in soon.

BenRussert:
Async tests can be difficult. Try unlinkSync or you may need to throw an `await` in there someplace.

BenRussert:
If you can get a test working that would make this better, but I'm ok with merging whenever you're ready. Also, if you don't think the test would add value, don't spend too much time on it. I just like tests that help prevent breaking something by mistake in the future.

aviatesk:
Hah, I couldn't make this through ! 😅 

But I would merge this into branch, since this change is small tweak and is't solid one that we wouldn't pay much time to try to keep this ;)

PR #1664, opened 2019-06-17
Add supported image mediaTypes

BenRussert:


aviatesk:
Yay ! This does work, thank @BenRussert !

PR #1670, opened 2019-06-22
Add `className` for markdown results

aviatesk:
## Purpose

Address #1632

## Approach

1. Separate markdown results from others in lib/display.js and add `className` to the markdown component
2. Tweak styles for it

Step 1. appears very clumsy, but still it works like this:
image

## Alternative approach

If `@nteract/output` starts to support adding custom `className` for each `Media` component, the code would be more cleaner.

## Misc

I also beautified indentations in hydrogen_style.less


wadethestealth:
I understand that an if statement does work here, and I know I have been busy lately so I apologise. But the if statement adds unnecessary complexity. What you should do is create a new file for a new component that returns a div wrapped around the media.markdown class, and also make sure that the default props for the new component are the same as the default props of media.markdown

BenRussert:
Check out [the source code for the Markdown component](https://github.com/nteract/nteract/blob/master/packages/outputs/src/components/media/markdown.tsx). The code is very simple, so it would be easy to copy into a custom component here like @wadethestealth said. As long as the `mediaType` default prop stays the same you won't need an if statement and the custom component can replace the current markdown component in `supportedMediaTypes`

aviatesk:
@wadethestealth @BenRussert 
Thanks for your reviews !

I just modified the files according to your suggestions, and would appreciate you took a look.

aviatesk:
I also removed `react-hot-loader` dependency in the last commit: 71f85b2.

I confirmed Hydrogen works after a clean installation.

wadethestealth:
Approved

PR #1676, opened 2019-06-25
update REAME plugin section

hoishing:
added Cell Navigation in the plugin section


aviatesk:
Looks good, thanks !

I'm pretty sure your package doesn't use [Hydrogen Plugin-API](https://nteract.gitbooks.io/hydrogen/docs/PluginAPI.html), but still I think this can be very useful in conjunction with Hydrogen, thus I gonna merge it.

PR #1683, opened 2019-06-27
Upgrade to flow-bin 0.102.0

wadethestealth:
My first attempt at updating our flow-bin and our types. Please review this carefully and let me know if anything is incorrectly. Also make sure I didn't change any code logic. It should just be safety checks, but just in case.

Closes #1677

### DIFF
--- a/lib/store/index.js
+++ b/lib/store/index.js
@@ -52,8 +52,9 @@ export class Store {
         k => k.grammar.scopeName === currentScopeName
       );
     }
-
-    const kernelOrMap = this.kernelMapping.get(this.filePath);
+    const file = this.filePath;
+    if (!file) return null;
+    const kernelOrMap = this.kernelMapping.get(file);
     if (!kernelOrMap) return null;
     if (kernelOrMap instanceof Kernel) return kernelOrMap;
     return this.grammar && this.grammar.name
@@ -70,7 +71,7 @@ export class Store {
   }
 
   @computed
-  get filePaths(): Array<?string> {
+  get filePaths(): Array<string> {
     return keys(this.kernelMapping);
   }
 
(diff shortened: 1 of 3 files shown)


lib/store/index.js
```diff
@@ -168,7 +169,7 @@ export class Store {
         this.kernelMapping.set(filePath, new Map());
       }
       const multiLanguageMap = this.kernelMapping.get(filePath);
-      multiLanguageMap.set(grammar.name, kernel);
+      if (multiLanguageMap) multiLanguageMap.set(grammar.name, kernel);
```
BenRussert: Alternatively, 
```js
multiLanguageMap && multiLanguageMap.set(grammar.name, kernel);
```

(Just a comment, it's up to you)

lib/store/index.js
```diff
@@ -199,10 +200,11 @@ export class Store {
     this.runningKernels = this.runningKernels.filter(k => k !== kernel);
   }
 
-  getFilesForKernel(kernel: Kernel): Array<?string> {
+  getFilesForKernel(kernel: Kernel): Array<string> {
     const grammar = kernel.grammar.name;
     return this.filePaths.filter(file => {
       const kernelOrMap = this.kernelMapping.get(file);
+      if (!kernelOrMap) return [];
```
BenRussert: We want to return boolean values here inside of the filter function. If I'm reading this right, an array returned would be coerced to `true` and this could potentially return an array with an irrelevant file.

aviatesk:
Nice try !
I pointed one part that I want to modify, but it's also up to you.

lib/store/index.js
wadethestealth: wait then why the `: Array<string>` is that not the return value in flow

wadethestealth:
I dont see your comment @aviatesk

lib/store/index.js
aviatesk: You may misunderstand the return value from `filter` function.
Refer to this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

wadethestealth: Oh no I see now. I didnt see the return filter was wrapping all of the other returns.

PR #1687, opened 2019-06-30
Make `Display` observer

### LINKED ISSUE #1678: 2.10 breaks inline output between jupyterlab and atom.
**Configuration**
windows 7 and ubuntu 19.04 atom clients connecting to jupyterlab server on lan.

**Description:**
- immediately after updating to 2.10, inline output from the kernel to atom client began malfunctioning.
- previous behavior: when program contains logger output, output was displayed in the client in sync with program execution in a streaming fashion.
- after upgrading to 2.10 from 2.9, ~3 or so lines of logger output would be returned to atom's inline window (very small, window, btw. not sure if relevant). the kernel would show as busy the entire time the program was running (expected). remainder of logger output generated during execution would be presented only upon completion/termination of the program. 

**Steps to Reproduce:**
1. write something which contains/generates a decent amount of python logger output (think debug level)
2. run all on the python script from atom.
3. make sad face.

**Versions:**
tested from windows 7 pro and ubuntu 19.04 desktop. same results.
atom 1.38.2

**Plugins:**
plugins ruled out as contributing factor as expected behavior resumed after downgrading back to 2.9 on both windows and ubuntu clients.

**Logs:**
checked syslog as well as interactive debugger. nothing glaring. don't think this is a bug, but a poorly designed feature between 2.9 and 2.10

just ballpark, but probably these guys: #1647, #1569, and possibly modifications to @nteract/outputs.

feels like the soft wrap was hard coded or something. i don't have it enabled under hydrogen, but my lines were absolutely getting wrapped under 2.10, and i was perfectly happy with how things were configured and working under 2.9. If changes are going to be made to wrapping and inline output, they should without a doubt be configurable via Hydrogen package settings via the UI.

if i can provide any additional info, let me know.

### LINKED ISSUE #1681: Time module not working well in Hydrogen
Hey devs:
Somethings happening again like this. 
The example is not working well:
```python
import time

for i in range(10):
    time.sleep(1)
    print(i)
```
Regular inline result view or dock view are both not behaving well: they all give the first 1 or 2 output then wait another 8 secs and give out all the output at once. Same thing happens with `tqdm`.

OS: MacOS Mojave 10.14.2
atom: 1.38.2
python: 3.7.0
tqdm: 4.28.1
hydrogen: 2.10.0

thx.

_Originally posted by @AstrickHarren in https://github.com/nteract/hydrogen/issues/1293#issuecomment-506135743_

aviatesk:
## Purpose

This PR just does what @BenRussert proposed here: https://github.com/nteract/hydrogen/issues/1678#issuecomment-505910843
Please see https://github.com/mobxjs/mobx-react#faq for why I took the `observer` class way.

## Misc

I confirmed this still works if we updated to mobx-react v6 as I'm doing in #1631 

- Fixes #1678
- Fixes #1681

### DIFF
--- a/lib/components/result-view/display.js
+++ b/lib/components/result-view/display.js
@@ -2,6 +2,7 @@
 
 import React from "react";
 import { toJS } from "mobx";
+import { observer } from "mobx-react";
 import {
   DisplayData,
   ExecuteResult,
@@ -15,8 +16,6 @@ import PlotlyTransform from "@nteract/transform-plotly";
 import { VegaLite1, VegaLite2, Vega2, Vega3 } from "@nteract/transform-vega";
 
 import Markdown from "./markdown";
-
-type Props = { output: any };
 
 // All supported media types for output go here
 export const supportedMediaTypes = (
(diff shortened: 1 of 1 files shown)


BenRussert:
Looks good to me. Do you think it will fix the stream issue in #1678? I think it _could_ be a fix, but I don't have a way to repro the stream issue to see.

wadethestealth:
@BenRussert I believe it is the fix. I know we dont have the exact test case from #1678, but #1681 should be the same issue

aviatesk:
@BenRussert I confirmed this PR fixes the stream issue in #1678 as far as I tested 😄

aviatesk:
Like this:
image

PR #1688, opened 2019-06-30
Add Soft Wrapping To Config

### LINKED ISSUE #1679: Add config setting to revert to pre-2.10.0 output soft-wrapping behaviour
**Description:**

Ever since the latest update (2.10.0) the inline output in python is being soft-wrapped in most cases, where earlier it would just provide a horizontal scroll-bar.

For example, when looking at the repr() of a list too large to be shown on a single line, atom will always put each element of the list in its own line. However, if any given element is longer than the available space in the output bubble, it will now be split across multiple lines, whereas before this last version each element would remain on its own line, and you'd just have to scroll right.

Here's a screenshot showing an example of what I mean:

image

However, when you ctrl+click to open it in a new pane, you see the repr() as it actually is, without the soft wrapping:

image

This is how it used to show in the output pane before.

Is there any way to recover the old functionality? If there isn't currently an option to do so, would it be possible to have a setting implemented for the next release to completely disable this new kind of soft-wrapping in the inline output panes?

I've found this: https://github.com/nteract/hydrogen/issues/1569, which looks like the issue that caused this change. Someone specifically stated: 'I think this is a fix that most people want, so a config probably isn't needed. If people want they can always set their personal atom stylesheet.'

However, I disagree. I much prefer the old functionality over the new one and I'm sure I'm not alone. Could we please get a config setting to toggle this behaviour?

In the meantime, could someone explain to me how I can revert this change just for myself using a stylesheet? I've never used any of that stuff, so I'm not sure how I'd go about doing that. I'd be so grateful.

**Versions:**

Windows 10 Professional
Atom: 1.38.2
Electron: 2.0.18
Chrome: 61.0.3163.1000
Node: 8.9.3

**Plugins:**

- [x] hydrogen-python
- [ ] Hydrogen Launcher
- [ ] Data Explorer

wadethestealth:
Using html attributes and css, I enabled a way for us to configure css styles, while still having their rules be inside the stylesheet and be able to override.

The purpose of the added css is to revert behavior of #1594, and make word-breaking/wrapping a configurable option since #1569 was not actually a bug but a feature request.

Closes #1679

### DIFF
--- a/lib/components/result-view/history.js
+++ b/lib/components/result-view/history.js
@@ -47,9 +47,9 @@ const History = observer(({ store }: { store: OutputStore }) => {
         style={{
           fontSize: atom.config.get(`Hydrogen.outputAreaFontSize`) || "inherit"
         }}
-        hydrogen-wrapOutput={(atom.config.get(
+        hydrogen-wrapOutput={((atom.config.get(
           `Hydrogen.wrapOutput`
-        ): boolean).toString()}
+        ): any): boolean).toString()}
       >
         <Display output={output} />
       </div>
--- a/lib/components/result-view/history.js
+++ b/lib/components/result-view/history.js
@@ -47,7 +47,9 @@ const History = observer(({ store }: { store: OutputStore }) => {
         style={{
           fontSize: atom.config.get(`Hydrogen.outputAreaFontSize`) || "inherit"
         }}
-        hydrogen-wrapOutput={atom.config.get(`Hydrogen.wrapOutput`).toString()}
+        hydrogen-wrapOutput={(atom.config.get(
+          `Hydrogen.wrapOutput`
+        ): boolean).toString()}
       >
         <Display output={output} />
       </div>
--- a/lib/components/result-view/history.js
+++ b/lib/components/result-view/history.js
@@ -47,9 +47,9 @@ const History = observer(({ store }: { store: OutputStore }) => {
         style={{
           fontSize: atom.config.get(`Hydrogen.outputAreaFontSize`) || "inherit"
         }}
-        hydrogen-wrapOutput={((atom.config.get(
-          `Hydrogen.wrapOutput`
-        ): any): boolean).toString()}
+        hydrogen-wrapOutput={Boolean(
+          atom.config.get(`Hydrogen.wrapOutput`)
+        ).toString()}
       >
         <Display output={output} />
       </div>
--- a/lib/components/result-view/history.js
+++ b/lib/components/result-view/history.js
@@ -47,9 +47,7 @@ const History = observer(({ store }: { store: OutputStore }) => {
         style={{
           fontSize: atom.config.get(`Hydrogen.outputAreaFontSize`) || "inherit"
         }}
-        hydrogen-wrapOutput={Boolean(
-          atom.config.get(`Hydrogen.wrapOutput`)
-        ).toString()}
+        hydrogen-wrapOutput={atom.config.get(`Hydrogen.wrapOutput`).toString()}
       >
         <Display output={output} />
       </div>
--- a/lib/components/result-view/history.js
+++ b/lib/components/result-view/history.js
@@ -47,6 +47,7 @@ const History = observer(({ store }: { store: OutputStore }) => {
         style={{
           fontSize: atom.config.get(`Hydrogen.outputAreaFontSize`) || "inherit"
         }}
+        hydrogen-wrapOutput={atom.config.get(`Hydrogen.wrapOutput`).toString()}
       >
         <Display output={output} />
       </div>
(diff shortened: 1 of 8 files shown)


lib/components/result-view/history.js
```diff
@@ -47,6 +47,9 @@ const History = observer(({ store }: { store: OutputStore }) => {
         style={{
           fontSize: atom.config.get(`Hydrogen.outputAreaFontSize`) || "inherit"
         }}
+        hydrogen-wrapOutput={Boolean(
```
aviatesk: Why is this `Boolean` casting needed ? The type of the config value seems to come as `boolean`

styles/hydrogen.less
```diff
@@ -111,6 +111,12 @@ svg:first-child {
 .multiline-container {
   display: flex;
   border: solid 1px @base-border-color;
+  &[hydrogen-wrapoutput="true"] {
+    code, pre {
+      white-space: pre-wrap;
+      word-wrap: break-word;
```
aviatesk: Nice to have `word-wrap` here.

lib/config.js
```diff
@@ -65,6 +65,15 @@ const Config = {
       default: true,
       order: 4
     },
+    wrapOutput: {
+      title: "Enable Soft Wrap for Output",
+      includeTitle: false,
+      description:
+        "If enabled, your output code from hydrogen will break long text and items.",
```
aviatesk: This is very small thing, but `hydrogen` would be better to be camel-case for consistency with other config descriptions: `Hydrogen`

lib/components/result-view/history.js
wadethestealth: For both flow, and for some reason when running the spec atom.config is apparently undefined so it cast undefined to false for specs.

lib/config.js
wadethestealth: Agreed.

lib/components/result-view/history.js
aviatesk: I see.

And I prefer just `wrapOutput` rather than `hydrogen-wrapOutput` since it's just an attribution like `style`, `onClick` etc.

styles/hydrogen.less
```diff
@@ -111,6 +111,12 @@ svg:first-child {
 .multiline-container {
   display: flex;
   border: solid 1px @base-border-color;
+  &[hydrogen-wrapoutput="true"] {
```
aviatesk: Maybe better to be `wrapOutput` (note the case of `Output`) even tho the Less might be case-insensitive.

lib/components/result-view/history.js
wadethestealth: The only reason I prefixed it was because of setting standard because what if someone unknowing later uses a common attribute or what if another package uses our same technique. I just have it there so there is no possible confusions.

styles/hydrogen.less
wadethestealth: If less isn't case sensitive I will change, however if it is I will leave it as is

lib/components/result-view/history.js
aviatesk: I'm not sure about that because `hydrogen-wrapOutput` is not a class name and thus if an user's stylesheet or the other packages's one has defined the style for that, I think this won't get impact by that.
But maybe I miss some point, and it's a small thing thus you can ignore the recommendation.

wadethestealth: The point is if someone else's package has a wrapoutput attribute and we have a wrapoutput attribute behaviors could get screwy because attribute aren't for css only they can be accessed from js. Of course this means another package would have to have a wrapoutput which is unlikely, but it is just a precaution.

styles/hydrogen.less
aviatesk: `=` operator is case-insensitive, and I see what you want.

aviatesk:
@wadethestealth 
I tweaked flow types so that the code body can be more simpler, but you can revert this if you really don't like it.

wadethestealth:
@aviatesk I don't mind it, but it feels pointless unless you can get the specs to work without using Boolean wrapper

aviatesk:
Ahh 😅 
I forgot about the specs probs you said. Seems better not to tweak specs just for this.
Sorry I gonna revert them.

wadethestealth:
@aviatesk very well done! Ready for merge if you are.

aviatesk:
Thanks.
I feel like what @BenRussert thinks on this PR first before we merge this tho 😃

aviatesk:
@wadethestealth @BenRussert 
I think this PR won't cause the problem and reverts the default setting of "wrapping" introduced in v2.10.0, thus I want to merge this now and craft the new release quickly, since I don't want to make things appear inconsistent.

PR #1696, opened 2019-07-03
Fix travis

aviatesk:
### Purpose

Fixes the failure of builds in CI.

- [x] Add `addons` and use the `xenial` image

### Misc

- [x] Split `script` to prevent it from being streamed from the remote server
- ~~[ ] Run only on `master` branch to prevent double builds on PRs~~ We basically only use `master` branch thus never mind
- ~~[ ] Split project-specific config vs. general setups~~ Out travis config is so simple and no need to do this

### References

- https://github.com/atom/ci/issues/90
- https://github.com/atom/wrap-guide/pull/84

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: generic
-dist: trusty
-sudo: false
+
 notifications:
   email:
     on_success: never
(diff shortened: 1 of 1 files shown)


wadethestealth:
Great work @aviatesk

PR #1701, opened 2019-07-07
Watches Syntax Highlighting and Autocomplete

### LINKED ISSUE #1557: Hydrogen Watch Pane could offer suggestions/auto-completes
When I brought up the Hydrogen Watch pane to add some code to inspect an object and use some built-in Python functions such as `print()`, I was surprised the entry did not provide any suggestions or auto-completions such as those that occur in the editor.

I could understandingly see a floating box with the suggestions useful if placed below the last line of the entry box.

### LINKED ISSUE #1558: Hydrogen Watch Pane could use syntax highlighting
The Hydrogen Watch pane offers an area to input code (such as `plt.plot(x, sin(x)); plt.show()` as the given example), but these entries do not offer any syntax highlighting.  It would seem both consistent and convenient to offer syntax highlighting on this code with respect to the editor itself.

wadethestealth:
# 🌟 Watches Enhancements
- Watches Syntax Highlighting #1558 
- Watches Autocomplete #1557 

## Concerns
- Making sure clean up works

## To-Do
- [x] Clean up if autocomplete is deactivated
- [x] Add autocomplete to already existing watches on package activation
---
Closes #1558 
Closes #1557

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -267,10 +267,7 @@ const Hydrogen = {
 
   /*-------------- Service Consumers --------------*/
   consumeAutocompleteWatchEditor(watchEditor: Function) {
-    return services.consumed.autocomplete.consumeWatchEditor(
-      store,
-      watchEditor
-    );
+    return services.consumed.autocomplete.consume(store, watchEditor);
   },
 
   consumeStatusBar(statusBar: atom$StatusBar) {
--- a/lib/main.js
+++ b/lib/main.js
@@ -267,22 +267,18 @@ const Hydrogen = {
 
   /*-------------- Service Consumers --------------*/
   consumeAutocompleteWatchEditor(watchEditor: Function) {
-    const disposable = services.consumed.autocomplete.consumeWatchEditor(
+    return services.consumed.autocomplete.consumeWatchEditor(
       store,
       watchEditor
     );
-    store.subscriptions.add(disposable);
-    return disposable;
   },
 
   consumeStatusBar(statusBar: atom$StatusBar) {
-    const disposable = services.consumed.statusBar.addStatusBar(
+    return services.consumed.statusBar.addStatusBar(
       store,
       statusBar,
       this.handleKernelCommand.bind(this)
     );
-    store.subscriptions.add(disposable);
-    return disposable;
   },
   /*-----------------------------------------------*/
 
--- a/lib/main.js
+++ b/lib/main.js
@@ -266,6 +266,15 @@ const Hydrogen = {
   /*-----------------------------------------------*/
 
   /*-------------- Service Consumers --------------*/
+  consumeAutocompleteWatchEditor(watchEditor: Function) {
+    const disposable = services.consumed.autocomplete.addWatchEditor(
+      store,
+      watchEditor
+    );
+    store.subscriptions.add(disposable);
+    return disposable;
+  },
+
   consumeStatusBar(statusBar: atom$StatusBar) {
     const disposable = services.consumed.statusBar.addStatusBar(
       store,
--- a/lib/main.js
+++ b/lib/main.js
@@ -35,6 +35,7 @@ import kernelManager from "./kernel-manager";
 import ZMQKernel from "./zmq-kernel";
 import WSKernel from "./ws-kernel";
 import Kernel from "./kernel";
+import type MarkerStore from "./store/markers";
 import HydrogenProvider from "./plugin-api/hydrogen-provider";
 import services from "./services";
 
@@ -284,14 +285,8 @@ const Hydrogen = {
   },
 
   handleKernelCommand(
-    {
-      command,
-      payload
-    }: {
-      command: string,
-      payload: ?Kernelspec
-    },
-    { kernel, markers }
+    { command, payload }: { command: string, payload: ?Kernelspec },
+    { kernel, markers }: { kernel: ?Kernel, markers: ?MarkerStore }
   ) {
     log("handleKernelCommand:", arguments);
 
--- a/lib/main.js
+++ b/lib/main.js
@@ -267,7 +267,7 @@ const Hydrogen = {
 
   /*-------------- Service Consumers --------------*/
   consumeAutocompleteWatchEditor(watchEditor: Function) {
-    const disposable = services.consumed.autocomplete.addWatchEditor(
+    const disposable = services.consumed.autocomplete.consumeWatchEditor(
       store,
       watchEditor
     );
--- a/lib/services/consumed/autocomplete.js
+++ b/lib/services/consumed/autocomplete.js
@@ -1,0 +1,50 @@
+/* @flow */
+import { CompositeDisposable, Disposable } from "atom";
+
+import type { Store } from "../../store";
+
+export const autocompleteProvider = {
+  disposables: new CompositeDisposable(),
+  watchEditor: (editor: atom$TextEditor, labels: Array<string>) => {
+    return;
+  }
+};
+window.autocomplete = autocompleteProvider;
+
+export function addWatchEditor(
+  store: Store,
+  watchEditor: atom$AutocompleteWatchEditor
+) {
+  autocompleteProvider.watchEditor = watchEditor;
+  for (let kernel of store.runningKernels) {
+    for (let watch of kernel.watchesStore.watches) {
+      const disposable = watchEditor(this.editor, [
+        "default",
+        "workspace-center",
+        "symbol-provider"
+      ]);
+      if (disposable) {
+        watch.autocompleteDisposable = disposable;
+        kernel.watchesStore.autocompleteDisposables.add(disposable);
+      }
+    }
+    autocompleteProvider.disposables.add(
+      kernel.watchesStore.autocompleteDisposables
+    );
+  }
+  return new Disposable(this.disableAutocomplete.bind(this, store));
+}
+
+export function disableAutocomplete(store: Store) {
+  autocompleteProvider.watchEditor = () => {
+    return;
+  };
+  for (let kernel of store.runningKernels) {
+    for (let watch of kernel.watchesStore.watches) {
+      watch.autocompleteDisposable = null;
+    }
+    kernel.watchesStore.autocompleteDisposables = new CompositeDisposable();
+  }
+  autocompleteProvider.disposables.dispose();
+  autocompleteProvider.disposables = new CompositeDisposable();
+}
--- a/lib/services/consumed/autocomplete.js
+++ b/lib/services/consumed/autocomplete.js
@@ -9,7 +9,7 @@ import type WatchStore from "../../store/watch";
  * This acts as a global storage for the consumed service.
  */
 export class AutocompleteWatchEditor {
-  disposables = new CompositeDisposable();
+  disposables: atom$CompositeDisposable;
   /** The `consumed autocompleteWatchEditor` */
   addAutocompleteToEditor = (
     editor: atom$TextEditor,
@@ -27,6 +27,7 @@ export class AutocompleteWatchEditor {
    * @returns {Disposable} This is for clean up when autocomplete or hydrogen deactivate.
    */
   consume(store: Store, watchEditor: atom$AutocompleteWatchEditor) {
+    this.disposables = new CompositeDisposable();
     this.addAutocompleteToEditor = watchEditor;
 
     // Add autocomplete capabilities to already existing watches
@@ -34,7 +35,9 @@ export class AutocompleteWatchEditor {
       for (let watch of kernel.watchesStore.watches) {
         this.addAutocompleteToWatch(kernel.watchesStore, watch);
       }
-      this.disposables.add(kernel.watchesStore.autocompleteDisposables);
+      const watchesStoreDisposable = new CompositeDisposable();
+      kernel.watchesStore.autocompleteDisposables = watchesStoreDisposable;
+      this.disposables.add(watchesStoreDisposable);
     }
 
     const disposable = new Disposable(() => this.disable(store));
@@ -65,12 +68,11 @@ export class AutocompleteWatchEditor {
       for (let watch of kernel.watchesStore.watches) {
         watch.autocompleteDisposable = null;
       }
-      kernel.watchesStore.autocompleteDisposables = new CompositeDisposable();
+      kernel.watchesStore.autocompleteDisposables = null;
     }
 
     // Disables autocomplete, Cleans up everything, and Resets.
     this.disposables.dispose();
-    this.disposables = new CompositeDisposable();
   }
 
   /**
--- a/lib/services/consumed/autocomplete.js
+++ b/lib/services/consumed/autocomplete.js
@@ -17,7 +17,7 @@ export function addWatchEditor(
   autocompleteProvider.watchEditor = watchEditor;
   for (let kernel of store.runningKernels) {
     for (let watch of kernel.watchesStore.watches) {
-      const disposable = watchEditor(this.editor, [
+      const disposable = watchEditor(watch.editor, [
         "default",
         "workspace-center",
         "symbol-provider"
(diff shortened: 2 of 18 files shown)


lib/main.js
```diff
@@ -260,6 +260,10 @@ const Hydrogen = {
     return this.hydrogenProvider;
   },
 
+  consumeWatchEditor(watchEditor: Function) {
+    return store.enableAutocomplete(watchEditor);
+  },
```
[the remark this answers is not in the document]
wadethestealth: This is called *every time* `autocomplete-plus` is activated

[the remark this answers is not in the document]
wadethestealth: If it is already active when `hydrogen` actives it is also called

lib/store/index.js
```diff
@@ -288,6 +289,21 @@ export class Store {
     this.kernelMapping.set(newKey, this.kernelMapping.get(oldKey));
     this.kernelMapping.delete(oldKey);
   }
+
+  enableAutocomplete(watchEditor: Function) {
+    this.addAutocomplete = watchEditor;
+    for (let kernel of this.runningKernels) {
+      kernel.watchesStore.addAutocomplete();
```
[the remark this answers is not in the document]
wadethestealth: I was thinking about making a store for other package consumption, which could separate hydrogen logic from package add-on logic since it is not essential to hydrogen running. Let me know what you think about that. 
As for the logic, if it doesn't live here I believe it would have to live in the `main.js` if we left everything else the same. The reason being is because there is no way to access the `watchesStore` without the `kernels` which a stored in the store. Like I was saying above though I could technically move it to a new file (store or no store) I just cant encapsulate all of the code in watchesStore.

spec/store/watch-spec.js
```diff
@@ -1,6 +1,10 @@
 "use babel";
 
+import { Disposable } from "atom";
+
+import { waitAsync } from "../helpers/test-utils";
 import WatchStore from "../../lib/store/watch";
+import _store from "../../lib/store";
```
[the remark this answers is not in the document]
wadethestealth: Good call

spec/store/watch-spec.js
```diff
@@ -52,4 +56,17 @@ describe("WatchStore", () => {
       expect(store.editor.element.focus).toHaveBeenCalled();
     });
   });
+  describe("addAutocomplete", () => {
+    beforeEach(
+      waitAsync(async () => {
+        const p = await atom.packages.activatePackage("autocomplete-plus");
+        //modelled after consumeWatchEditor in lib/main.js
+        _store.enableAutocomplete(p.mainModule.provideWatchEditor());
```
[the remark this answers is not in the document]
wadethestealth: agreed

lib/store/watch.js
```diff
@@ -1,25 +1,30 @@
 /* @flow */
 
-import { TextEditor } from "atom";
+import { TextEditor, Disposable } from "atom";
```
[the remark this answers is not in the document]
wadethestealth: Actually no imports are needed it was just left for type checking forgot about `atom$` in flow. Thanks

spec/store/watch-spec.js
```diff
@@ -1,6 +1,10 @@
 "use babel";
 
+import { Disposable } from "atom";
+
+import { waitAsync } from "../helpers/test-utils";
```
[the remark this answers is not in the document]
wadethestealth: Yes I saw the jest, for another PR I may migrate us

spec/store/watch-spec.js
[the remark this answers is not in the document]
wadethestealth: As it stands I can't not use the global store since thats what the code inside `watchStore` uses, but I could instead just always be passing the `watchEditor` function down the chain to avoid using the global store

lib/services/consumed/autocomplete.js
```diff
@@ -0,0 +1,49 @@
...
+};
+
+export function addWatchEditor(
+  store: Store,
+  watchEditor: atom$AutocompleteWatchEditor
+) {
+  autocompleteProvider.watchEditor = watchEditor;
+  for (let kernel of store.runningKernels) {
+    for (let watch of kernel.watchesStore.watches) {
+      const disposable = watchEditor(this.editor, [
```
BenRussert: What does `this.editor` refer to here? I get nervous with `this` outside of a `class`.

lib/services/consumed/autocomplete.js
```diff
@@ -0,0 +1,49 @@
...
+      if (disposable) {
+        watch.autocompleteDisposable = disposable;
+        kernel.watchesStore.autocompleteDisposables.add(disposable);
+      }
+    }
+    autocompleteProvider.disposables.add(
+      kernel.watchesStore.autocompleteDisposables
+    );
+  }
+  return new Disposable(this.disableAutocomplete.bind(this, store));
```
BenRussert: Do we need to bind `this` and wouldn't `this.disableAutocomplete` be the same as `disableAutocomplete` since we are not inside of a `class`? 

In other words, would this work?

```js
return new Disposable(() => disableAutocomplete(store));
```

lib/services/consumed/autocomplete.js
```diff
@@ -0,0 +1,49 @@
...
+
+import type { Store } from "../../store";
+
+export const autocompleteProvider = {
+  disposables: new CompositeDisposable(),
+  watchEditor: (editor: atom$TextEditor, labels: Array<string>) => {
+    return;
+  }
+};
+
```
BenRussert: One thing the code base could improve on is writing comments for exported objects and functions. This would help down the line if someone wants to understand the purpose of `autocompleteProvider` for example.

What I have in mind is informal jsdoc style (so it will show up in text editor tooltips on hover):

```js
/** 
* Stores and handles cleanup of...
* Add x to this every time you do y
* 
*/
```
What do you think?

lib/services/consumed/autocomplete.js
wadethestealth: left from when I ported code. It works weirdly enough, but should be `watch.editor`

lib/services/consumed/autocomplete.js
wadethestealth: Sure will add. I would let @aviatesk know for #1710 too.

lib/services/consumed/autocomplete.js
wadethestealth: I'll do some testing 😃

wadethestealth: oh wait yes this work. I had it like that originally, but I thought my way better. After some thought, I like your way better, cleaner and almost exactly the same.

lib/services/consumed/autocomplete.js
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * Adds a disposable as an autocomplete disposable.
+ *
+ * @param {Disposable | CompositeDisposable} disposable
+ */
+export function registerAutocompleteDisposable(
+  disposable: Disposable | CompositeDisposable
+) {
+  _autocompleteConsumer.disposables.add(disposable);
+}
```
BenRussert: I think this file and it's exported functions might be better off as a class, but I wouldn't block the PR over that just wondered if you had thoughts on a class vs a set of functions.

lib/services/consumed/status-bar/status-bar.js
```diff
@@ -25,14 +25,16 @@ export function addStatusBar(
     priority: 100
   });
 
-  const onClick = showKernelCommands.bind(this, handleKernelCommand);
+  const onClick = _showKernelCommands.bind(this, handleKernelCommand);
```
BenRussert: I worry about using `this`/`bind` outside of a class because I think it might be confusing to people who don't know the in's and outs of js `this`. Again, I wouldnt block on this but I think using either a class or: `let signalListView;` at the top of the file and referring to `signalListView` instead of `this.signalListView`

types/atom.js.flow
```diff
@@ -2055,6 +2056,10 @@ type atom$AutocompleteProvider = {
   ) => void,
 };
 
+type atom$AutocompleteWatchEditor = (
+  editor: atom$TextEditor, labels: Array<string>
+) => ?atom$Disposable;
+
```
BenRussert: I was not aware of this api and looked into it. I think we should put this in another file because it doesnt look like core atom api. Maybe even just at the top of `autocomplete.js`. The api is part of [autocomplete plus](https://github.com/atom/autocomplete-plus#the-watcheditor-api). We might want to link to that because `WatchEditor` api there might have ambiguous meaning since it's another type of `watch`

lib/services/consumed/autocomplete.js
wadethestealth: I could make this a class for easier naming purposes

lib/services/consumed/status-bar/status-bar.js
wadethestealth: Yeah I could probably class this too

types/atom.js.flow
wadethestealth: Good idea I like it in file.

aviatesk: I gonna introduce another non-core atom apis in #1710. 
So I think we can just make a file like `atom-pkgs.js.flow` (or i don't know, `atom-pkg-services.js.flow` ?) and moves atomcomplete types into this, that might make things more clear ?

lib/services/consumed/autocomplete.js
```diff
@@ -0,0 +1,130 @@
...
+    return;
+  };
+
+  /**
+   * This function is called on activation of autocomplete, or if autocomplete is
+   * already active, then it is called when hydrogen activates.
+   *
+   * @param {Store} store - The global Hydrogen store.
+   * @param {Function} watchEditor - The function provided by `autocomplete.watchEditor`.
+   * @returns {Disposable} This is for clean up when autocomplete or hydrogen deactivate.
```
aviatesk: `-` between `{Disposable}` and `This` ?
Very small thing tho.

types/atom.js.flow
wadethestealth: I was thinking of creating a `hydrogen-services.js.flow` where all types follow the format `hs_<package>$<type name>` hs stands for hydrogen services, and that way we know we typed this. package so we know where it is coming from in case the type every changes. type name is pretty self explanatory. What do you guys think? @aviatesk @BenRussert

wadethestealth: Services should be the only place we every require types outside atom or electron so thats why I think we should label it that way

lib/services/consumed/autocomplete.js
```diff
@@ -0,0 +1,130 @@
...
+   */
+  consume(store: Store, watchEditor: atom$AutocompleteWatchEditor) {
+    this.addAutocompleteToEditor = watchEditor;
+
+    // Add autocomplete capabilities to already existing watches
+    for (let kernel of store.runningKernels) {
```
aviatesk: It's just a talk of code style, but maybe `forEach` is preferred since we usually use that in most part of the other code base.

But if you prefer native `for` iterator for some reason, go with your style :wink:

lib/services/consumed/autocomplete.js
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  /**
+   * This function is just for cleaning up when either autocomplete or hydrogen is deactivating.
+   *
+   * @param {Store} store - The global Hydrogen store.
+   */
+  disable(store: Store) {
+    // Removes the consumed function `watchEditor`
+    this.addAutocompleteToEditor = (
```
aviatesk: I think something like
```js
if (this.addAutocompleteToEditor) this.addAutocompleteToEditor.dispose();
```
would do the trick ?

lib/services/consumed/autocomplete.js
```diff
@@ -0,0 +1,130 @@
...
+    /*
+     * Removes disposables from all watches (leaves references inside
+     * `this.disposables` to be disposed at the end).
+     * Autocomplete is only actually disabled on dispose of `this.disposables`
+     */
+    for (let kernel of store.runningKernels) {
+      for (let watch of kernel.watchesStore.watches) {
+        watch.autocompleteDisposable = null;
+      }
+      kernel.watchesStore.autocompleteDisposables = new CompositeDisposable();
```
aviatesk: I think it would be clearer if you move these `new CompositeDisposable` stuff into `consume`, since these kind of actions are not `disabling`.

aviatesk:
IMHO `hydrogen_services.js.flow` can sound like the types for Hydrogen API definitions.
Rather I prefer `atom-services.js.flow` or just make each file for each service, like @BenRussert said with `autocomplete.js.flow`.

lib/services/consumed/autocomplete.js
wadethestealth: No its worth it 🤣

lib/services/consumed/autocomplete.js
wadethestealth: If you look at the type declaration of `addAutocompleteToEditor` at the top of the class its a function. I am replacing the function I consumed with one that does nothing to avoid doing null/undefined checks everywhere it is used.

types/atom.js.flow
```diff
@@ -1535,6 +1535,7 @@ declare class atom$GrammarRegistry {
   removeGrammarForScopeName(scopeName: string): ?atom$Grammar,
   loadGrammarSync(grammarPath: string): atom$Grammar,
   selectGrammar(filePath: string, fileContents: string): atom$Grammar,
+  assignLanguageMode(bufferOrEditor: atom$TextBuffer | atom$TextEditor, languageId:string): boolean,
```
aviatesk: I would appreciate if you fix this: just by adding a spece.
image

lib/services/consumed/autocomplete.js
aviatesk: Yeah, but if you `this.addAutocompleteToEditor.dispose();`, I believe the actual function object would stay there and behaves like the function returning `null` on call.

lib/services/consumed/autocomplete.js
wadethestealth: just because its nested I think it is more clear

wadethestealth:
lets do @BenRussert's flow file reccomendation

wadethestealth:
Going to take a break. Only bits left to do are to move flow types and update the old test case thats failing

lib/services/consumed/autocomplete.js
wadethestealth: Dispose doesn't exist in the function prototype

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/prototype

Is it apart of something else?

wadethestealth: As far as I know dispose is a custom method of atom disposable and compositedisposable class

aviatesk: NVM. I misunderstood that `this.addAutoCompleteToEditor` itself is disposable.

lib/store/watches.js
```diff
@@ -12,20 +14,23 @@ export default class WatchesStore {
   kernel: Kernel;
   @observable
   watches: Array<WatchStore> = [];
+  autocompleteDisposables: ?atom$CompositeDisposable;
 
   constructor(kernel: Kernel) {
     this.kernel = kernel;
 
     this.kernel.addWatchCallback(this.run);
+    AutocompleteConsumer.register(this.autocompleteDisposables);
```
aviatesk: This part causes obvious and serious errors.

AFAIU `this.autocompleteDisposables` in watches.js does nothing other than that currently.
Can you explain your purpose of this `register` stuff ?

wadethestealth: Oh I remember why the new ones where in disable now too. The purpose of register is to notify the consumer class that this disposable must be disposed when autocomplete is disabled.

wadethestealth: The error is because now it is no longer predefined I need to create a new disposable right above that line

wadethestealth: `this.autocompleteDisposables = new CompositeDisposable`

wadethestealth: I would much rather revert my disable commit though because that way there is always a disposable even if it is empty because there is no autocomplete it will still be there

aviatesk: I see.
Don't you think `this.autocompleteDisposables = new CompositeDisposable();` in class declaration is not enough tho ?

wadethestealth: I am not to sure. I removed it during my disable commit but maybe I shouldn't have. It was originally where the type declaration is.

PR #1703, opened 2019-07-08
Improve Plotly Support

### LINKED ISSUE #1653: Downloaded files from plotly saved to nowhere
When creating a plot using plotly:

```
from plotly import offline
offline.init_notebook_mode()

offline.iplot([{"y": [1, 2, 1]}])
```

When clicking the "Download plot as png" button:

Screenshot from 2019-06-05 15-54-54

The messages
"Taking Snapshot - this may take a few seconds"
"Snapshot succeeded - newplot.png"
are displayed:

Screenshot from 2019-06-05 15-50-45

but this newplot.png file does not exist. In Jupyter this would cause your browser to download the image, but in Atom, it seems like it doesn't do anything...

wadethestealth:
# 🌟 Improve Plotly Support 
*This pull request contains modified code from @nteract/transform-plotly*

### Fixes:
- `Download plot as png` button was previously broken only inside an atom context, so using built-in electron functions it has been refactored to work. See #1653.
- While on our master branch, using simple plotly code (example from nteract.io), and output directed to the output pane, the plotly would only render a quarter of itself. This adds styling to override that behavior.

### To-Do:
- [x] test cases modified from @nteract/transform-plotly.

### Incompatible PRs:
- #1699 
---
Closes #1653

### DIFF
--- a/lib/components/result-view/display.js
+++ b/lib/components/result-view/display.js
@@ -22,7 +22,7 @@ export const supportedMediaTypes = (
   <RichMedia>
     <Vega3 />
     <Vega2 />
-    <Plotly />
+    <Plotly data="" />
     <VegaLite2 />
     <VegaLite1 />
     <Media.Json />
--- a/lib/components/result-view/display.js
+++ b/lib/components/result-view/display.js
@@ -12,7 +12,7 @@ import {
   Media,
   RichMedia
 } from "@nteract/outputs";
-import PlotlyTransform from "@nteract/transform-plotly";
+import Plotly from "./plotly";
 import { VegaLite1, VegaLite2, Vega2, Vega3 } from "@nteract/transform-vega";
 
 import Markdown from "./markdown";
@@ -22,7 +22,7 @@ export const supportedMediaTypes = (
   <RichMedia>
     <Vega3 />
     <Vega2 />
-    <PlotlyTransform />
+    <Plotly />
     <VegaLite2 />
     <VegaLite1 />
     <Media.Json />
--- a/lib/components/result-view/display.js
+++ b/lib/components/result-view/display.js
@@ -22,7 +22,7 @@ export const supportedMediaTypes = (
   <RichMedia>
     <Vega3 />
     <Vega2 />
-    <Plotly data="" />
+    <Plotly />
     <VegaLite2 />
     <VegaLite1 />
     <Media.Json />
--- a/lib/components/result-view/plotly.js
+++ b/lib/components/result-view/plotly.js
@@ -38,12 +38,6 @@ declare class PlotlyHTMLElement extends HTMLDivElement {
   redraw: () => void;
 }
 
-/*
- * As part of the init notebook mode, Plotly sneaks a <script> tag in to load
- * the plotlyjs lib. We have already loaded this though, so we "handle" the
- * transform by doing nothing and returning null.
- */
-
 export class PlotlyTransform extends React.Component<Props> {
   static defaultProps = {
     data: "",
--- a/lib/components/result-view/plotly.js
+++ b/lib/components/result-view/plotly.js
@@ -1,0 +1,148 @@
+/* @flow */
+/**
+ * Adapted from https://github.com/nteract/nteract/blob/master/packages/transform-plotly/src/index.tsx
+ * Copyright (c) 2016 - present, nteract contributors
+ * All rights reserved.
+ *
+ * This source code is licensed under the license found in the LICENSE file in
+ * the root directory of this source tree.
+ *
+ * @NOTE: This `PlotlyTransform` component could be used exactly same as the original `PlotlyTransform` component of @nteract/transform-plotly,
+ *        except that this file adds the ability to download a plot from an electron context.
+ */
+
+import { cloneDeep } from "lodash";
+import React from "react";
+
+interface Props {
+  data: string | object;
+  mediaType: "application/vnd.plotly.v1+json";
+}
+
+type ObjectType = object;
+
+interface FigureLayout extends ObjectType {
+  height?: string;
+  autosize?: boolean;
+}
+
+interface Figure extends ObjectType {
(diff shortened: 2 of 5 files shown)


lib/components/result-view/plotly.js
```diff
@@ -0,0 +1,148 @@
+/* @flow */
```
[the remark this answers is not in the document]
wadethestealth: I am moving all custom transforms to a separate folder in the transform api PR. Also I believe that the error is sent through a different component than execute result or display data

lib/components/result-view/plotly.js
```diff
@@ -0,0 +1,148 @@
...
+    this.Plotly.newPlot(this.plotDiv, figure.data, figure.layout, {
+      modeBarButtonsToRemove: ["toImage"],
+      modeBarButtonsToAdd: [
+        {
+          name: "Download plot as a png",
+          icon: this.Plotly.Icons.camera,
+          click: this.downloadImage
+        }
+      ]
+    });
```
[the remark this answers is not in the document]
wadethestealth: I have not technically confirmed, but I believe the package goal (extracted from the readme) is to work in react (not just electron), which is does work in react. Also I don't know what nteract app does with links, but I could test it. The problem is atom (not all electron apps) don't provide ability to click links and open/download the contents, which is how plotly handles the download.

lib/components/result-view/display.js
```diff
@@ -22,7 +22,7 @@ export const supportedMediaTypes = (
   <RichMedia>
     <Vega3 />
     <Vega2 />
-    <PlotlyTransform />
+    <Plotly data="" />
```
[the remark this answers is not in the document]
wadethestealth: Good idea it was because of flow throwing an error that I added it.

lib/components/result-view/plotly.js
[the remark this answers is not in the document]
wadethestealth: Just confirmed nteract behaves as expected and does not require this change

lib/components/result-view/plotly.js
```diff
@@ -0,0 +1,149 @@
...
+interface Figure extends ObjectType {
+  data: Object;
+  layout: FigureLayout;
+}
+
+declare class PlotlyHTMLElement extends HTMLDivElement {
+  data: Object;
+  layout: Object;
+  newPlot: () => void;
+  redraw: () => void;
```
BenRussert: I always glad to see this instead of `Function`

lib/components/result-view/plotly.js
```diff
@@ -0,0 +1,149 @@
...
+}
+
+const NULL_MIMETYPE = "text/vnd.plotly.v1+html";
```
BenRussert: We can remove the `NULL_MIMETYPE` stuff, I believe.

lib/components/result-view/plotly.js
```diff
@@ -0,0 +1,149 @@
...
+PlotlyNullTransform.defaultProps = {
+  mediaType: NULL_MIMETYPE
+};
+
+export class PlotlyTransform extends React.Component<Props> {
+  static MIMETYPE = MIMETYPE;
+
+  static defaultProps = {
+    data: "",
+    mediaType: MIMETYPE
```
BenRussert: You could just hardcode the `MIMETYPE` string here, `@nteract/outputs` looks for `mediaType` and only the jupyter extension still uses `MIMETYPE` as far as I know (we don't need it for hydrogen)

The reason I mention it is that [`MIME` is a term that has been replaced with `Media Type`](https://www.iana.org/assignments/media-types/media-types.xhtml).

lib/components/result-view/plotly.js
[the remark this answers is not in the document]
BenRussert: Awesome, thanks for checking that. I wonder why that is? Electron version or some atom csp?

spec/components/result-view/plotly-spec.js
```diff
@@ -0,0 +1,123 @@
...
+          {
+            name: "Download plot as a png",
+            icon: plotly.Icons.camera,
+            click: instance.downloadImage
+          }
+        ]
+      }
+    );
+  });
+
```
BenRussert: I'm not so sure how useful some of the tests ported over from the monorepo would be. Do you think they are useful for us here?

wadethestealth: I think that they are useful since we will basically have to make sure we stay up to date with plotly and don't break this component

lib/components/result-view/plotly.js
wadethestealth: Agreed

lib/components/result-view/plotly.js
wadethestealth: Ok will do

wadethestealth:
Yes I prefer to merge this first @BenRussert

package.json
```diff
@@ -63,7 +63,7 @@
     "@nteract/markdown": "^4.0.0",
     "@nteract/mathjax": "^4.0.0",
     "@nteract/outputs": "^2.1.5",
-    "@nteract/transform-plotly": "^5.0.0",
+    "@nteract/plotly": "^1.48.3",
```
aviatesk: So mainly this dependency is needed only for specs ?
If so, let's move this into `devDependency`

aviatesk: (For better loading time)

wadethestealth: @aviatesk no it's needed for both

aviatesk: You mean it's implicitly imported somewhere ?

wadethestealth: Inside result-view/plotly.js

aviatesk: Ah I missed it. I see.

BenRussert:
Looks like @aviatesk's feedback was addressed so I will merge this now.

aviatesk:
Nice work @wadethestealth 😉

Issue #1415, opened 2018-09-16
Tooltip for introspection messages


kylebarron:
What would others think about a PR that implements introspection as a tooltip? I'm thinking of something similar to the tooltip shown in @lgeiger's `ide-python` package. It could be turned off in the settings if desired (i.e. if introspection takes too long).

This could reduce the barrier for something like this for languages that don't have a language server implementation.

![image](https://user-images.githubusercontent.com/15164633/45591881-6ee57b00-b92d-11e8-9661-8dd611f11f7b.png)

Related to issue #371.

kylebarron:
I looked into this a little bit more. `ide-python` doesn't create the tooltip itself; it just provides the connection for Atom IDE to start the Python language server and communicate. So I think we'd have to connect directly to the Atom IDE UI package.

The datatip library is here: [`modules/atom-ide-ui/pkg/atom-ide-datatip`](https://github.com/facebook-atom/atom-ide-ui/tree/master/modules/atom-ide-ui/pkg/atom-ide-datatip).

The datatip part of Atom IDE UI is called from `atom-languageclient` here:
- https://github.com/atom/atom-languageclient/blob/7a52fbb774422cdc839614ace613086294af6b1f/lib/auto-languageclient.ts
- https://github.com/atom/atom-languageclient/blob/3f8a3ae62281fc89410ff806fe16e03773e5b9f3/lib/adapters/datatip-adapter.ts

lgeiger:
I really like the idea to connect to the atom-ide-ui package, mainly because we don't need to build and maintain the UI for it ;)

One thing we need to check is, that it doesn't interfere with other running language servers.

kylebarron:
> One thing we need to check is, that it doesn't interfere with other running language servers.

This is a good point. There can be an option in the settings, at least, to turn tooltips on and off.

PR #1706, opened 2019-07-11
Refactor All Services APIs

wadethestealth:
## Description: 
This is a sample of us consuming the `Atom Status Bar` and providing `Autocomplete` outside of `main.js`.

**Notes:**
- Fixed the bad disposable return of the Status Bar API
- Removes some imports of `store` in favor of a `store` reference passed as an arg
- Autocomplete is now dynamically responsive to settings changes (previously required either autocomplete or hydrogen to be re-activated)

## Motivation:
Whilst reviewing #1701, @BenRussert asked if there was any way to remove the autocomplete code out of our `Store`. After brainstorming, I came up with a way to move all of the Service API related code to its own unique folder separate from the main code. 

## Analysis:
Pros:
   - Main code is separate from "extra" features of hydrogen aka `Services`/reliance on other packages
   - `main.js` consumption and providing will become much more clear
   - Reduces the amount of non-essential code in `main.js`

Cons:
   - Some tasks require `Function.prototype.bind`, which make it easier, but can be slightly confusing to new contributors of `Services`

### New file structure:
`./lib/services/consumed/` & `./lib/services/provided`

### Hydrogen Plugin API
Should the Hydrogen Plugin-API remain in its own folder?
   - It is technically a provider service, but it may be better to keep it separate for easy access and so that it is easily distinguishable.

### Goals for This PR
- [x] Status Bar
   - [x] Move our `consumer` to `services`
- [x] Autocomplete Plus
   - [x] Move our `provider` to `services`

### Goals for Future PR's
- [ ] Add Autocomplete to Watches #1701


lib/services/status-bar/status-bar.js
```diff
@@ -3,23 +3,25 @@
...
 
 type Props = {
-  store: { kernel: ?Kernel, configMapping: Map<string, mixed> },
+  store: Store,
   onClick: Function
 };
 
-const StatusBar = observer(
-  ({ store: { kernel, configMapping }, onClick }: Props) => {
+@observer
```
BenRussert: Is this change prep for #1284?

package.json
```diff
@@ -91,17 +91,12 @@
       "versions": {
         "^1.0.0": "consumeStatusBar"
       }
-    },
-    "tree-view": {
-      "versions": {
-        "^1.0.0": "consumeTreeView"
-      }
```
BenRussert: Any idea what this was for?

lib/main.js
```diff
@@ -37,8 +35,8 @@ import kernelManager from "./kernel-manager";
 import ZMQKernel from "./zmq-kernel";
 import WSKernel from "./ws-kernel";
 import Kernel from "./kernel";
-import AutocompleteProvider from "./autocomplete-provider";
 import HydrogenProvider from "./plugin-api/hydrogen-provider";
+import services from "./services";
```
BenRussert: I think adding a short README.md in `./services` explaining what type of thing belongs there would be really useful to both to new contributors and maintainers.

lib/services/status-bar/status-bar.js
wadethestealth: not that PR but the outstanding one #1631

package.json
wadethestealth: No clue, but it isn't being used since that function isn't in main (or anywhere from a file search)

lib/main.js
wadethestealth: Good Idea

package.json
wadethestealth: Looking at the git-blame looks like you added this to add an import notebook command when right-clicking on `.ipynb` files in the tree-view, not sure where this code was moved, but it has been implemented somewhere else as it still works in this PR

lib/services/status-bar/status-bar.js
aviatesk: We may not need to change the status-bar component into class one, and in the future, react components will be preferred to be functional component.
Thus please hold on it a bit.

package.json
aviatesk: This is because we just retrieved file path from class name, not thought via valid tree-view API here: https://github.com/nteract/hydrogen/blame/renovate/mobx-react-6.x/lib/import-notebook.js#L29

I think the current implementation works thus we can leave this as is.

aviatesk:
I like the idea in general too.

> ./lib/services/consumed/ & ./lib/services/provided
This is really clearer and preferable for me IMHO.

And this is also just an suggestion, but how about renaming below ?
- `services/(consumed/)status-bar/index.js` -> `services/(consumed/)status-bar.js`
- `services/(consumed/)status-bar/status-bar.js` -> `services/(consumed/)status-bar-component.js`: The consistency with the other component files
- (The same kind goes for autocomplete)
Since the services are not so much many after all.

lib/main.js
```diff
@@ -261,39 +259,15 @@ const Hydrogen = {
...
-      item: statusBarElement,
-      priority: 100
-    });
-
-    const onClick = this.showKernelCommands.bind(this);
-
-    reactFactory(
-      <StatusBar store={store} onClick={onClick} />,
-      statusBarElement
+    return services.statusBar.addStatusBar(
```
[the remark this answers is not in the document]
BenRussert: Good question. The easiest way to fix that would be to use the `store.subscriptions` composite disposable which gets cleaned up on deactivate. So instead of returning the disposable, just add it to `store.subscriptions.add(new Disposable(() => statusBarTile.destroy()))` inside of `addStatusBar`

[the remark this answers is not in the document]
wadethestealth: This shouldn't be necessary according to how the service api works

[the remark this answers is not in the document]
wadethestealth: Please view the [Service API Docs](https://flight-manual.atom.io/behind-atom/sections/interacting-with-other-packages-via-services/). 

>These methods will be called any time a package is activated that provides their corresponding service. They will receive the service object as an argument. You will usually need to perform some kind of cleanup in the event that the package providing the service is deactivated. To do this, return a Disposable from your service-consuming method:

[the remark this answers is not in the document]
aviatesk: Cool. I thought an new `CompositeDisposables` for consumed services would be better than `store.subscriptions` might be better, but it's just my opinion. We can go with `store.subscriptions` of course.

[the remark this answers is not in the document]
wadethestealth: Why would adding a new one be better though?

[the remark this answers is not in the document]
aviatesk: Umm nvm. I felt like that since I'm currently implementing #1415, and temporary making new subscripition in main.js.

But now I'm sure adding it to `store`'s would be simpler and much better.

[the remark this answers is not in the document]
BenRussert: > in the event that the package providing the service is deactivated. 

Thanks for the link. The difference is if we are talking about hydrogen deactivating (store Composite disposable gets cleaned up) or the status bar package deactivating ( the disposable returned by the consume function gets cleaned up). I was taking about the former case (@aviadesk was also I believe)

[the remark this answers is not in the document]
wadethestealth: It seems very unclear to me what needs a clean up function and what doesn't because I feel like they would have added that to the docs. I asked in the slack and am awaiting a response.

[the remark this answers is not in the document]
BenRussert: It's such a minor thing here I wouldn't worry about it so much. If you're curious try deactivating hydrogen and looking at the dom in devtools after. There will be a Hydrogen status bar element with no size (cant be seen by users). It's been that way forever,but I never thought it was worth looking into.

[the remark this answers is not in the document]
aviatesk: > I was taking about the former case (@aviadesk was also I believe)

My point was exactly what you said there :smile:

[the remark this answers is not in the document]
wadethestealth: From Atom
>@Thomas Denton In the example from the Flight Manual, the cleanup that is performed when the consumed package deactivates is all on the side of the consumer. If the consumer deactivates, then it should already know to clean itself up and shouldn't need to use a `Disposable` to do it.

So we shouldn't need to add it to the store. I'll do some testing though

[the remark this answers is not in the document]
aviatesk: image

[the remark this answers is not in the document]
aviatesk: > If we dispose it when deactivating

image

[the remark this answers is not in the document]
aviatesk: Let's add it to our store.
And note that the disposing an already disposed `Disposable` won't cause the error.

aviatesk:
@wadethestealth 
I changed the flow type notation of autocomplete provider function so that it uses the one in our types directory, since it has more definitions for properties and code can get more simpler.

spec/services/consumed/status-bar-spec.js
```diff
@@ -6,12 +6,13 @@ import Adapter from "enzyme-adapter-react-16";
...
-import KernelTransport from "../../lib/kernel-transport";
-import Kernel from "../../lib/kernel";
-import StatusBar from "../../lib/components/status-bar";
+import { Store } from "../../../lib/store";
+import KernelTransport from "../../../lib/kernel-transport";
+import Kernel from "../../../lib/kernel";
+import { StatusBar } from "../../../lib/services/consumed/status-bar";
 
-describe("Status Bar", () => {
+console.log(StatusBar);
```
aviatesk: Maybe you forget to remove it ?

lib/services/consumed/status-bar/status-bar.js
```diff
@@ -0,0 +1,41 @@
...
+  const onClick = showKernelCommands.bind(this, handleKernelCommand);
+
+  reactFactory(<StatusBar store={store} onClick={onClick} />, statusBarElement);
+
+  return new Disposable(() => statusBarTile.destroy());
+}
+
+function showKernelCommands(handleKernelCommand, kernel) {
+  if (!this.signalListView) {
+    this.signalListView = new SignalListView(kernel);
```
aviatesk: Why should `SignalListView`  turn to be constructed from `kernel` ?
Does this work with multiple kernels ?

lib/services/README.md
```diff
@@ -0,0 +1,10 @@
+# Hydrogen <img src="https://cdn.rawgit.com/nteract/hydrogen/17eda245/static/animate-logo.svg" alt="hydrogen animated logo" height="50px" align="right" />
+
+## Services API
```
aviatesk: Good to have the documentation here IMO.

lib/services/consumed/status-bar/signal-list-view.js
```diff
@@ -50,7 +50,7 @@ export default class SignalListView {
       this.cancel();
     }
 
-    const kernel = store.kernel;
+    const kernel = this.kernel;
```
aviatesk: This doesn't work if the kernel is not the one with which this `SignalListView` instance gets constructed.

lib/main.js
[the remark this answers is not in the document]
wadethestealth: Thanks for the testing. Lets add to the store

spec/services/consumed/status-bar-spec.js
wadethestealth: yes

lib/services/consumed/status-bar/status-bar.js
wadethestealth: It does work for multiple kernels, and we should be passing more arguments than just importing store. It makes for easier test cases and debugs. It also only provides the necessary data for the task. You don't need the whole store. Also I recommend you try testing this because it works with multiple kernels in my tests.

lib/services/consumed/status-bar/signal-list-view.js
wadethestealth: good catch, because of the handle kernel command. I will update that function to require a kernel arg

wadethestealth:
> 
> 
> @wadethestealth
> I changed the flow type notation of autocomplete provider function so that it uses the one in our types directory, since it has more definitions for properties and code can get more simpler.

Works for me, wasn't my code anyways

lib/services/consumed/status-bar/status-bar.js
aviatesk: I see. Then changing the kernel by `onClick` handler would be needed.

lib/services/consumed/status-bar/signal-list-view.js
aviatesk: Doesn't work:
- Launch two different kernels for two editors
- Click status bar tile inside one editor: Would show kernel 1
- Then switch the editor and click again the tile: Still shows kernel 2.

What you really should do is to update `kernel` of the `SignalListView` instance. It's not the talks about `handleKernelCommand` and thus I think you should revert the change on it.

wadethestealth: The reason it wasn't working was because I had a logic error when I ported the old code. It works now for real. Is their some other reason why you are against this change because it really is not ideal to use global variables everywhere

wadethestealth: Also when removing the logic error one option was to update the kernel, so I did that as suggested.

aviatesk: I agree with the idea of avoiding to use the global variables, of course ! 😃 
My point was just to point out the logic inside `SignalListView` seemed to be broken, and it seemed to me like you unnecessarily changed `handleKernelCommand`.

I confirmed `handleKernelCommand` is just used within `SignalListView`, thus it would just be okay as far as the logic inside `SignalListView` works though.

Personally I want to complete things with minimal code change (just my preference), thus I recommended that you revert the change. So I shouldn't have said "you _should revert_" since it' may sound kinda too strong, sorry 😅

aviatesk:
The PR seems to be good to go, and I'll merge this.

Thanks so much, @wadethestealth !

PR #1710, opened 2019-07-14
WIP: Inspectation datatip

### LINKED ISSUE #1415: Tooltip for introspection messages

aviatesk:
## :tada: New feature - Inspection datatip

Use datatip service from Atom IDE, we can show inspection result of the code on hover.

## Overview

Screenshot 2019-07-15 02 26 51

Screenshot 2019-07-15 02 30 10

## Tasks

- [x] Datatip for non-active editors
- [ ] Make inspected code range valid:
    * We need the start of inspected code range as well as the end of that
    *  I could implement the end of the range correctly referring to our `getCodeToInspect` function, but as for the _start_ still not
- May merge #1706 first, and I should refactor this PR according to that
- Configs
    * [ ] Config to turn off this feature
    * [ ] Config to set the priority of the Hydrogen datatip
- [ ] Update documentation

## :question: Questions and Discussions

- How does it work ?
    * "Atom IDE packages" (I will elaborate them below) offers the datatip service and we can consume it just like Status bar.
    * The datatip service creates the base react component and we can pass an inspection component using `@nteract` packages to it.
- Atom IDE is already dead and this can be a source of problem ?
    * `nuclide` is certainly already dead, but Atom IDE is not IMO.
    * There is a community maintained package for the datatip service at least: https://atom.io/packages/atom-ide-datatip
- What packages do offer the service ?
    * Old [Atom-IDE-UI](https://github.com/facebookarchive/atom-ide-ui):
        + Bundles the other frontend packages for Atom LSPs as well, but they are not necessary for this feature to work
        + As you may know, dead
        + [x] Works
    * [Atom-IDE-Datatip](https://github.com/atom-ide-community/atom-ide-datatip)
        + Community maintained (and hopefully alive)
        + [x] Confirmed to work as well
    * **An user should install one of the above to enjoy this feature**
        + :question: Shall we add them to our dependencies, or just recommend in our docs ?
- If an user doesn't install a package providing datatip service ?
    * Just nothing happens

## Refs

- https://github.com/atom-ide-community/atom-ide-datatip
- Closes #1415

## Misc

I also tweaked a style in our inspector pane:
- Use editor font in inspection result
- Use our custom `Markdown` component instead of `@nteract/outputs.Media.Markdown`


styles/hydrogen.less
```diff
@@ -360,8 +360,22 @@ svg:first-child {
...
-
+  code {
+    font-family: var(--editor-font-family);
+  }
   span[style] {
     color: @text-color-info !important;
   }
 }
+
+.hydrogen.datatip {
```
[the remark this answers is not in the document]
aviatesk: I want to keep them here since there are global `pre, code` style definition inside the `hydrogen` scope: https://github.com/nteract/hydrogen/blob/c046662301988657e91f3fd1a5a84cb340017e14/styles/hydrogen.less#L26

And overriding everything is boring for me.

styles/hydrogen.less
```diff
@@ -360,8 +360,22 @@ svg:first-child {
 .hydrogen.inspector {
```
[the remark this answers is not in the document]
aviatesk: Please see the above comment.

lib/datatip-provider.js
```diff
@@ -0,0 +1,82 @@
+/* @flow */
```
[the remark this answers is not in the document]
aviatesk: Yeah, I stated this on the `Tasks` section above as well.

styles/hydrogen.less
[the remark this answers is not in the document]
aviatesk: Ahh it works. Thanks !
I will refactor all of them.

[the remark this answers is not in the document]
aviatesk: @wadethestealth 
But wait, can you explain why it _"should"_ be ?

wadethestealth:
@aviatesk you should not make it a dependency, but maybe we shoud have a docs section that says "if you have datatip, autocomplete, status bar... etc. packages enabled you get x feature" Also instead of merging mine first, while I am working on my you can `rebase` off my branch

styles/hydrogen.less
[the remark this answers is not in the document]
wadethestealth: Less is a css preprocessor, and the whole reason they created the & was for this purpose. It creates cleaner stylesheets and can be more maintainable if names change around. For other benefits, read the docs. There is a bunch of neat feature less provides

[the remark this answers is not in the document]
aviatesk: Thanks ! I got it :smile:

aviatesk:
Tests would be hard since they would requires the providing package (which is not bundled within Atom by default) to be installed on every machine including the one on CI, and it's not impossible but would definitely increase CI time to be setup (and of course implementing).

Thus I want to avoid tackling that here.

aviatesk:
You can find the other IDE packages lacks the basic specs maybe because of these difficulties.

lib/store/index.js
```diff
@@ -233,16 +233,24 @@ export class Store {
...
-  getEmbeddedGrammar(editor: atom$TextEditor): ?atom$Grammar {
+  /**
+   * Returns the embedded grammar for multilanguage, normal grammar otherwise.
+   *
+   * @param {atom$Point} bufferPosition - If specified, returns the embedded grammar at the buffer
+   *                                      position, the one at the cursor position otherwise.
+   */
+  getEmbeddedGrammar(
+    editor: atom$TextEditor,
+    bufferPosition: ?atom$Point
```
wadethestealth: if the `bufferPosition` is optional, it is better to put it as `bufferPosition?: atom$Point` which means optional parameter that is an atom point vs `bufferPosition: ?atom$Point` which means parameter is required but can be null, undefined or an atom point

lib/store/index.js
```diff
@@ -288,6 +296,33 @@ export class Store {
     this.kernelMapping.set(newKey, this.kernelMapping.get(oldKey));
     this.kernelMapping.delete(oldKey);
   }
+
+  /**
+   * Returns a kernel attatched to an (multilanguage) ditor
+   */
+  getKernelForEditorBufferPosition(
```
wadethestealth: Shouldn't kernel return the correct kernel even in markdown files with multi grammars? If not why?

lib/store/index.js
aviatesk: Yeah definitely. Thanks for pointing out !

lib/store/index.js
aviatesk: As the message of the commit that introduced this function may suggest, this function is needed to support datatip for non-active editors (like the split pane on right) 🦊

aviatesk:
@wadethestealth 
> Left some comments, and I think you and I should open a separate PR for the docs to list "extra" features

Yeah, definitely. Let's do compositions 🤣

lib/services/consumed/datatip-provider.js
```diff
@@ -0,0 +1,85 @@
...
+import { RichMedia, Media } from "@nteract/outputs";
+
+import { Markdown } from "./../../components/result-view/markdown";
+import { getRow } from "./../../code-manager";
+import store from "./../../store";
+import { js_idx_to_char_idx } from "./../../utils";
+
+class DatatipProvider {
+  providerName = "Hydrogen";
+  priority = 1;
```
wadethestealth: There should probably be a config to set the priority like autocomplete has. to do it dynamically you should be able to just use atom.config, but if not, I would make this into a getter function which returns the config

lib/main.js
```diff
@@ -275,6 +275,14 @@ const Hydrogen = {
     store.subscriptions.add(disposable);
     return disposable;
   },
+
+  consumeDatatip(datatipService: atom$DatatipService) {
+    const disposable = datatipService.addProvider(
```
wadethestealth: You are consuming, but also providing. Never thought that this would happen 😆. I think you made the right choice to put it as consumed since the package.json is consuming.

lib/services/consumed/datatip-provider.js
```diff
@@ -0,0 +1,85 @@
...
+
+    const { row, column } = bufferPosition;
+    const code = getRow(editor, row);
+    if (!code) return null;
+
+    let cursorPositionStart = column;
+    let cursorPositionEnd = column;
+    const identifierEnd = code.slice(column).search(/\W/);
+    if (identifierEnd !== -1) cursorPositionEnd += identifierEnd;
+    cursorPositionStart = js_idx_to_char_idx(cursorPositionStart, code);
```
wadethestealth: good to see you remembered these cursor position functions

lib/services/consumed/datatip-provider.js
```diff
@@ -0,0 +1,85 @@
...
+        const { data, found } = result;
+        if (
+          !found ||
+          (!data["text/html"] && !data["text/markdown"] && !data["text/plain"])
+        ) {
+          reject(null);
+        } else {
+          resolve({
+            range: new Range(
+              new Point(row, cursorPositionStart),
```
wadethestealth: I believe you have to convert from a char idx to a js idx here, but I may be wrong

wadethestealth: `𝐚𝐚𝐚𝐚𝐚` I use these guys to test there js idx is 10 but char idx is 5

lib/services/consumed/datatip-provider.js
wadethestealth: Just saw your config goals, so I know this is on your todo. I am leaving it open so you know how to do it dynamically. If you need examples look at autocomplete

aviatesk:
I flag this as `WIP` since I have solutions on my mind to fix some issues around datatip but can't fix immediately since I'm kinda busy now.

Will solve in August ?

wadethestealth:
To get rid of the extra commits you need to change the branch this PR applies to and change it back to master

aviatesk:
It's okay anyways. I could rebase `master` branch successfully, although commit history seems a little weird.

wadethestealth:
That's what I am referencing. On this page change the branch you are applying the change to and then save and then change it back to master. It only does a refresh of the commit history and will rerun ci. Please do so as I use commit history often when reviewing.

aviatesk:
I don't use atom-ide-ui anymore, so let me close this PR.

PR #1711, opened 2019-07-15
Refactor #1706

aviatesk:
### Desc

Please refer to https://nteract.slack.com/archives/C4S4H8DN1/p1563163274016900

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -135,11 +135,11 @@ const Hydrogen = {
         "hydrogen:update-kernels": () => kernelManager.updateKernelSpecs(),
         "hydrogen:toggle-inspector": () => toggleInspector(store),
         "hydrogen:interrupt-kernel": () =>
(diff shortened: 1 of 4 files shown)


lib/main.js
```diff
@@ -284,14 +284,24 @@ const Hydrogen = {
...
 
+    const { kernel, grammar, markers } = store;
+
+    if (!grammar) {
+      atom.notifications.addError("Undefined grammar");
+      return;
+    }
+
     if (!kernel) {
-      const message = "No running kernel for grammar or editor found";
```
aviatesk: @wadethestealth 
I just made this change since I reverted this file from the old master state, but why did you make this change ?
I just found this here.

wadethestealth: In my opinion there is no reason to list out the grammar, and it required an extra argument

aviatesk:
Close this and wait for @wadethestealth to fix this 👍

PR #1712, opened 2019-07-15
Fix Kernel Command Bug Introduced in #1706

wadethestealth:
Discussion Was Here: https://nteract.slack.com/archives/C4S4H8DN1/p1563163274016900

### DIFF
--- a/lib/services/consumed/status-bar/status-bar.js
+++ b/lib/services/consumed/status-bar/status-bar.js
@@ -35,10 +35,10 @@ export function addStatusBar(
 function showKernelCommands(handleKernelCommand, { kernel, markers }) {
   if (!this.signalListView) {
     this.signalListView = new SignalListView(kernel);
-    this.signalListView.onConfirmed = (kernelCommand: { command: string }) =>
-      handleKernelCommand(kernelCommand, { kernel, markers });
   } else {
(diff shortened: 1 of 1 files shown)


lib/services/consumed/status-bar/status-bar.js
```diff
@@ -35,10 +35,10 @@ export function addStatusBar(
 function showKernelCommands(handleKernelCommand, { kernel, markers }) {
   if (!this.signalListView) {
     this.signalListView = new SignalListView(kernel);
-    this.signalListView.onConfirmed = (kernelCommand: { command: string }) =>
-      handleKernelCommand(kernelCommand, { kernel, markers });
   } else {
     this.signalListView.kernel = kernel;
   }
+  this.signalListView.onConfirmed = (kernelCommand: { command: string }) =>
+    handleKernelCommand(kernelCommand, { kernel, markers });
```
BenRussert: The issue (in my mind, I havent experimented) is that we want `kernel` to refer to `store.kernel` (essentially the active kernel). If we de-reference the kernel from `store` inside a function, that kernel reference never changes so it would still be the kernel passed in initially even if `store.kernel` changes.

A way to handle this is to pass `store` as an argument instead of de-referencing `{ kernel }` from store whenever this function gets initialized.

PR #1713, opened 2019-07-16
Fix #1564

### LINKED ISSUE #1564: Syntax for multiple StartUp Commands for same key
So far I understand that  the syntax for the Settings->StartUp Code is the following:

`{"display_name_of_the_kernel" : "command_that_should_be_interpretet"}`

In my case 
`display_name_of_the_kernel = Python 3`
and 
`command_that_should_be_interpretet = %run ~/output_python/start_up_script.py `

I thought I could add commands to the start_up_script.py, such as
`%matplotlib qt`. But the command got ignored. 

Also if I append it like this

`{"Python 3" : "%run ~/output_python/start_up_script.py %matplotlib qt"}`

the `%matplotlib qt ` command gets ignored. Howevery using it as a single command works:

`{"Python 3" : "%matplotlib qt"}`

What is the correct syntax of adding multiple StartUp Commands?

wadethestealth:
I am not to sure why, but for some reason using a template literal doesn't allow for 2 magic commands to be used in the start up config. I am not even sure there is a benefit to using a template literal, but using a primitive seems to fix the ending comments and issue of #1564. 
The only reason I realized in was the template literally is because of one of our log functions inside `zmq-kernel execute` was surrounding startup code with quotation, but not inline code that worked.

Authors of the commit were @rgbkrk and @lgeiger, so maybe if the peek over they could provide some insight.

- Closes #1564

### DIFF
--- a/lib/zmq-kernel.js
+++ b/lib/zmq-kernel.js
@@ -162,7 +162,7 @@ export default class ZMQKernel extends KernelTransport {
     let startupCode = Config.getJson("startupCode")[displayName];
     if (startupCode) {
       log("KernelManager: Executing startup code:", startupCode);
-      startupCode = `${startupCode} \n`;
+      startupCode += "\n";
       this.execute(startupCode, (message, channel) => {});
     }
   }


lib/zmq-kernel.js
```diff
@@ -162,7 +162,7 @@ export default class ZMQKernel extends KernelTransport {
     let startupCode = Config.getJson("startupCode")[displayName];
     if (startupCode) {
       log("KernelManager: Executing startup code:", startupCode);
-      startupCode = `${startupCode} \n`;
+      startupCode += "\n";
```
BenRussert: Or wait, do we want a newline or an escaped newline here?

wadethestealth: I believe a newline is the character we want.

PR #1714, opened 2019-07-16
Refactor Result Code

wadethestealth:
# ♻️ Clean Up Main
Move result code out of main, and pass store as an arg.

## To-Do:
- [x] add JsDoc comments

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -317,21 +317,24 @@ const Hydrogen = {
       return;
     }
 
+    const codeNullable = codeBlock.code;
+    if (codeNullable === null) return;
+
     const { row } = codeBlock;
-    let { code } = codeBlock;
     const cellType = codeManager.getMetadataForRow(editor, new Point(row, 0));
-    if (code || code === "") {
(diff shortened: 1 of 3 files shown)


lib/main.js
```diff
@@ -423,7 +327,10 @@ const Hydrogen = {
       if (moveDown === true) {
         codeManager.moveDown(editor, row);
       }
-      this.createResultBubble(editor, { code, row, cellType });
+      this.checkForKernel(store, kernel => {
+        //$FlowFixMe
+        result.createResult(store, { code, row, cellType });
```
wadethestealth: I can't fix this flow error for the life of me. It is also in 3 other places

aviatesk: Will try to fix this in next week or so.

aviatesk: Fixed in a1877f2 :)

aviatesk:
Gonna merge this if @wadethestealth gets satisfied with my changes.

wadethestealth:
Looks good @aviatesk

PR #1715, opened 2019-07-18
Import Results

### LINKED ISSUE #1502: Import Jupyter Notebook and create result cells
Creating an issue to track progress/thoughts on a possible "import notebook with results" command (it seems helpful for this to be a different issue+PR than "import notebook without results". (or a global setting for whether notebooks should be imported with results).

Result bubbles are usually created from the response of kernel execution:
https://github.com/nteract/hydrogen/blob/ea0004326a666c679119fbec582530123c4e7cca/lib/main.js#L393-L396

But all the responses from kernel execution _should_ also be stored in the Jupyter Notebook file. So it doesn't seem too crazy to think that it should be possible to import the results from the file and display them as it would from "run all cells".

### LINKED ISSUE #1552: Notebook Markdown import/export bug with consecutive newlines
Running the export on
```py
# %% markdown
# Hello world
#
# Test
# %%
print('hi')
```
gives cells of 
```json
{
  "cells": [
    {
      "cell_type": "markdown",
      "source": [
        "Hello world\n",
        "\nTest"
      ],
      "metadata": {}
    },
    {
      "cell_type": "code",
      "source": [
        "print('hi')"
      ],
      "outputs": [],
      "execution_count": null,
      "metadata": {
        "collapsed": false,
        "outputHidden": false,
        "inputHidden": false
      }
    }
  ],
  "metadata": {},
  "nbformat": 4,
  "nbformat_minor": 4
}
```
The issue is that there's a leading `\n`. So when it gets imported back into Atom, the newline is for some reason stripped and no comment symbol is included:
```py
# %% markdown
# Hello world
# 
Test
# %%
print('hi')
```

I think it's standard in Jupyter notebook files to not have leading newlines; I think they're usually appended to the end. But not sure if that's a requirement or just common.

Either way current Hydrogen behavior is bad on import or export or both.

wadethestealth:
# 🌟 Importing Results
- Now on import of a jupyter notebook, results are rendered inline 😄 
- Fixes export/ import bug #1552

## To-Do:
- [x] add JSdoc comments
   - [x] import-notebook.js
   - [x] result.js
- [x] add Docs
- [ ] add Specs
---

### Notes
~To be merged **AFTER** #1714.~

Closes #1502 
Closes #1552

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -47,6 +47,14 @@ const Config = {
       default: true,
       order: 3
     },
+    importNotebookResults: {
+      title: "Enable Import Of Notebook Results",
+      description:
+        "If enabled, anytime you import a notebook, the saved results are also rendered inline. If disabled, you can still import notebooks as normal.",
+      type: "boolean",
+      default: true,
+      order: 1.7
+    },
     statusBarDisable: {
       title: "Disable the Hydrogen status bar",
       description:
--- a/lib/config.js
+++ b/lib/config.js
@@ -48,7 +48,7 @@ const Config = {
       order: 3
     },
     importNotebookResults: {
-      title: "Enable Import Of Notebook Results",
+      title: "Enable Import of Notebook Results",
       description:
         "If enabled, anytime you import a notebook, the saved results are also rendered inline. If disabled, you can still import notebooks as normal.",
       type: "boolean",
--- a/lib/config.js
+++ b/lib/config.js
@@ -53,7 +53,7 @@ const Config = {
         "If enabled, anytime you import a notebook, the saved results are also rendered inline. If disabled, you can still import notebooks as normal.",
       type: "boolean",
       default: true,
-      order: 1.7
+      order: 4
     },
     statusBarDisable: {
       title: "Disable the Hydrogen status bar",
@@ -61,7 +61,7 @@ const Config = {
         "If enabled, no kernel information will be provided in Atom's status bar.",
       type: "boolean",
       default: false,
-      order: 4
+      order: 5
     },
     debug: {
       title: "Enable Debug Messages",
@@ -69,7 +69,7 @@ const Config = {
       description: "If enabled, log debug messages onto the dev console.",
       type: "boolean",
       default: false,
-      order: 5
+      order: 6
     },
     autoScroll: {
       title: "Enable Autoscroll",
@@ -78,7 +78,7 @@ const Config = {
         "If enabled, Hydrogen will automatically scroll to the bottom of the result view.",
       type: "boolean",
       default: true,
-      order: 6
+      order: 7
     },
     wrapOutput: {
       title: "Enable Soft Wrap for Output",
@@ -87,7 +87,7 @@ const Config = {
         "If enabled, your output code from Hydrogen will break long text and items.",
       type: "boolean",
       default: false,
-      order: 7
+      order: 8
     },
     outputAreaDefault: {
       title: "View output in the dock by default",
@@ -95,7 +95,7 @@ const Config = {
         "If enabled, output will be displayed in the dock by default rather than inline",
       type: "boolean",
       default: false,
-      order: 8
+      order: 9
     },
     outputAreaDock: {
       title: "Leave output dock open",
@@ -103,7 +103,7 @@ const Config = {
         "Do not close dock when switching to an editor without a running kernel",
       type: "boolean",
       default: false,
-      order: 9
+      order: 10
     },
     outputAreaFontSize: {
       title: "Output area fontsize",
@@ -112,7 +112,7 @@ const Config = {
       type: "integer",
       minimum: 0,
       default: 0,
-      order: 10
+      order: 11
     },
     globalMode: {
       title: "Enable Global Kernel",
@@ -120,7 +120,7 @@ const Config = {
         "If enabled, all files of the same grammar will share a single global kernel (requires Atom restart)",
       type: "boolean",
       default: false,
-      order: 11
+      order: 12
     },
     kernelNotifications: {
       title: "Enable Kernel Notifications",
@@ -129,7 +129,7 @@ const Config = {
         "Notify if kernels writes to stdout. By default, kernel notifications are only displayed in the developer console.",
       type: "boolean",
       default: false,
-      order: 12
+      order: 13
     },
     startDir: {
       title: "Directory to start kernel in",
@@ -151,7 +151,7 @@ const Config = {
         }
       ],
       default: "firstProjectDir",
-      order: 13
+      order: 14
     },
     languageMappings: {
       title: "Language Mappings",
@@ -160,7 +160,7 @@ const Config = {
         'Custom Atom grammars and some kernels use non-standard language names. That leaves Hydrogen unable to figure out what kernel to start for your code. This field should be a valid JSON mapping from a kernel language name to Atom\'s grammar name ``` { "kernel name": "grammar name" } ```. For example ``` { "scala211": "scala", "javascript": "babel es6 javascript", "python": "magicpython" } ```.',
       type: "string",
       default: '{ "python": "magicpython" }',
-      order: 14
+      order: 15
     },
     startupCode: {
       title: "Startup Code",
@@ -169,7 +169,7 @@ const Config = {
         'This code will be executed on kernel startup. Format: `{"kernel": "your code \\nmore code"}`. Example: `{"Python 2": "%matplotlib inline"}`',
       type: "string",
       default: "{}",
-      order: 15
+      order: 16
     },
     gateways: {
       title: "Kernel Gateways",
@@ -178,7 +178,7 @@ const Config = {
         'Hydrogen can connect to remote notebook servers and kernel gateways. Each gateway needs at minimum a name and a value for options.baseUrl. The options are passed directly to the `jupyter-js-services` npm package, which includes documentation for additional fields. Example value: ``` [{ "name": "Remote notebook", "options": { "baseUrl": "http://mysite.com:8888" } }] ```',
       type: "string",
       default: "[]",
-      order: 16
+      order: 17
     }
   }
 };
--- a/lib/import-notebook.js
+++ b/lib/import-notebook.js
@@ -99,18 +99,41 @@ export async function _loadNotebook(filename: string) {
   let cellBreakpoints = getBreakpoints(editor);
   let cellNumber = 1;
   for (let cell of nbCells) {
-    if (cell.cell_type == "code" && cell.outputs.toArray().length > 0) {
-      importResult(
-        { editor, markers },
-        {
-          outputs: cell.outputs.toArray(),
-          row: escapeBlankRows(
-            editor,
-            cellBreakpoints[cellNumber - 1].row,
-            cellBreakpoints[cellNumber].row - 1
-          )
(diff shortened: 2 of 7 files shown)


wadethestealth:
Ping @aviatesk @BenRussert

lib/config.js
```diff
@@ -47,6 +47,14 @@ const Config = {
       default: true,
       order: 3
     },
+    importNotebookResults: {
```
aviatesk: Does this really need to be a setting ? If this works without bugs or some drawbacks, I don't this this can be default behaviour.

lib/import-notebook.js
```diff
@@ -77,18 +113,34 @@ export async function _loadNotebook(filename: string) {
...
     return;
   }
-  const sources = _.map(nb.cells, cell => {
-    return getCellSource(cell, commentStartString + " ");
-  });
+  const nbCells = nb.cellMap.valueSeq().toArray();
+  let sources = [];
+  let resultRows = [];
+  let previousBreakpoint = -1;
+  for (let cell of nbCells) {
```
aviatesk: I think we usually use `forEach` for for loops without break/return statements inside, thus I believe it's more preferable here too.

If you stick to use native `for`, I prefer `const` over `let` for `cell` 's identifier

lib/import-notebook.js
```diff
@@ -47,25 +69,39 @@ export function importNotebook(event?: atom$CustomEvent) {
...
+    nb = fromJS(JSON.parse(data));
   } catch (err) {
     if (err.name === "SyntaxError") {
       atom.notifications.addError("Error not a valid notebook", {
         detail: err.stack
       });
     } else {
       atom.notifications.addError("Error reading file", {
-        detail: err.message
+        detail: err
```
aviatesk: What's the content of `err` ?
If it's markdown string, we may want to use `description` instead of `detail`.

lib/import-notebook.js
```diff
@@ -144,18 +208,88 @@ function getGrammarForKernelspecName(name: string): ?atom$Grammar {
...
+  resultRows: Array<number>
+) {
+  if (nbCells.length != resultRows.length) return;
+  let markers = store.markersMapping.get(editor.id);
+  markers = markers ? markers : store.newMarkerStore(editor.id);
+  let cellNumber = 0;
+  for (let cell of nbCells) {
+    const row = resultRows[cellNumber];
+
+    switch (cell.cell_type) {
```
aviatesk: We don't need branch for `cell_type == "raw"` ?

lib/config.js
wadethestealth: I am sorry but I dont understand what that means. the reason for the setting is because it is rendered inline not the output pane and this could be undesired for many reasons so I made it optional. Also there are no draw backs and there should be very little impact since the only real change is adding a result bubble to the normal import procedure. I have no idea what bug could possibly prevent this from being a default implementation.

lib/import-notebook.js
wadethestealth: I couldn't know since its a generic catch all.

lib/import-notebook.js
wadethestealth: I will look into this.

wadethestealth:
Also I will try to come up with some specs.

lib/config.js
aviatesk: Ah I'm sorry, horrible typos.
I tried to say "I think this can be default behavior" in the last sentence.

If you don't come up with any drawback, then let's make this the default and remove the config.


I feel like it would be better to keep configs clean as possible, since we have rather many.

aviatesk: If an user don't like this, s/he can just
clear results.

And more, importing a notebook won't happen so much.

lib/import-notebook.js
aviatesk: Well, we still don't support exporting/rendering raw cells, thus it would be okay to just import raw cells as is for now :)

lib/import-notebook.js
aviatesk: So, you found some cases that `err.message` is undefined or something like that ?

wadethestealth: I am pretty sure it was there before me I just removed the message portion because err should have a built-in to string that represents what the user should see,

lib/config.js
wadethestealth: Maybe you would understand better after local test because I could very easily see this as an annoyance. I know our config is a little messy, but I would rather clean that up rather than leave out a customization option.

aviatesk: Okay sounds fair.
How about making a separate command for this feature ? i.e.: `Hydrogen: Import-Notebook-With-Results` will behaves like `importNotebookResults: true` while `Hydrogen: Import-Notebook` keeps behaving the same way as before.

I personally don't like changing configs so often and think using different command is easier.

lib/import-notebook.js
```diff
@@ -77,18 +113,34 @@ export async function _loadNotebook(filename: string) {
   const editor = await atom.workspace.open();
   const grammar = getGrammarForNotebook(nb);
   if (!grammar) return;
-  editor.setGrammar(grammar);
+  atom.grammars.assignLanguageMode(editor, grammar.scopeName);
```
aviatesk: What's the reason for this change ?
It seems to fail to recognise Julia notebook as julia file in my local test.

aviatesk:
If you find writing specs difficult, it's okay to leave it to future :)

aviatesk:
I found sometimes cells are ordered incorrectly on this PR, while this doesn't not happen on master.

Maybe the logic around this is kinda broken ?:

https://github.com/nteract/hydrogen/blob/dbef962ac7403471df96e836f5bf0d85f4feeee1/lib/import-notebook.js#L122-L132

This _seems_ to happen when the cells are executed randomly (i.e. not run from top to bottom) or including plots.

***

As a further question, what might happen when the original notebook includes cells that can't be rendered in Hydrogen, like ipywidgets or something like that ?
This might relate to the raw cells stuff, but I suspect some handling of those cases might be preferable here.

lib/import-notebook.js
aviatesk: Hmm, but I found the issue only happens on this PR, and not on master thus I suspect the change is causing something bad.

lib/config.js
aviatesk: We can register those commands on `atom-workspace` and then we can call them from tree-view context menus of course ?

aviatesk: Just add the new command to import with results here ...
https://github.com/nteract/hydrogen/blob/bd951e8a471dea932f5f5c7f0c27e8936f71a106/menus/hydrogen.cson#L3

lib/import-notebook.js
aviatesk: Maybe for `getGrammarForNotebook` work, we still need the `setGrammar`.
You didn't face the issue maybe because Python grammar is handled with hard code.

***

Rather off topic, but I would really really
appreciate if you also test with various kernels other than IPython in general.

lib/config.js
wadethestealth: That won't allow you to set import with results to default though. But I agree it should still be added there.

lib/import-notebook.js
wadethestealth: You are probably right and I believe that getGrammarForNotebook may need to be updated too.

wadethestealth:
Update: I have fixed cell ordering, and it had nothing to do with execution, but rather I was using commutable wrong as cells are stored in a HashMap so there ordering is not guaranteed, which is why they also supply a List with the key order you should retrieve in.

lib/import-notebook.js
wadethestealth: Fixed

lib/config.js
wadethestealth: I firmly believe this should be a config because we cannot register two file openers when left clicking a file to open it, and therefore I am leaving this as implemented

aviatesk: How about creating two commands for importing notebooks, and register them in package menus ?
Then we can right-click on tree-view and select either way: https://github.com/nteract/hydrogen/blob/master/menus/hydrogen.cson#L1

The default behaviour (which is invoked by left-click) can be configured via config tho.

wadethestealth: @aviatesk 
Roleplay:
So lets say I set default behavior with the config to be importing with results. Then the left click tree view uses that no problem and follows directions, but then what if I want to import via command palatte.
So I go and I run `hydrogen:import-notebook` and something weird happened, I see no results and I am confused, because I have my config to import results. I then realize that I have to call a completely separate command `hydrogen:import-notebook-with-results` and I think that is weird.
End Roleplay: I think to simplify for the user, we should be consistent with which version we are using. If someone wishes to load them without results, but does not want to switch it in the config. They will just have to clear results.

wadethestealth: Also I don't think that this is blocking, so even if we disagree now, after merging your could always open a new PR and show your use case. @aviatesk

aviatesk: Fair points. Thanks for the detailed reply :)

> Also I don't think that this is blocking, so even if we disagree now, after merging your could always open a new PR and show your use case. @aviatesk

yep, definitely. But actually I don't have such a strong opinion on this, I just wanted to make a suggestion ?

So let's resolve this as is.

lib/import-notebook.js
```diff
@@ -47,89 +69,175 @@ export function importNotebook(event?: atom$CustomEvent) {
...
+    kernel_info, // Official nbformat v4
+    language, // Sometimes used in nbformat v3
+    language_info // Offical nbformat v4
+  } = nb.metadata.toJS();
+  const kernel = kernelspec ? kernelspec : kernel_info;
+  const lang = language_info
+    ? language_info
+    : language
+    ? { name: language }
+    : null;
```
aviatesk:
```js
const lang = language_info ?
  language_info :
  language ?
    { name: language } :
    null
```

would look better and be consistent with the other parts of our code

lib/import-notebook.js
```diff
@@ -47,89 +69,175 @@ export function importNotebook(event?: atom$CustomEvent) {
...
+  for (let g of grammars) {
+    if (
+      g &&
+      ((g.name && g.name.toLowerCase() == name.toLowerCase()) ||
+        g.scopeName == scopeName)
+    )
+      return g;
+  }
+  return null;
+}
```
aviatesk: Why not just use [`atom.grammars.grammarForScopeName`](https://flight-manual.atom.io/api/v1.9.4/GrammarRegistry/#instance-grammarForScopeName) ?

aviatesk:
@wadethestealth 
Do you think this PR is good to be merged as is ? If so, I'm going to merge this and let people try this out. Or you want me to do some local tests ?

wadethestealth:
@aviatesk I say go ahead and merge but I need you to test raw cells, I don't know how to use them and don't have a notebook with them. Just make sure nothing crazy happens

wadethestealth:
I will update the specs for this in a newer PR

lib/import-notebook.js
wadethestealth: Agreed that was probably an auto lint as that looks nothing like i would do.

lib/import-notebook.js
wadethestealth: this way I only have to iterate once for both name and scope checking

wadethestealth:
Going to merge this, and if any bugs come in I will do quick fix + specs and release a patch

PR #1614, opened 2019-05-09
Fix `npm run prettier` command's script

aviatesk:
The prettier command seems to be broken:
```
$ npm run prettier

> Hydrogen@2.9.0 prettier C:\Users\aviat\GitHub\Hydrogen
> prettier --write 'lib/**/*.js' 'spec/**/*.js'

[error] No matching files. Patterns tried: 'lib/**/*.js' 'spec/**/*.js' !**/node_modules/** !./node_modules/** !**/.{git,svn,hg
}/** !./.{git,svn,hg}/**
npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! Hydrogen@2.9.0 prettier: `prettier --write 'lib/**/*.js' 'spec/**/*.js'`
npm ERR! Exit status 2
npm ERR!
npm ERR! Failed at the Hydrogen@2.9.0 prettier script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\aviat\AppData\Roaming\npm-cache\_logs\2019-05-09T11_08_03_665Z-debug.log
```

I stripped single quotes and then it works fine

```
$ npm run prettier

> Hydrogen@2.9.0 prettier C:\Users\aviat\GitHub\Hydrogen
> prettier --write lib/**/*.js spec/**/*.js

lib\autocomplete-provider.js 125ms
lib\code-manager.js 189ms
lib\commands.js 36ms
lib\components\inspector.js 40ms
lib\components\kernel-monitor.js 34ms
lib\components\output-area.js 56ms
lib\components\result-view\history.js 42ms
lib\components\result-view\index.js 40ms
lib\components\result-view\list.js 21ms
lib\components\result-view\result-view.js 62ms
lib\components\result-view\status.js 22ms
lib\components\result-view\transforms.js 20ms
lib\components\status-bar.js 12ms
lib\components\watch-sidebar\index.js 18ms
lib\components\watch-sidebar\watch.js 22ms
lib\config.js 33ms
lib\existing-kernel-picker.js 29ms
lib\export-notebook.js 18ms
lib\import-notebook.js 43ms
lib\input-view.js 40ms
lib\kernel-manager.js 75ms
lib\kernel-picker.js 20ms
lib\kernel-transport.js 19ms
lib\kernel.js 64ms
lib\main.js 115ms
lib\panes\inspector.js 19ms
lib\panes\kernel-monitor.js 14ms
[...]

PR #1717, opened 2019-07-20
Add missing quotations in Dockerfile example

mhsekhavat:
Docker build raises error without those quotations.

### DIFF
--- a/docs/Usage/RemoteKernelConnection.md
+++ b/docs/Usage/RemoteKernelConnection.md
@@ -151,8 +151,8 @@ RUN chmod +x /tini
 ENV JUPYTER_TOKEN=my_secret_token # you can also pass this at runtime
 
 EXPOSE 8888
-ENTRYPOINT [/tini, --]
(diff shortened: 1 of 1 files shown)


aviatesk:
@mhsekhavat Thanks for your contribution !
Could you tell me your OS ?

I think this may cause something like [this](https://github.com/nteract/hydrogen/pull/1614).
In short, the need for quotations may depend on an user's OS.

mhsekhavat:
@aviatesk My OS is Debian Linux. 

As [docker documentation](https://docs.docker.com/engine/reference/builder/) says, exec form will be parsed as JSON:

> Note: The exec form is parsed as a JSON array, which means that you must use double-quotes (“) around words not single-quotes (‘).

wadethestealth:
@mhsekhavat thanks for the contribution

aviatesk:
@mhsekhavat 
Thanks for your description and contributions ! 😉

Issue #1702, opened 2019-07-08
Fix Hydrogen's Autocomplete Logo


wadethestealth:
<img src="https://user-images.githubusercontent.com/8658372/60779976-4cae2b00-a102-11e9-96be-e8ace0eba18c.png"/>

### :memo: Summary

A little gap appears beneath the hydrogen logo, also maybe we should change the background to match something similar to the default autocomplete suggestions.

### :checkered_flag: Motivation

This is just a user experience and design fix.

Issue #873, opened 2017-06-14
DISCUSSION: How to display docstrings?


lgeiger:
Future versions of `ipykernel` will support additional docstrings. See https://github.com/ipython/ipykernel/pull/249

I played a bit around with different ways of displaying them, let me know what you think:

## Option 1: Description and function signature
![](https://user-images.githubusercontent.com/13285808/27143968-7900a152-5130-11e7-8b38-ae54730024e9.png)

## Option 2: Only description
![](https://user-images.githubusercontent.com/13285808/27143973-7915c7d0-5130-11e7-952f-2cad038c5cd0.png)

## Option 3: Description and function signature as a snippet
![](https://user-images.githubusercontent.com/13285808/27143969-790175d2-5130-11e7-850d-4ce636d22970.png)
This will complete the whole shown string:
![](https://user-images.githubusercontent.com/13285808/27143971-79030e92-5130-11e7-9bd8-337825b236e3.png)

## Option 4: Description and function signature as right label
![](https://user-images.githubusercontent.com/13285808/27143970-79028e5e-5130-11e7-9ed1-b4e863df79ff.png)

rgbkrk:
I like (3) best.

BenRussert:
I like (3), but don't know if i'd always want the whole signature completed. Might change my mind if I try it live. 

(1) looks great too.

t9md:
I like "Option 2: Only description".
Because when I use autocomplete I normal use it for just complete the word I already know of.

So at least for me, extra information is bit noisy( I explicit refer document or src code if I want those info).

[idea] use these signature info as  dynamic snippet?  dynamically create snippet, and `tab` can expand these snippet.

Carreau:
One other possibility is to have the signature be "display:none" except when selected by the cursor.
 - 1: 😐 
 - 2: 😐 
 - 3: :+1:
  - 4: :-1:

dnowacki-usgs:
I like #1 best, then #2. 

I do not like #3; completing the whole snippet seems overkill... but maybe I'm missing the point.

#4 is ok but I don't see much improvement (or difference) compared to #1.

PR #1727, opened 2019-07-31
Enhance autocomplete

aviatesk:
- Update autocomplete's API version to 4.0.0: 35bee56
- Ease `jupyter_types_experimental` feature condition check: 400686d
     *  Allow IJulia to get type information in suggestions
     * If `start` and `end` is not in completions, just use `cursor_start`,  `cursor_end`
- Update `showAutocompleteFirst` to `autocompleteSuggestionPriority`: 1e9a8b1
     * Enables an user to specify the number of sort
     * Useful in conjunction with other IDE packages, like IDE-Python
     * **Breaking change** to users
- Show kernel inspection results in autocomplete description: ef667b0
     * Use `getSuggestionDetailsOnSelect` feature: https://github.com/atom/autocomplete-plus/wiki/Provider-API
     * Move `autocomplete` styles into global scope to enable styling
     * Make autocomplete types specific to Hydrogen's implementation

## Overview

Screenshot 2019-07-31 16 53 19

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -23,13 +23,12 @@ const Config = {
       default: true,
       order: 0
     },
-    showAutocompleteFirst: {
-      title: "Show Hydrogen's autocomplete suggestions first",
-      includeTitle: false,
+    autocompleteSuggestionPriority: {
+      title: "Autocomple Suggestion Priority",
       description:
-        "If enabled, Hydrogen's autocomplete suggestions will be listed before those of other providers, like snippets.",
-      type: "boolean",
-      default: true,
+        "Specify the sort order of Hydrogen's autocomplete suggestions.\nNote the default providers like snippets have priority of 1.",
+      type: "integer",
+      default: 3,
       order: 1
     },
     importNotebookURI: {
--- a/lib/config.js
+++ b/lib/config.js
@@ -32,9 +32,9 @@ const Config = {
       order: 1
     },
     showInspectorResultsInAutocomplete: {
-      title: "Enable Autocomplete description",
+      title: "Enable Autocomplete description (Experimental)",
       description:
-        "If enabled, Hydrogen will try to show [the results from kernel inspection](https://nteract.gitbooks.io/hydrogen/docs/Usage/GettingStarted.html#hydrogen-toggle-inspector) in each autocomplete suggestion's description. May slow down the autocompletion performance.",
+        "If enabled, Hydrogen will try to show [the results from kernel inspection](https://nteract.gitbooks.io/hydrogen/docs/Usage/GettingStarted.html#hydrogen-toggle-inspector) in each autocomplete suggestion's description. May slow down the autocompletion performance. (Note: Even if you disable this, you would still get autocomplete suggestions.)",
       type: "boolean",
       default: true,
       order: 2
--- a/lib/config.js
+++ b/lib/config.js
@@ -31,13 +31,21 @@ const Config = {
       default: 3,
       order: 1
     },
+    showInspectorResultsInAutocomplete: {
+      title: "Enable Autocomplete description",
+      description:
+        "If enabled, Hydrogen will try to show [the results from kernel inspection](https://nteract.gitbooks.io/hydrogen/docs/Usage/GettingStarted.html#hydrogen-toggle-inspector) in each autocomplete suggestion's description. May slow down the autocompletion performance.",
+      type: "boolean",
+      default: true,
+      order: 2
+    },
     importNotebookURI: {
       title: "Enable Notebook Auto-import",
       description:
         "If enabled, opening a file with extension `.ipynb` will [import the notebook](https://nteract.gitbooks.io/hydrogen/docs/Usage/NotebookFiles.html#notebook-import) file's source into a new tab. If disabled, or if the Hydrogen package is not activated, the raw file will open in Atom as normal.",
       type: "boolean",
       default: true,
-      order: 1.5
+      order: 3
     },
     statusBarDisable: {
       title: "Disable the Hydrogen status bar",
@@ -45,7 +53,7 @@ const Config = {
         "If enabled, no kernel information will be provided in Atom's status bar.",
       type: "boolean",
       default: false,
-      order: 2
+      order: 4
     },
     debug: {
       title: "Enable Debug Messages",
@@ -53,7 +61,7 @@ const Config = {
       description: "If enabled, log debug messages onto the dev console.",
       type: "boolean",
       default: false,
-      order: 3
+      order: 5
     },
     autoScroll: {
       title: "Enable Autoscroll",
@@ -62,7 +70,7 @@ const Config = {
         "If enabled, Hydrogen will automatically scroll to the bottom of the result view.",
       type: "boolean",
       default: true,
-      order: 4
+      order: 6
     },
     wrapOutput: {
       title: "Enable Soft Wrap for Output",
@@ -71,7 +79,7 @@ const Config = {
         "If enabled, your output code from Hydrogen will break long text and items.",
       type: "boolean",
       default: false,
-      order: 4.5
+      order: 7
     },
     outputAreaDefault: {
       title: "View output in the dock by default",
@@ -79,7 +87,7 @@ const Config = {
         "If enabled, output will be displayed in the dock by default rather than inline",
       type: "boolean",
       default: false,
-      order: 5
+      order: 8
     },
     outputAreaDock: {
       title: "Leave output dock open",
@@ -87,7 +95,7 @@ const Config = {
         "Do not close dock when switching to an editor without a running kernel",
       type: "boolean",
       default: false,
-      order: 6
+      order: 9
     },
     outputAreaFontSize: {
       title: "Output area fontsize",
@@ -96,7 +104,7 @@ const Config = {
       type: "integer",
       minimum: 0,
       default: 0,
-      order: 7
+      order: 10
     },
     globalMode: {
       title: "Enable Global Kernel",
@@ -104,7 +112,7 @@ const Config = {
         "If enabled, all files of the same grammar will share a single global kernel (requires Atom restart)",
       type: "boolean",
       default: false,
-      order: 8
+      order: 11
     },
     kernelNotifications: {
       title: "Enable Kernel Notifications",
@@ -113,7 +121,7 @@ const Config = {
         "Notify if kernels writes to stdout. By default, kernel notifications are only displayed in the developer console.",
       type: "boolean",
       default: false,
-      order: 9
+      order: 12
     },
     startDir: {
       title: "Directory to start kernel in",
@@ -135,7 +143,7 @@ const Config = {
         }
       ],
       default: "firstProjectDir",
-      order: 10
+      order: 13
     },
     languageMappings: {
       title: "Language Mappings",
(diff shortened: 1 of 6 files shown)


lib/config.js
```diff
@@ -23,38 +23,45 @@ const Config = {
...
+    autocompleteSuggestionPriority: {
+      title: "Autocomple Suggestion Priority",
+      description:
+        "Specify the sort order of Hydrogen's autocomplete suggestions.\nNote the default providers like snippets have priority of 1.",
+      type: "integer",
+      default: 3,
+      order: 1
+    },
+    showInspectorResultsInAutocomplete: {
+      title: "Enable Autocomplete description",
```
wadethestealth: Maybe instead `Enhance Autocomplete Description` also maybe explain that disabled will still provide autocomplete results (even though this is self explanatory, it can help newer coders/users)

lib/services/provided/autocomplete.js
```diff
@@ -12,7 +14,7 @@ type CompleteReply = {
     _jupyter_types_experimental?: Array<{
       start?: number,
       end?: number,
-      text?: string,
+      text: string,
```
wadethestealth: Are you sure that the object will always have a text object (do you have the docs?)?

styles/hydrogen.less
```diff
@@ -373,3 +366,17 @@ svg:first-child {
     color: @text-color-info !important;
   }
 }
+
+}
+
+// -------------- Overwrite Autocomplete-Plus style -----------------
+
+.autocomplete-plus {
```
wadethestealth: Is this still inside of .hydrogen? If not because it applies to the editor, can you explain why we are setting the max width is it off the page because if it is just really long but not off the page, maybe we should have it?

types/autocomplete.js.flow
```diff
@@ -1,21 +1,22 @@
 /* @flow */
 
-// One of text or snippet is required.
-// TODO(hansonw): use a union + intersection type
 type atom$AutocompleteSuggestion = {
-  text?: string,
-  snippet?: string,
+  // @NOTE: One of `text` or `snippet` is required.
```
wadethestealth: While yes only one is required, in theory, they are both optional with a condition of having the other. Therefore, you should leave this as 2 optional properties.

types/autocomplete.js.flow
```diff
@@ -1,21 +1,22 @@
 /* @flow */
 
-// One of text or snippet is required.
-// TODO(hansonw): use a union + intersection type
```
wadethestealth: This todo is suggesting a good change to the one require problem. Something like

```
type AutocompleteSuggestionText = {| text: string |} & _AutocompleteSuggestionPartial
type AutocompleteSuggestionSnippet = {| snippet: string |} & _AutocompleteSuggestionPartial
type _AutocompleteSuggestionPartial = {|
  replacementPrefix: string,
  replacedText: string,
  iconHTML?: string,
  type?: string,
  description?: string,
  descriptionMoreURL?: string,
  displayText?: string,
  leftLabel?: string,
  leftLabelHTML?: string,
  rightLabel?: string,
  rightLabelHTML?: string,
  className?: string
|}
type atom$AutocompleteSuggestion = AutocompleteSuggestionText  | AutocompleteSuggestionSnippet

File: lib/services/provided/autocomplete.js [hunk 6]
```diff
@@ -39,29 +41,41 @@ function parseCompletions(results: CompleteReply, prefix: string) {
...
-      comps[0].end != null
-    ) {
-      return _.map(comps, match => ({
-        text: match.text,
-        replacementPrefix: prefix.slice(match.start, match.end),
-        type: match.type,
-        iconHTML:
-          !match.type || match.type === "<unknown>" ? iconHTML : undefined
-      }));
+    if (comps.length > 0 && comps[0].text) {
```
wadethestealth: I don't understand why we are adding cursor logic. Can you explain a case where previous code was failing and you code is necessary?

types/autocomplete.js.flow
aviatesk: In theory, yes. But note that these type definitions are only used for Hydrogen, and we don't need stick to the strict (, which is made for general public usages) definition here.
We all know we only use `text` property for Hydrogen, since no Jupyter kernel offers `snippet` or `signature` completions, and why not use this knowledge ?

If we leave `text` as nullable field, we need additional (and no actually needed) condition checks, which seems to me very weird.

types/autocomplete.js.flow
aviatesk: Do you this is this worth to do ?

styles/hydrogen.less
aviatesk: This is not inside `.hydrogen` scope, since we can't tweak autocompletion styles otherwise.

I put `max-width` style here since completions from Jupyter kernels are kinda non-formatted (we can use whatever kernel as far as it follows jupyter messaging protocol, and there is no general docstring style for them).
It means, if we don't have this style, it looks really weird, like autocompletion expanding in a silly way.

lib/services/provided/autocomplete.js
aviatesk: This is simply because IJulia offers completions with "experimental type info" but they are not bundled with `start` or `end` property.
Like [this commit message](https://github.com/nteract/hydrogen/pull/1727/commits/400686db0821e2eb210552d048850adf792cdd3a) says.

lib/config.js
aviatesk: Yeah good point, thanks ;)

lib/services/provided/autocomplete.js
aviatesk: There is no valid and general docs for this experimental completion AFAIK.
Only IPython kernel doc says like https://ipython.readthedocs.io/en/stable/api/generated/IPython.core.completer.html#IPython.core.completer.Completion

I made `text` non nullable since I changed `autocompleteSuggesion` type definition and it makes flow complains otherwise, but at the same time I can't think completion without completion text itself.

aviatesk: And the case is still handled in the condition check in `parseCompletion` , thus I think it's enough.

styles/hydrogen.less
wadethestealth: Well I am not sure how my styles to the watches autocomplete will affect the global scope. Be aware this could cause styling issues. For now I am ok with this change however.

types/autocomplete.js.flow
wadethestealth: I already did it, and I think it is better than setting test required and not snippet.

types/autocomplete.js.flow
wadethestealth: In my opinion you should mention this in a comment then. Something like we only use text for Hydrogen but another option is using snippet. I would then be fine with this change and you can ignore the other related flow type comments

lib/services/provided/autocomplete.js
wadethestealth: making text optional only means that it can be undefined. It can never be null. However as long as there are no bugs in use I am fine with the change

lib/services/provided/autocomplete.js
wadethestealth: Oh I see what this means now thanks 👍

wadethestealth: if it is missing then can we just set cursor_start to match.start to avoid duplicate splicing logic?

aviatesk: Not sure why IPython offers `start` and `end`, but as far as they propose them I just respect them and use our `cursor_*` as fallback.

lib/services/provided/autocomplete.js
```diff
@@ -39,29 +41,41 @@ function parseCompletions(results: CompleteReply, prefix: string) {
...
+      return _.map(comps, match => {
+        const text = match.text;
```
wadethestealth: after this can we add
```
const start = match.start ? match.start : cursor_start
const end = match.end ? match.end : cursor_end
```
then update all the use cases below it makes it more readable

lib/config.js
aviatesk: Addressed.

lib/services/provided/autocomplete.js
aviatesk: Done in 0b616ed

wadethestealth: No lol. Thats not what I meant can you revert.

lib/services/provided/autocomplete.js
```diff
@@ -39,29 +41,41 @@ function parseCompletions(results: CompleteReply, prefix: string) {
...
+        const replacementPrefix =
+          match.start && match.end
+            ? prefix.slice(match.start, match.end)
+            : prefix.slice(cursor_start, cursor_end);
+        const replacedText =
+          (match.start && match.end
+            ? prefix.slice(0, match.start)
+            : prefix.slice(0, cursor_start)) + text;
```
wadethestealth: Before:
```javascript
        const replacementPrefix =
          match.start && match.end
            ? prefix.slice(match.start, match.end)
            : prefix.slice(cursor_start, cursor_end);
        const replacedText =
          (match.start && match.end
            ? prefix.slice(0, match.start)
            : prefix.slice(0, cursor_start)) + text;
```
After
```javascript
        const start = match.start ? match.start : cursor_start
        const end= match.end? match.end: cursor_end
        const replacementPrefix = prefix.slice(start, end);
        const replacedText = prefix.slice(0, start) + text;
```

lib/services/provided/autocomplete.js
aviatesk: Umm I doubt that we can have an assumption that `match.start` always comes with `match.end`, thus this kinda condition check would be preferred here.

lib/services/provided/autocomplete.js
aviatesk: Refer to the commen above.

lib/services/provided/autocomplete.js
wadethestealth:
```javascript
        const start = match.start && match.end ? match.start : cursor_start
        const end= match.start && match.end ? match.end: cursor_end
        const replacementPrefix = prefix.slice(start, end);
        const replacedText = prefix.slice(0, start) + text;
```

aviatesk: You may miss the case like below. I suspect it never happens, but if it does, it would cause some very annoying error.
```js
start = match.start
end = cursor_end
```

wadethestealth: Look above @aviatesk

aviatesk: Much neater finally :)

aviatesk:
> Well I am not sure how my styles to the watches autocomplete will affect the global scope. Be aware this could cause styling issues. For now I am ok with this change however.

I think your style for autocomplete watches were even not activated, were it ?

wadethestealth:
@aviatesk My styles were activated. There are some styling bugs if you try to use autocomplete inside watches. Those lines fixed those bugs. See #1702 and go back one edit

aviatesk:
Okay, then I think move only that part into `.hydrogen` scope again

aviatesk:
@wadethestealth 
Why did you add `char_idx_to_js_idx` to these lines, or any validation for this ?
I doesn't work for example texts below.
```julia
𝒩 = ran # After completion, it would become `𝒩 = rrand`
```

wadethestealth:
@aviatesk look at the PR for that change it should have the linked issue

wadethestealth:
Your issue is likely because of the replacement bits we made more readable

aviatesk:
No this happens on `master` as well

wadethestealth:
That's fine I am saying that the replacement bits probably need the change too. I'll look into it

aviatesk:
I already have a fix.
I'm asking just because you are sure about the changes.

wadethestealth:
Yes we need them, the issue explains it well.

wadethestealth:
@aviatesk is #873 in any way related?

aviatesk:
Really related.
But the functionality isn't implemented already, right ?

For now I just borrow the result from inspection and it's not specific to IPython though.

wadethestealth:
Idk I assume not though

PR #1729, opened 2019-08-01
Fixup Autocomplete PRs (#1658 and #1727)

aviatesk:


lib/services/provided/autocomplete.js
```diff
@@ -35,9 +35,8 @@ const regexes = {
 
 function parseCompletions(results: CompleteReply, prefix: string) {
   const { matches, metadata } = results;
-  let { cursor_start, cursor_end } = results;
-  cursor_start = char_idx_to_js_idx(cursor_start, prefix);
-  cursor_end = char_idx_to_js_idx(cursor_end, prefix);
+  const cursor_start = results.cursor_start;
+  const cursor_end = char_idx_to_js_idx(results.cursor_end, prefix);
```
aviatesk: These changes would make `replacementPrefix` valid for completions within a line that includes unicode characters.
E.g.:
```
𝒓𝒂𝒏𝒅𝒏 = rand # `replacementPrefix` should be `rand`
```

lib/services/provided/autocomplete.js
```diff
@@ -85,9 +84,9 @@ export function provideAutocompleteResults(
     inclusionPriority: 1,
 
     // The default provider has a suggestion priority of 1.
-    get suggestionPriority() {
-      return atom.config.get("Hydrogen.autocompleteSuggestionPriority");
-    },
+    suggestionPriority: atom.config.get(
+      "Hydrogen.autocompleteSuggestionPriority"
+    ),
```
aviatesk: @wadethestealth 
Please refer to the commit message 0c5c21 for why this static `suggestionPriority` is preferable.

lib/services/provided/autocomplete.js
```diff
@@ -145,19 +146,32 @@ export function provideAutocompleteResults(
...
       if (!kernel || kernel.executionState !== "idle") return null;
 
-      return new Promise(resolve => {
-        kernel.inspect(replacedText, replacedText.length, ({ data, found }) => {
-          if (!found || !data["text/plain"]) resolve(null);
-          resolve({
+      const promise = new Promise(resolve => {
+        kernel.inspect(replacedText, replacedText.length, ({ found, data }) => {
+          if (!found || !data["text/plain"]) {
+            return resolve(null);
```
aviatesk: Without `return` statement here, this callback function will continue to execute if `!found || !data["text/plain"]`, and it would cause Atom crash.

lib/services/provided/autocomplete.js
```diff
@@ -145,19 +146,32 @@ export function provideAutocompleteResults(
...
-            description: ansiToText(data["text/plain"])
+            description
           });
         });
       });
+
+      return Promise.race([promise, this.timeout()]);
+    },
+
+    timeout() {
```
aviatesk: `timeout` and `Promise.race` is introduced here just because it would make autocompletions smoother when a kernel is likely to be busy (e.g. just after starting up or when with experimental description completions)

lib/services/provided/autocomplete.js
wadethestealth: I think in this one use case it is fine either way. Plus a getter this simple will not affect performance. I don't mind the change though. I just thought it was neat to not have to restart.

lib/services/provided/autocomplete.js
wadethestealth: what do you mean replacement prefix should be rand? why is your cursor what do you have selected?

lib/config.js
```diff
@@ -26,17 +26,17 @@ const Config = {
     autocompleteSuggestionPriority: {
       title: "Autocomple Suggestion Priority",
       description:
-        "Specify the sort order of Hydrogen's autocomplete suggestions.\nNote the default providers like snippets have priority of 1.",
+        "Specify the sort order of Hydrogen's autocomplete suggestions. Note the default providers like snippets have priority of `1`. Requires Atom restart to take an effect.",
```
wadethestealth: Requiring an atom restart is not true. It **only requires** autocomplete-plus restart **or** hydrogen restart. Restarting atom does both of those, but is not required.

lib/services/provided/autocomplete.js
aviatesk: The cursor is at the end of `rand` and try to complete to `rand` or `randn` or something.
On the `master`, the `replacementPrefix` would something like just `r`. Just give a try.

lib/config.js
aviatesk: Is it possible to _restart_ a package ?
I suspect the cached loaded code or something like that remains, non ?

lib/services/provided/autocomplete.js
wadethestealth: what kernel? because i get no replacement

wadethestealth: i get no suggestions at all for rand

lib/services/provided/autocomplete.js
aviatesk: Yes, the computation may be negligible, but I don't like computing `atom.config.get` every time we make completions. It maybe a talk of coding styles.

You can see other great plugins just follow the rules:
https://github.com/TypeStrong/atom-typescript/blob/master/lib/main/atom/autoCompleteProvider.ts

And the API documentation:
https://github.com/atom/autocomplete-plus/wiki/Provider-API

lib/services/provided/autocomplete.js
wadethestealth: i used `𝒓𝒂𝒏𝒅𝒏 = fil` and it correctly replaces for `file` and for `%%file`

aviatesk: Ah, IJulia.
IPython would bring `match.start`  and `match.end`, thus it may not happen.

aviatesk: Maybe the same issue would happen for any kernel other than IPython.

lib/services/provided/autocomplete.js
aviatesk: And once an user set up their priority number, there wouldn't be any need for changing it dynamically.
Dynamic change would be happy only while s/he tweaking around it.

That's why I think the dynamic computation here would be really no needed.

lib/config.js
wadethestealth: enable and disable of a package trigger the services api. In other words our provider is recreated every time one of the packages is enable. You can easily test this by going to package settings and disabling and re-enabling

wadethestealth: also on disable, the service api cleans up our provider

aviatesk: Okay then it would be more precise.
But do you think the concept of "restarting autocomplete-plus" or "restarting Hydrogen" is clear to users ? I mean, to say like "require to disable and enable the package" can sound like weird.
If you so, I'm okay to change the description.

lib/services/provided/autocomplete.js
wadethestealth: Since I don't have another kernel I can't test this but can you do these test cases and verify.
The `|` denotes cursor, anything after this should not be included as a prefix. everything before until the space should.
I know there isn't a great way of testing, but do you believe these would pass.
`𝒓𝒂𝒏𝒅𝒏 = 𝒓𝒂nd|`
`𝒓𝒂𝒏𝒅𝒏 = 𝒓𝒂n|d`
`𝒓𝒂𝒏𝒅𝒏 = 𝒓𝒂|nd`
`𝒓𝒂𝒏𝒅𝒏 = 𝒓|𝒂nd`
Also are you sure that parsing cursor start is causing the problem. I mean is it literally return a different number like 1 instead of 0 using your test. Can you screenshot the scope variables in debug?

lib/config.js
wadethestealth: I know there is no restart button, but restart is equivalent to off and back on again and disable is equivalent to off and enable is equivalent to on. I think native english speakers would understand restart Hydrogen or Autocomplete-Plus, but I am unsure of language barriers

lib/services/provided/autocomplete.js
aviatesk: I think no completion happens for `𝒓𝒂nd` by default in any language.
You mean bind `𝒓𝒂nd` to some function and try to complete ?

lib/config.js
aviatesk: Hmm I don't feel something weird in "restart" word itself here, but rather feel like the concept of "restart" in Atom package can be ambiguous.

lib/services/provided/autocomplete.js
wadethestealth: https://jupyter-client.readthedocs.io/en/latest/messaging.html#completion
Read this carefully. Note the change in 5.2.

wadethestealth: We should remove the char to js parsing according to the docs. The must be a different logic error

aviatesk: The problem may come from that no all the Jupyter kernels follow the 5.2 protocol.

lib/config.js
wadethestealth: Seems like reload is better

lib/services/provided/autocomplete.js
wadethestealth: Than we either need a way to detect the protocol on kernel startup and save it in the kernel object. Or just support the latest protocol.

aviatesk: Detecting protocol stuff doesn't sound good way, since it would make many branching and code will lead to be more complex.
To only support latest protocol isn't definitely sane approach. We lose support for many kernels, and it's destroying Hydrogen's beauty.

Rather I would like to take a regex-based approach, which is independent of `cursor_` feedback from kernels.

For now, I will restore `js_ind` stuff.

lib/config.js
aviatesk: I just would like to hear you suggestion of the alternative to

> Requires Atom restart to take an effect.

lib/services/provided/autocomplete.js
wadethestealth: I didn't think about regex. That could be interesting.

lib/config.js
wadethestealth: "Requires a Hydrogen restart to take effect"

wadethestealth: "Requires a Hydrogen restart to take effect"

aviatesk: Let's go just with "restart" allowing general interpretations for users :)

lib/services/provided/autocomplete.js
aviatesk: Leaving `TODO` and revert back to the previous implementation for now.
The real fix will be by another PR.

aviatesk:
@wadethestealth 
Sorry I just made a mistake and pushed this branch into master ...
Basically this PR was already ready for merging tho.

Ask me if you have something to be fixed around this PR.

BenRussert:
> Sorry I just made a mistake and pushed this branch into master

I just turned on 1 approval required for pushing/merging to master to help prevent that 😄

wadethestealth:
@aviatesk still says atom restart which isn't true.

aviatesk:
@BenRussert

> I just turned on 1 approval required for pushing/merging to master to help prevent that 😄

Cool ! Thanks !

PR #1763, opened 2019-09-03
tweak autocompletion style

aviatesk:
- #1727 can show silly large description, and this PR suppress that more

### DIFF
--- a/styles/hydrogen.less
+++ b/styles/hydrogen.less
@@ -395,5 +395,6 @@ svg:first-child {
 .autocomplete-plus {
   autocomplete-suggestion-list.select-list.popover-list .suggestion-description {
     max-width: 500px;
+    max-height: 300px;
   }
 }


wadethestealth:
Does this add scroll bar or use an elipse? Can I get a screenshot example?

aviatesk:
> Does this add scroll bar or use an elipse?
No, would be better if we could tho

wadethestealth:
Can I see what it looks like and I think both options are fine.

aviatesk:
Ah sorry, I found that the other package (atom-ink) had a change like:
```
autocomplete-suggestion-list.select-list.popover-list {
    .suggestion-description {
      line-height: 1;
      .suggestion-description-content {
        overflow-y : hidden;
        display    : inline;
        white-space: normal; // can be useful to workaround https://github.com/atom/autocomplete-plus/issues/868
      }
    }
  }
```

and actually `display    : inline;` made something like:
image

But it's now fixed with: https://github.com/JunoLab/atom-ink/pull/217
and there is no more annoyance, so gonna close this.

Sorry for taking your time !

Issue #1780, opened 2019-09-20
Uncaught Error: ENOENT: no such file or directory, unlink '~/Library/Jupyter/runtime/kernel-c0601...


lkondapi:
[Enter steps to reproduce:]

1. ...
2. ...

**Atom**: 1.40.1 x64
**Electron**: 3.1.10
**OS**: Mac OS X 10.14.6
**Thrown From**: [Hydrogen](https://github.com/nteract/hydrogen) package 2.12.0

### Stack Trace

Uncaught Error: ENOENT: no such file or directory, unlink '/Users/lkondapi/Library/Jupyter/runtime/kernel-c0601500-e3f7-46e6-9814-568132358ad7.json'

```
At fs.js:143

Error: ENOENT: no such file or directory, unlink '/Users/lkondapi/Library/Jupyter/runtime/kernel-c0601500-e3f7-46e6-9814-568132358ad7.json'
    at Object.fs.unlinkSync (fs.js:1061:3)
    at ZMQKernel.destroy (/packages/Hydrogen/lib/zmq-kernel.js:393:8)
    at Kernel.destroy (/packages/Hydrogen/lib/kernel.js:435:20)
    at /packages/Hydrogen/lib/main.js:300:14)
    at Object.handleKernelCommand (/packages/Hydrogen/lib/main.js:310:4)
    at SignalListView.signalListView.onConfirmed (/packages/Hydrogen/lib/services/consumed/status-bar/status-bar.js:59:14)
    at Object.didConfirmSelection (/packages/Hydrogen/lib/services/consumed/status-bar/signal-list-view.js:40:36)
    at SelectListView.confirmSelection (/packages/Hydrogen/node_modules/atom-select-list/src/select-list-view.js:400:20)
    at SelectListView.didClickItem (/packages/Hydrogen/node_modules/atom-select-list/src/select-list-view.js:286:10)
    at ListItemView.onclick (/packages/Hydrogen/node_modules/atom-select-list/src/select-list-view.js:221:31)
    at ListItemView.didClick (/packages/Hydrogen/node_modules/atom-select-list/src/select-list-view.js:448:10)
```

### Commands

```
     -0:52.5.0 hydrogen:run-and-move-down (input.hidden-input)
     -0:52.5.0 autocomplete-plus:cancel (atom-text-editor.editor.is-focused)
     -0:51.7.0 hydrogen:run-and-move-down (input.hidden-input)
[...]

Issue #1778, opened 2019-09-20
Uncaught Error: ENOENT: no such file or directory, unlink '~/Library/Jupyter/runtime/kernel-2c70a...


77QingLiu:
[Enter steps to reproduce:]

1. ...
2. ...

**Atom**: 1.40.1 x64
**Electron**: 3.1.10
**OS**: Mac OS X 10.14.5
**Thrown From**: [Hydrogen](https://github.com/nteract/hydrogen) package 2.12.0

### Stack Trace

Uncaught Error: ENOENT: no such file or directory, unlink '/Users/liuqing/Library/Jupyter/runtime/kernel-2c70ad34-a293-4fbb-9d55-098c51049bb5.json'

```
At fs.js:143

Error: ENOENT: no such file or directory, unlink '/Users/liuqing/Library/Jupyter/runtime/kernel-2c70ad34-a293-4fbb-9d55-098c51049bb5.json'
    at Object.fs.unlinkSync (fs.js:1061:3)
    at ZMQKernel.destroy (/packages/Hydrogen/lib/zmq-kernel.js:393:8)
    at Kernel.destroy (/packages/Hydrogen/lib/kernel.js:435:20)
    at /packages/Hydrogen/lib/main.js:300:14)
    at Object.handleKernelCommand (/packages/Hydrogen/lib/main.js:310:4)
    at SignalListView.signalListView.onConfirmed (/packages/Hydrogen/lib/services/consumed/status-bar/status-bar.js:59:14)
    at Object.didConfirmSelection (/packages/Hydrogen/lib/services/consumed/status-bar/signal-list-view.js:40:36)
    at SelectListView.confirmSelection (/packages/git-plus/node_modules/atom-select-list/src/select-list-view.js:400:20)
    at SelectListView.didClickItem (/packages/git-plus/node_modules/atom-select-list/src/select-list-view.js:286:10)
    at ListItemView.onclick (/packages/git-plus/node_modules/atom-select-list/src/select-list-view.js:221:31)
    at ListItemView.didClick (/packages/git-plus/node_modules/atom-select-list/src/select-list-view.js:448:10)
```

### Commands

```
     -9:22.4.0 hydrogen:run-and-move-down (input.hidden-input)
     -9:22.4.0 autocomplete-plus:cancel (atom-text-editor.editor.is-focused)
     -9:20.3.0 hydrogen:clear-results (input.hidden-input)
     -8:56.1.0 hydrogen:run (input.hidden-input)
[...]

Issue #1770, opened 2019-09-12
Unable to close / restart Kernels without specific reason


thomasbazeille:
Hello, 

Since the last update, I ran several times in a new error : the command to manage the running kernel (local or remote) in atom command bar doesn't manage to trigger any action from the kernel anymore (either restarting / shutting down/ disconneting).

I don't know to what it could be linked but I never saw this one before and now run systematically into it.

Uncaught Error: ENOENT: no such file or directory, unlink '~/Library/Jupyter/runtime/kernel-88282...

[Enter steps to reproduce:]

1. ...
2. ...

**Atom**: 1.40.0 x64
**Electron**: 3.1.10
**OS**: Mac OS X 10.13.1
**Thrown From**: [Hydrogen](https://github.com/nteract/hydrogen) package 2.12.0

### Stack Trace

Uncaught Error: ENOENT: no such file or directory, unlink '/Users/thomasbazeille/Library/Jupyter/runtime/kernel-88282f81-d3df-40a6-bd15-ed71dc610cf0.json'

```
At fs.js:143

Error: ENOENT: no such file or directory, unlink '/Users/thomasbazeille/Library/Jupyter/runtime/kernel-88282f81-d3df-40a6-bd15-ed71dc610cf0.json'
    at Object.fs.unlinkSync (fs.js:1061:3)
    at ZMQKernel.destroy (/packages/Hydrogen/lib/zmq-kernel.js:393:8)
    at Kernel.destroy (/packages/Hydrogen/lib/kernel.js:435:20)
    at /packages/Hydrogen/lib/main.js:300:14)
    at Object.handleKernelCommand (/packages/Hydrogen/lib/main.js:310:4)
    at SignalListView.signalListView.onConfirmed (/packages/Hydrogen/lib/services/consumed/status-bar/status-bar.js:59:14)
    at Object.didConfirmSelection (/packages/Hydrogen/lib/services/consumed/status-bar/signal-list-view.js:40:36)
    at SelectListView.confirmSelection (/packages/git-plus/node_modules/atom-select-list/src/select-list-view.js:400:20)
    at SelectListView.didClickItem (/packages/git-plus/node_modules/atom-select-list/src/select-list-view.js:286:10)
    at ListItemView.onclick (/packages/git-plus/node_modules/atom-select-list/src/select-list-view.js:221:31)
    at ListItemView.didClick (/packages/git-plus/node_modules/atom-select-list/src/select-list-view.js:448:10)
```

### Commands

```
[...]

Issue #1767, opened 2019-09-07
Uncaught Error: ENOENT: no such file or directory, unlink '~/Library/Jupyter/runtime/kernel-07ace...


alanpang1990301:
[Enter steps to reproduce:]

Click interrupt kernel.

**Atom**: 1.40.1 x64
**Electron**: 3.1.10
**OS**: Mac OS X 10.14.6
**Thrown From**: [Hydrogen](https://github.com/nteract/hydrogen) package 2.12.0

### Stack Trace

Uncaught Error: ENOENT: no such file or directory, unlink '/Users/alanpang/Library/Jupyter/runtime/kernel-07ace1fb-742c-4b45-bb2b-757b0ad39bd7.json'

```
At fs.js:143

Error: ENOENT: no such file or directory, unlink '/Users/alanpang/Library/Jupyter/runtime/kernel-07ace1fb-742c-4b45-bb2b-757b0ad39bd7.json'
    at Object.fs.unlinkSync (fs.js:1061:3)
    at ZMQKernel.destroy (/packages/Hydrogen/lib/zmq-kernel.js:393:8)
    at Kernel.destroy (/packages/Hydrogen/lib/kernel.js:435:20)
    at /packages/Hydrogen/lib/main.js:300:14)
    at Object.handleKernelCommand (/packages/Hydrogen/lib/main.js:310:4)
    at SignalListView.signalListView.onConfirmed (/packages/Hydrogen/lib/services/consumed/status-bar/status-bar.js:59:14)
    at Object.didConfirmSelection (/packages/Hydrogen/lib/services/consumed/status-bar/signal-list-view.js:40:36)
    at SelectListView.confirmSelection (/packages/Hydrogen/node_modules/atom-select-list/src/select-list-view.js:400:20)
    at SelectListView.didClickItem (/packages/Hydrogen/node_modules/atom-select-list/src/select-list-view.js:286:10)
    at ListItemView.onclick (/packages/Hydrogen/node_modules/atom-select-list/src/select-list-view.js:221:31)
    at ListItemView.didClick (/packages/Hydrogen/node_modules/atom-select-list/src/select-list-view.js:448:10)
```

### Commands

```
     -0:21.1.0 hydrogen:run-and-move-down (input.hidden-input)
     -0:21.1.0 autocomplete-plus:cancel (atom-text-editor.editor.is-focused)
     -0:21 hydrogen:run-and-move-down (input.hidden-input)
[...]

PR #1781, opened 2019-09-22
Update status-bar.js

jh-wu:
fix "Not able to Shut Down kernel twice (Uncaught Error: ENOENT) #1760"
duplicated issues: #1756, #1780, #1778, #1770, #1767, #1764, #1765

Tested scenarios:
1. from #1760 

> 1. open atom
> 2. Initialize kernel (python 3)
> 3. shut down lernel
> 4. Initialize kernel again
> 5. shut down lernel

Pass

2. from #1756

> 1. Open two atom sessions using "atom ." in the command line
> 2. initialize python 3 in one of them
> 3. Change working directory using
> 4. import os
> 5. os.chdir(script_dir)
> 6. Shutdown python kernel

Pass

3. from #1770
test interrupt and restart kernel.

Pass

### DIFF
--- a/lib/services/consumed/status-bar/signal-list-view.js
+++ b/lib/services/consumed/status-bar/signal-list-view.js
@@ -4,6 +4,7 @@ import SelectListView from "atom-select-list";
 
 import WSKernel from "../../../ws-kernel";
 import { log } from "../../../utils";
+import type { Store } from "../../../store";
 
 const basicCommands = [
   { name: "Interrupt", value: "interrupt-kernel" },
@@ -21,7 +22,7 @@ export default class SignalListView {
   previouslyFocusedElement: ?HTMLElement;
   selectListView: SelectListView;
   store: ?Store;
-  handleKernelCommand: Function;
+  handleKernelCommand: ?Function;
 
   constructor(store: Store, handleKernelCommand: Function) {
     this.store = store;
--- a/lib/services/consumed/status-bar/signal-list-view.js
+++ b/lib/services/consumed/status-bar/signal-list-view.js
@@ -38,7 +38,7 @@ export default class SignalListView {
       },
       didConfirmSelection: item => {
         log("Selected command:", item);
-        if (this.onConfirmed) this.onConfirmed(item);
+        this.onConfirmed(item);
         this.cancel();
       },
       didCancelSelection: () => this.cancel(),
@@ -47,14 +47,16 @@ export default class SignalListView {
   }
 
   onConfirmed(kernelCommand: { command: string }){
-    	this.handleKernelCommand(kernelCommand, this.store);
+    	if (this.handleKernelCommand) {
+        this.handleKernelCommand(kernelCommand, this.store);
+      }
   }
 
   async toggle() {
     if (this.panel != null) {
       this.cancel();
     }
-
+    if (!this.store) return;
     const kernel = this.store.kernel;
     if (!kernel) return;
     const commands =
--- a/lib/services/consumed/status-bar/signal-list-view.js
+++ b/lib/services/consumed/status-bar/signal-list-view.js
@@ -17,15 +17,15 @@ const wsKernelCommands = [
 ];
 
 export default class SignalListView {
-  onConfirmed: ?(command: { command: string }) => void;
   panel: ?atom$Panel;
   previouslyFocusedElement: ?HTMLElement;
   selectListView: SelectListView;
-  kernel: ?Kernel;
+  store: ?Store;
+  handleKernelCommand: Function;
 
-  constructor(kernel: Kernel) {
-    this.kernel = kernel;
-    this.onConfirmed = null;
+  constructor(store: Store, handleKernelCommand: Function) {
+    this.store = store;
+    this.handleKernelCommand = handleKernelCommand;
     this.selectListView = new SelectListView({
       itemsClassList: ["mark-active"],
       items: [],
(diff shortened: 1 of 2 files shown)


lib/services/consumed/status-bar/status-bar.js
```diff
@@ -45,19 +44,13 @@ export class StatusBarConsumer {
...
+  showKernelCommands(store: Store, handleKernelCommand: Function) {
     let signalListView = this.signalListView;
     if (!signalListView) {
-      signalListView = new SignalListView(kernel);
-      signalListView.onConfirmed = (kernelCommand: { command: string }) =>
-        this.handleKernelCommand(kernelCommand, { kernel, markers });
+      signalListView = new SignalListView(store, handleKernelCommand);
       this.signalListView = signalListView;
+    }else{
+      signalListView.store = store
```
wadethestealth: please run a linter on your code, the else should have white-space around it and line 53 should have a semi-colon

lib/services/consumed/status-bar/signal-list-view.js
```diff
@@ -17,15 +17,15 @@ const wsKernelCommands = [
 ];
 
 export default class SignalListView {
-  onConfirmed: ?(command: { command: string }) => void;
   panel: ?atom$Panel;
   previouslyFocusedElement: ?HTMLElement;
   selectListView: SelectListView;
-  kernel: ?Kernel;
+  store: ?Store;
```
wadethestealth: flow does not understand what this means please import type Store

lib/services/consumed/status-bar/signal-list-view.js
```diff
@@ -17,15 +17,15 @@ const wsKernelCommands = [
...
+  handleKernelCommand: Function;
 
-  constructor(kernel: Kernel) {
-    this.kernel = kernel;
-    this.onConfirmed = null;
+  constructor(store: Store, handleKernelCommand: Function) {
+    this.store = store;
+    this.handleKernelCommand = handleKernelCommand;
```
wadethestealth: you must declare this field outside the constructor and give it a flow type before you can use it.

lib/services/consumed/status-bar/status-bar.js
jh-wu: Hi @wadethestealth, thanks for review. I am not familiar with linter for javascript. I have tried installing linter-spell-javascript, linter-flow and language-babel in Atom. But none of them seems working and giving me any hint about code style. I have made changes according your comments though. Can you recommend a plugin for this please?

lib/services/consumed/status-bar/signal-list-view.js
```diff
@@ -45,12 +46,16 @@ export default class SignalListView {
...
+    	this.handleKernelCommand(kernelCommand, this.store);
+  }
+
   async toggle() {
     if (this.panel != null) {
       this.cancel();
     }
 
-    const kernel = this.kernel;
+    const kernel = this.store.kernel;
```
wadethestealth: You must check if store is defined before using it.

lib/services/consumed/status-bar/signal-list-view.js
```diff
@@ -45,12 +46,16 @@ export default class SignalListView {
     });
   }
 
+  onConfirmed(kernelCommand: { command: string }){
+    	this.handleKernelCommand(kernelCommand, this.store);
```
wadethestealth: you must check if `handleKernelCommand` is defined first before using it

lib/services/consumed/status-bar/status-bar.js
```diff
@@ -31,8 +30,8 @@ export class StatusBarConsumer {
 
     this.handleKernelCommand = handleKernelCommand;
```
wadethestealth: By removing line 17, `handleKernelCommand: Function;`, `handleKernelCommand` is now not defined as a class member before you reference it. If you do not need to save the function, remove this line and refactor the code elsewhere accordingly, or add line 17 back.

jh-wu: hi @wadethestealth thank you very much for your thorough instruction and patience. I don't have npm in my computer yet. Let's see, if there are more issues, I will install npm and do flow typechecking.

jh-wu:
Done.

lib/services/consumed/status-bar/status-bar.js
```diff
@@ -45,19 +42,13 @@ export class StatusBarConsumer {
...
+  showKernelCommands(store: Store, handleKernelCommand: Function) {
     let signalListView = this.signalListView;
     if (!signalListView) {
-      signalListView = new SignalListView(kernel);
-      signalListView.onConfirmed = (kernelCommand: { command: string }) =>
-        this.handleKernelCommand(kernelCommand, { kernel, markers });
+      signalListView = new SignalListView(store, handleKernelCommand);
       this.signalListView = signalListView;
+    } else {
+      signalListView.store = store;
```
wadethestealth: curious as to whether this line is truly necessary as store never changes because it is global among hydrogen. If it is or isn't necessary why not update the handle kernel command too?

jh-wu: uh-huh, I know you will question this line. Actually I didn't expect it when I was changing kernel to store. I though this line should be useless as store.kernel should always return the active kernel. However, this line has to be here in reality, otherwise the store.kernel is aways the initial one, i.e. the bug remains. My guess of the reason of this is that the onClick function actually gets a copy of Store instead of a reference. But I can't confirm it as I am not familiar with react.
As for the handleKernelCommand, it's not necessary to update it. It's stateless, not interior information carried.
I guess you might have to do a big refactor on the Store for improvement, to separate information and actions, maybe create a class call context or something similar IMHO.

PR #1789, opened 2019-10-06
Remove unneccessery else with return

polemius:
```
if (a) {
   return 'object a';
} else {
   return 'not object a';
}
```

It is the same as:
```
if (a) {
   return 'object a';
}

return 'not object a';
```

But more readable code.


wadethestealth:
Thanks for your contribution! Feel free to contribute more. If you are interested in being a maintainer, you should join our slack, let us know, and keep contributing.

PR #1790, opened 2019-10-06
Center screen on cursor on run[-cell]-and-move-down

divieira:
When using _run-cell-and-move-down_ (or _run-and-move-down_), the cursor moves to the next cell (or line). But if it is off-screen or too further down, only a few lines of code of the next cell (or lines to come) are visible.

In Jupyter Notebook, the "run cell and select next" command centers the screen on the next cell.

This change modifies the behavior of the _run-cell-and-move-down_ and _run-and-move-down_ to match that of Jupyter Notebook and helps to make the next code to run more visible.

### DIFF
--- a/lib/code-manager.js
+++ b/lib/code-manager.js
@@ -338,6 +338,15 @@ export function getCellsForBreakPoints(
   );
 }
 
+function centerScreenOnCursorPosition(editor: atom$TextEditor) {
+  const cursorPosition = editor.element.pixelPositionForScreenPosition(
+    editor.getCursorScreenPosition()
+  ).top;
+  const editorHeight = editor.element.getHeight();
+
+  editor.element.setScrollTop(cursorPosition - editorHeight / 2);
+}
+
 export function moveDown(editor: atom$TextEditor, row: number) {
   const lastRow = editor.getLastBufferRow();
 
@@ -356,6 +365,8 @@ export function moveDown(editor: atom$TextEditor, row: number) {
     row,
     column: 0
   });
+
+  centerScreenOnCursorPosition(editor);
 }
 
 export function findPrecedingBlock(


wadethestealth:
I love this idea, but instead of using DOM code, please use the atom API.

divieira:
Ah, I should have made a comment on that. There seems to be no API for this. My first approach was to try the more obvious editor.scrollToCursorPosition (and [variants](https://flight-manual.atom.io/api/v1.0.19/TextEditor/#instance-scrollToCursorPosition)), but it has no effect if the cursor is already visible. This [open issue](https://github.com/atom/atom/issues/11429) confirms that and that there is no alternative.

So I took the same approach as the vim-mode-plus package ([link to relevant code](https://github.com/t9md/atom-vim-mode-plus/blob/master/lib/scroll-manager.js#L22)) and accessed the DOM directly. I'm following the issue and will be happy to update if there is a change in the API.

Glad you liked the change proposal! I have actually just started to use Atom because of Hydrogen, and so far am loving it!

BenRussert:
@wadethestealth I agree with both your initial feedback and @divieira's reply. Of course, we always prefer the atom api, but I think this change is isolated enough that it will be safe.

Either way, I'll let @wadethestealth reply/merge in case there is anything I haven't thought of.

wadethestealth:
I am going to test this the next couple of days, if I feel its not choppy or doesn't feel weird I am going to merge. Other hydrogen users feel free to put in your input about the feature.

aviatesk:
I tried with this PR and found it gives a better behaviour while the implementation is kept to be so simple enough.
I would like to merge this and include in the coming release, and see how uses feel on this.

Because implementation is so simple that we can do future refactor/improvements easily.

kylebarron:
I think there needs to be an option to turn this off. I just updated to 2.14.0, and using `run-and-move-down` feels extremely jerky. Any time I run a line of code at the top or bottom of the viewport, the screen now centers on that line. I would argue that the centering behavior is too unexpected to the user. 

I think instead of always keeping the line/cell centered, a better solution would be: if the next line is off the screen, then scroll the window down to that line. That would be way less jerky.

PR #1842, opened 2019-12-13
Use `console.log` instead of `console.debug` and honour Hydrogen setting

n-riesco:
Recent versions of Atom filter out `console.debug` messages from the dev console.

This PR ensures all log messages use `console.log`, by:
* replacing `console.debug` with `console.log` in `utils.js#log`
* replacing the use of `console.info`, `console.warn` and `console.log` with `log`, so that Hydrogen honours the setting to disable debug messages.

### DIFF
--- a/lib/kernel.js
+++ b/lib/kernel.js
@@ -363,7 +363,7 @@ export default class Kernel {
             stream: "status"
           });
         } else {
-          console.warn(
+          log(
             "Kernel: ignoring unexpected value for message.content.status"
           );
         }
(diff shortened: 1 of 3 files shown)


wadethestealth:
Not too sure where you see them filtering out dev logs (especially since logs are only sent when in `atom --dev` mode). Do you have a source? However, it really doesn't change that much and I would rather use something I know will be there.

n-riesco:
This is what I see in Atom v1.41 (verbose messages aren't enabled):

Screenshot from 2019-12-15 19-18-03

wadethestealth:
@n-riesco I don't think that merits us using `console.log` for everything especially since if you enable verbose messages it shows the logs and will remember that for next launch. The reason verbose is not enabled is because its not default and shouldn't be. However, since we don't have a standard for what is verbose and what is not, `log` is probably fine for now.

PR #1846, opened 2019-12-17
Replace jmp with enchannel-zmq-backend

wadethestealth:
This updates the `zmq-kernel` in Hydrogen to use the nteract library for handling jupyter zmq sockets instead of the jmp library for sockets.

Please review this code thoroughly because I am new to using these libraries (RxJs, jmp, and enchannel-zmq-backend) and I will likely have made a mistake in clean up or misunderstood a fundamental requirement.


lib/zmq-kernel.js
```diff
@@ -23,16 +23,27 @@ export type Connection = {
   version: number
 };
 
+const nullChannel = {
+  next: ()=>{},
+  complete: ()=>{},
+  subscribe: ()=>{},
+};
```
rgbkrk: I assume this is just a filler until set up at the start?

rgbkrk: Ah, looks like it gets used when the channel is closed down too.

wadethestealth:
@aviatesk Can you test this and publish as a patch if it works? This change is only for local kernels. It works on my machine.

aviatesk:
@wadethestealth Sure.
Couple of questions:
- what the main difference between jmp and enchannel-zmq-backend ? why do we need to migrate into the latter ?
- do you want me to look through the changes, or is it okay to just make a manual test ?

aviatesk:
So far I found an issue on starting [common-lisp jupyter kernel](https://github.com/yitzchak/common-lisp-jupyter) on Windows.

Here is the kernel runtime information, but there seems no errro symptoms in logs...
> jupyter/runtime/kernel_**.json
```
{
  "version":5,
  "key":"43027e2c-b236-42d5-9a33-ceffce261ccc",
  "signature_scheme":"hmac-sha256",
  "transport":"tcp",
  "ip":"127.0.0.1",
  "hb_port":9001,
  "control_port":9005,
  "shell_port":9006,
  "stdin_port":9007,
  "iopub_port":9008
}
```

> common lisp jupyter log
```
2020-02-14 22:33:40 [INFO] <KERNEL> Starting common-lisp kernel
2020-02-14 22:33:40 [INFO] <KERNEL> Parsing connection file C:\\Users\\aviat\\AppData\\Roaming\\jupyter\\runtime\\kernel-76723a10-6ce1-49da-9236-43f08da7ce86.json
2020-02-14 22:33:40 [INFO] <MAC> Starting message authentification
2020-02-14 22:33:40 [INFO] <MAC> Using hmac with sha256 digest for message signing.
2020-02-14 22:33:40 [INFO] <HB-CHANNEL> Starting channel on tcp://127.0.0.1:9001
2020-02-14 22:33:40 [INFO] <IOPUB-CHANNEL> Starting channel on tcp://127.0.0.1:9008
2020-02-14 22:33:40 [INFO] <HB-CHANNEL> Starting thread
2020-02-14 22:33:40 [INFO] <SHELL-CHANNEL> Starting channel on tcp://127.0.0.1:9006
2020-02-14 22:33:40 [INFO] <STDIN-CHANNEL> Starting channel on tcp://127.0.0.1:9007
2020-02-14 22:33:40 [INFO] <HISTORY> Starting history
```

wadethestealth:
Manual is fine. I will test lisp on my machine. Also jmp is owned by @n-riesco and enchannel-zmq-backend is owned by nteract. So basically we are just making sure that if we run into issues we have more control over getting fast fixes and publishes as we are pretty much guarenteed to always have some from nteract available rather than relying on just @n-riesco. I am also unsure of the development status of jmp, but I think it is no longer maintained on a serious level.

n-riesco:
For the record, I'm still maintaining jmp. Seriously! :stuck_out_tongue_closed_eyes: 

I'm not very familiar with `enchannel-zmq-backend`, but my understanding is that `enchannel-zmq-backend` is a wrapper on top of `jmp` to access Jupyter messaging socket as RxJS subjects.

Since only `ZMQKernel` has access to these sockets, I don't think the extra level of indirection is necessary.

wadethestealth:
Interesting @n-riesco. Why was this on the Todo then?

n-riesco:
@wadethestealth I think that was one of Will's TODO items. I don't know what he had in mind.

Anyway, please, take my comment above as an opinion. In general, my thoughts are that the number of abstractions/dependencies/... in Hydrogen should be reduced, so that the knowledge-threshold to get contributions is lowered. I also beleive this would reduce the burden that maintaining Hydrogen has become.

aviatesk:
@n-riesco makes sense. Thanks for taking an initiative on this @wadethestealth , but I would like to close this and discard the TODO.

Issue #1854, opened 2019-12-31
Uncaught SyntaxError: Invalid or unexpected token


qwertytam:
### Steps to Reproduce
1. Install Hydrogen from Atom

**Atom**: 1.42.0 x64
**Electron**: 4.2.7
**OS**: Microsoft Windows 10 Home
**Thrown From**: [Hydrogen](https://github.com/nteract/hydrogen) package 2.13.1

### Stack Trace

Uncaught SyntaxError: Invalid or unexpected token

```
At C:\Users\Tom\AppData\Local\atom\app-1.42.0\resources\app\static\<embedded>:11

C:\Users\Tom\.atom\packages\Hydrogen\node_modules\github-markdown-css\github-markdown.css:1
(function (exports, require, module, __filename, __dirname, process, global, Buffer) { return function (exports, require, module, __filename, __dirname) { @font-face {
                                                                                                                                                           ^

SyntaxError: Invalid or unexpected token
    at new Script (vm.js:79:7)
    at NativeCompileCache.runInThisContext (~/AppData/Local/atom/app-1.42.0/resources/app/static/<embedded>:11:146322)
    at Module.get_Module._compile (~/AppData/Local/atom/app-1.42.0/resources/app/static/<embedded>:11:147206)
    at Object.value [as .js] (~/AppData/Local/atom/app-1.42.0/resources/app/static/<embedded>:11:150977)
    at Module.load (internal/modules/cjs/loader.js:602:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:541:12)
    at Function.Module._load (internal/modules/cjs/loader.js:533:3)
    at Module.require (/app.asar/static/index.js:72:46)
    at require (~/AppData/Local/atom/app-1.42.0/resources/app/static/<embedded>:11:146745)
    at /packages/Hydrogen/node_modules/@nteract/markdown/lib/markdown-render.js:18:1)
    at /packages/Hydrogen/node_modules/@nteract/markdown/lib/markdown-render.js:31:3)
    at Module.get_Module._compile (~/AppData/Local/atom/app-1.42.0/resources/app/static/<embedded>:11:147429)
    at Object.value [as .js] (~/AppData/Local/atom/app-1.42.0/resources/app/static/<embedded>:11:150977)
    at Module.load (internal/modules/cjs/loader.js:602:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:541:12)
[...]

Issue #1874, opened 2020-02-07
Can't install Hydrogen? Read this issue first before making a new issue.


wadethestealth:
### Explanation and Background information to all of the most recent install problems

Hydrogen is an open source package that is mostly maintained by myself with occasional check-ins from previous developers. Currently Hydrogen has a few reoccurring install errors that seem to always come up so I am going to document them with explanations here.

- If your error has the name zmq or zeromq in it, this is most likely indication of a bad install.
   - The best steps are for you to install hydrogen via a terminal using atom's `apm install hydrogen --verbose` command. Copy the entire log and if your hydrogen isn't fixed from the install please open a new issue with both the error you are getting and the install log.
- If you are an Archlinux user, unfortunately you are on your own. Both Atom and Hydrogen are not officially supported on Archlinux and variations. However you may still open an issue and other users who see the issue may be able to give you insights on your problem. Just make sure you aren't running to an issue that may be OS independent as that will most likely already have a posted solution/workaround
- If your error contains something about python, you need to make sure that both you global and kernel python paths are pointing to their correct locations.

Now if you are facing an install issue, you are most likely facing a global install issue that affects all users. Unfortunately these have been extremely common recently, and we are working on putting blockage in place to prevent these. The most common issues are caused by Hydrogen using a package built for the nteract desktop app, which builds in a different context. Because of the context difference changes specific to the desktop version can leak into the package without causing test failure as they aren't tested globally (we are working on creating these tests). During 2.9 and later is when hydrogen began to depend on these packages, but v2.8 does not and relies only on packages that are not receiving updates and for this reason it has remained our stable fallback. If you face a new install issue, installing `v2.8.0` should not have the issue and you should make sure that there is currently an issue open for your issue on v2.9 or later (there most likely is as people spam open these).

aviatesk:
Well, Hydrogen@2.14.0 addressed some installation problems, so `apm install Hydrogen` may work out-of-box for now.

Lee1125:
I can't install the hydrogen too.

PR #1876, opened 2020-02-09
Pin @nteract/markdown dependency to 4.2.0

lhm:
This is the last version published before the introduction of 
github-markdown-css which is currently causing problems during install
as reported in #1854

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "@babel/runtime-corejs2": "^7.0.0",
     "@jupyterlab/services": "^0.52.0",
     "@nteract/commutable": "^7.1.4",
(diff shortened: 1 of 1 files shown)


lhm:
This is the commit in upstream that introduced the problem: https://github.com/nteract/nteract/commit/7dc92ca5e529ca71111b51d5e0fe0f81915a63ab#diff-7ee89dfc316cddaf5dfba777aa7dc611

lhm:
I believe this would temporarily fix the installation problem mentioned in #1874 until the underlying issue can be resolved.

aviatesk:
@lhm Thanks.

PR #1879, opened 2020-02-12
show detailed info in status bar

aviatesk:
image

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -63,13 +63,21 @@ const Config = {
       default: false,
       order: 5
     },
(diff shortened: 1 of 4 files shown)


wadethestealth:
I feel like these stats are better for the kernel monitor as many atom packages use the status bar so we probably don't want to consume so much room. But also we should be able to open kernel stats by clicking on the python 3 idle part

aviatesk:
We have a config to control this feature and so an user who feels those info are verbose can disable this.

PR #1907, opened 2020-04-09
rm mobx-react-devtools

aviatesk:
As stated in [mobx-react-devtools's README](https://github.com/mobxjs/mobx-react-devtools), mobx≧6.0 doesn't require this tool actually.
It would be appreciated if you could tell me how to recover the same kind of functionality since I've never used this tool. Well, I think we can remove this tool for now even if we're not sure about the alternative, I dunno.

/cc @BenRussert @lgeiger

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -5,7 +5,7 @@ import {
   CompositeDisposable,
   Disposable,
   Point,
-  TextEditor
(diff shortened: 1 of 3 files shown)


aviatesk:
The updated prettier added some automated formatted lines.

PR #1961, opened 2020-06-17
Add ability to display results from markdown cells in Output dock

MaxMifkovic:
Hi,

This is my first pull request to a GitHub project, please let me know if I should create an issue before diving straight into a PR.

Currently when the output dock is open, code cells display their output in the dock, but the output of markdown cells is shown inline. This PR adds a configuration `showMarkdownInOutputArea` to the settings menu in the Hydrogen package. When activated, outputs from markdown cells will be displayed in the output dock instead of inline. This setting is valuable because:

1. it separates the actual code from the results and prevents markdown output from cluttering the code window;

2. enables users to view their code on one half of the screen and have a living rough draft of what the HTML (or other notebook viewer) view of their final notebook will look like (think something like [Pweave](http://mpastell.com/pweave/) or [VS Code's Python Interactive window](https://code.visualstudio.com/docs/python/jupyter-support-py#_convert-jupyter-notebooks-to-python-code-file)).

This setting only works if the output dock is already open or if the setting "View output in the dock by default" is also enabled. If "View output in the dock by default" is not activated and the output dock is closed, then the markdown cell results are shown with the default inline behavior.

I would have liked to extend this behavior to regular `.py` files (or whichever language is being used), but that should probably be a separate issue/PR because currently no markdown outputs are shown at all in non-`.ipynb` files; only the outputs from code cells are displayed.

By the way, the [Hydrogen contributing page](https://github.com/nteract/hydrogen/blob/master/CONTRIBUTING.md#flow) states that developers should run `npm run flow` before making a PR. I did that and it (or perhaps whatever githooks are enabled) made a bunch of random syntax changes to the two files I made my changes to. I thought about undoing those changes, but ultimately decided to leave them alone.

Please let me know if I should create an issue instead of a PR or if there are any changes I should make.

Thanks,

Max

### DIFF
--- a/lib/result.js
+++ b/lib/result.js
@@ -58,10 +58,13 @@ export function createResult(
   if (code.search(/[\S]/) != -1) {
     switch (cellType) {
       case "markdown":
-        outputStore.appendOutput(convertMarkdownToOutput(code));
-        outputStore.appendOutput({ data: "ok", stream: "status" });
-        if (globalOutputStore)
+        if (
+          globalOutputStore &&
+          atom.config.get("Hydrogen.showMarkdownInOutputArea")
+        )
           globalOutputStore.appendOutput(convertMarkdownToOutput(code));
+        else outputStore.appendOutput(convertMarkdownToOutput(code));
+        outputStore.appendOutput({ data: "ok", stream: "status" });
         break;
       case "codecell":
         kernel.execute(code, (result) => {
--- a/lib/result.js
+++ b/lib/result.js
@@ -58,10 +58,7 @@ export function createResult(
   if (code.search(/[\S]/) != -1) {
     switch (cellType) {
       case "markdown":
-        if (
-          globalOutputStore &&
-          atom.config.get("Hydrogen.showMarkdownInOutputArea")
-        )
+        if (globalOutputStore)
           globalOutputStore.appendOutput(convertMarkdownToOutput(code));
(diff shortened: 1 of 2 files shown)


lib/result.js
```diff
@@ -58,11 +58,16 @@ export function createResult(
   if (code.search(/[\S]/) != -1) {
     switch (cellType) {
       case "markdown":
-        outputStore.appendOutput(convertMarkdownToOutput(code));
+        if (
+          globalOutputStore &&
+          atom.config.get("Hydrogen.showMarkdownInOutputArea")
```
aviatesk, line 63: Why not just watch on `globalOutputStore` ?
Introducing another configuration only for this specific case sounds like too verbose to me.

MaxMifkovic: The only reason I decided to add a new configuration was because I did not want to change the existing behavior of the output dock. But you're right, it is a pretty specific use case and I can remove the `showMarkdownInOutputArea` setting from the configurations if you prefer.

aviatesk: yeah let's remove that, that would make markdown cell rendering consistent with code evaluation, and that would be better, I think.

MaxMifkovic: Done. The changes to `lib/config.js` have been removed and the changes to `lib/result.js` has been updated to only depend on `globalOutputStore`.

aviatesk:
lgtm, thanks for your help again ;)

Issue #1963, opened 2020-06-19
Copy&Paste does not work with inline display of results


PhilippImhof:
### Description

When results are displayed inline, there should be a tooltip indicating _Click to copy_ and _Cmd/Ctrl+Click to open in editor_ I do not get to see this tooltip and thus cannot copy results.

### Steps to Reproduce

1. Create new window and set grammar to Python or Sage. (I do not think it is related to the grammar, though.)
2. Input `2+3` or `print("bla")` and press Cmd+Enter or Ctrl+Enter to execute _Hydrogen: Run_
3. The result is shown in a `<div class="hydrogen_cell_display" ...>` inside a `<div class="inline-container native-key-bindings" ...>`. Move the cursor to that div and hover. Wait for the tooltip to show.

I also tried just clicking the div, but nothing happened either.

### Versions

OS version: macOS Mojave 10.14.6

> `atom --version`
```
Atom    : 1.48.0
Electron: 5.0.13
Chrome  : 73.0.3683.121
Node    : 12.0.0
```

> `apm --version`
```
apm  2.5.0
npm  6.14.5
node 10.20.1 x64
atom 1.48.0
python 3.7.6
git 2.21.1
```

Hydrogen version: 2.14.1

### Plugins

Have you installed and activated any of the Hydrogen plugins below ?
- [x] hydrogen-python 0.0.8
- [x] Hydrogen Launcher 1.2.2
- [ ] Data Explorer

The error also persisted after disabling `hydrogen-python` and `hydrogen-launcher`.

### Logs

No error or warning messages.

### Expected behavior

I would expect the tooltip to appear and the copying to happen upon clicking the div.

### Additional context

I downloaded the code and tried to investigate a bit. Maybe this helps to track down the issue:

- The function `addCopyTooltip` in `lib/components/result-view/result-view.js` is called. Inside that function `atom.tooltips.add` returns a non-empty tooltip. Examining this tooltip, it has Disposal action -> Scopes -> Closures -> Target set to `div.hydrogen_cell_display`, the exact div that it should be attached to. 

- Inside `addCopyTooltip` when `comp.add` is called in order to add the tooltip, the `comp` is correctly initialized (is is the `div` containing the result) and it has its `disposed` property set to `false`. So there is no reason, why the tooltip should not be added.

[...]

PR #1964, opened 2020-06-19
Close issue #1963 (tooltip not shown)

PhilippImhof:
This closes issue #1963 by giving the display div for inline results a bit more height.

### DIFF
--- a/styles/hydrogen.less
+++ b/styles/hydrogen.less
@@ -105,6 +105,7 @@ svg:first-child {
 
 .inline-container .hydrogen_cell_display {
   padding: 0 0.25em 0 0.25em;
+  min-height: 2.5em;
 }
 
 .inline-container.icon::before {


aviatesk:
imho this looks really ugly, no ?
image

PhilippImhof:
I agree that it is not the most beautiful solution, however, I would not call that _really ugly_. Do you have a better idea? (Making it larger did not work for me.) A solution could be to add a small clipboard icon to the right, like you do for multi-line output.

Maybe it would look better, if the content was vertically centered. Further testing showed me that `min-height: 1.8em` would be enough; smaller sizes would not work anymore.

If I have the choice between a very beautiful inline result that cannot be copied and a copyable that is a bit less beautiful, I would definitely choose the latter. I would even prefer a _real ugly_ solution that works to a beautiful one that doesn't.

aviatesk:
The problem is that I can't reproduce your problem and we've not received issues like yours.
So my hunch is that your have some weird stylesheet setup or your theme does that.
I suggest you can keep this change in your `~/.atom/stylesheet.less`, and imho this change is not worth to be upstreamed (i.e. in this repository).

PhilippImhof:
OK,  thanks. I have no custom stylesheet setup and I tried atom-dark and atom-light themes, so both pretty standard. It might be OS specific, though. But if you do not want to or see no need to fix it, I will just change my local stylesheet.

aviatesk:
> I have no custom stylesheet setup and I tried atom-dark and atom-light themes, so both pretty standard. It might be OS specific, though.

ah, really ? Which OS, if you wouldn't mind ? 
I agree with that we don't have such a good stylesheet that looks all fine on all the environment (I personally have some hydrogen-specific setups in my stylesheet as well), but imho forcing the look above for every user would be a very big deal.

PhilippImhof:
Ah sorry, I thought you saw that on the bug report. I'm on macOS 10.14.6. 

I see your point that having the output look worse might not be acceptable to users who do not have the problem. I will try a solution with a small clipboard icon on the right side, keeping the regular height.

PR #1966, opened 2020-06-20
Close issue #1963 (tooltip not shown, content not copyable on mac OS)

PhilippImhof:
Closing issue #1963.

Apparently the problem was that atom's `<div>` containing the editor's line of text swallowed the click event that should have been captured by the output cell. This is why a larger output sell solved the problem: one was now able to click below the editor line.

Setting `position: relative;` for the hydrogen marker solves the issue, because now the newly inserted `<div>` lies above the editor line and becomes clickable. No visible changes.

### DIFF
--- a/styles/hydrogen.less
+++ b/styles/hydrogen.less
@@ -17,6 +17,7 @@
 
 .hydrogen.marker {
   display: flex;
+  position: relative;
 }
 
 .hydrogen {


aviatesk:
thanks !

PhilippImhof:
Thanks to you, mainly for developing hydrogen, but also for including this so quickly. This might probably be the smallest PR in GitHub's history....

PR #2075, opened 2021-04-13
fix: update dependencies: react, atom-select-list, etc - remove unused imports - use native instead of lodash

aminya:
- Upgrades the react-related dependencies.
- Updated atom-select-list and fixed its types
- Removed unused imports
- Replaced lodash with native functions where possible

I didn't upgrade:
- mobx -> which has breaking changes -> should be done in a separate PR
- react-table -> which has broken the API significantly -> switch to solid-simple-table?


aminya (commented):
Everything seems to work. Verified using the unit tests and the files in the examples.

PR #2106, opened 2021-05-25
Bugfix ratio distortion of media in sidebar

### LINKED ISSUE #1443: Hydrogen Output Area Scalable Figure Issue
**Description:**

Hi there,

I'm a new user to Atom and  generally new to the coding .

I've been using Jupyter Notebook but I do not like their layout and I am really glad I discovered the Hydrogen package for Atom.

I have been having one issue where my matplotlib figure outputs to the "Hydrogen Output Area" seems to have a limitation in size and a locked ratio. I was wondering if there was a way I can have the figure output to behave just like how the figures would if it were to be outputted with the `%inline qt` command. In other words, a way that I can increase or decrease the size of the figure without a size and ratio constraint. I was able to have the effect without the constraints on the "Hydrogen Output Area" the first time I plotted something. I'm not sure what I did but it switched over the next time I plotted the same code. I hope I am explaining this clearly enough, I'm not too familiar with the terminology.

**Steps to Reproduce:**

This is the code I am running to get the figure.

`%matplotlib inline`
`import numpy as np`
`import pandas as pd`
`import matplotlib.pyplot as plt`

`train = pd.read_csv('train.csv')`
`missing = train.isnull().sum()`
`missing = missing[missing > 0]`
`missing.sort_values(inplace=True)`
`missing.plot.bar()`

**Versions:**
OS: macOS Mojave
Atom: 1.31.2 x64
Hydrogen: 2.6.0

Output

Thank you for your time and your effort!

alexisthual:
This PR closes #1443.
In short, it makes sure that media (`img, svg, video`) displayed in the Hydrogen Output Area keep their initial ratio instead of being distorted to fill the whole space.

Before:
Screenshot from 2021-05-25 10-26-36

After:
Screenshot from 2021-05-25 10-26-21

### DIFF
--- a/styles/hydrogen.less
+++ b/styles/hydrogen.less
@@ -248,9 +248,11 @@
     video,
     img,
     svg {
-      width: unset !important;
-      height: unset !important;
+      object-fit: contain !important;
+      width: auto !important;
+      height: auto !important;
       max-width: 100% !important;
+      max-height: 100% !important;
     }
   }
 


aminya (approved):
Thanks

PR #2108, opened 2021-05-27
Fix altair plots

aminya:


lib/components/result-view/display.tsx
```diff
@@ -21,7 +21,7 @@ import {
   Vega4,
   Vega5,
 } from "@nteract/transform-vega";
-import Markdown from "./markdown";
+import { Markdown } from "./markdown";
 // All supported media types for output go here
 export const supportedMediaTypes = (
   <RichMedia>
```
aminya, line 27: As I am unfamiliar with @nteract/transform-vega and @nteract/outputs, it appears that *examples/vega_datasets-altair.py* does not plot anything because of this particular part of the code.

cc: @captainsafia

PR #2117, opened 2021-06-01
fix: update @jupyterlab/services

aminya:


aminya (commented):
Remote kernel doesn't work after this

PR #2138, opened 2021-11-21
Remove duplicate keybindings

wjandrea:
`ctrl-enter` is duplicated in the same schema.
`shift-enter` is duplicated in Darwin.
`ctrl-shift-backspace` is duplicated in Windows/Linux editor.

I'm not an expert on Coffeescript or Atom config, so if there's a valid reason for the duplicates to exist, please let me know.


wjandrea:
cleaning up now that Atom is sunsetted

PR #2162, opened 2023-01-09
Fix bug while checking for an old version

mauricioszabo:
We're keeping a fork of Atom called Pulsar. We decided that our "first beta version" was going to be `1.100`, but the way Hydrogen checks for versions is using `parseFloat`, which translates `1.100` to `1.1` and then it proceeds to use some outdated API.

This PR removes the check (Atom 1.2 is _really old_ and considering that the project is dead anyway...)

### DIFF
--- a/lib/utils.tsx
+++ b/lib/utils.tsx
@@ -218,16 +218,12 @@ export function rowRangeForCodeFoldAtBufferRow(
   editor: TextEditor,
   row: number
 ) {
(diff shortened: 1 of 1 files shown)


LRydin:
I second this. I am facing an issue due to this.

shr00mie:
> I second this. I am facing an issue due to this.

i fif this.

aminya (approved):

PR #2180, opened 2026-04-02
Security: Untrusted kernel output can execute arbitrary JavaScript in renderer

tuanaiseo:
## Problem

The output renderer explicitly enables `Media.JavaScript` and `Media.HTML` for all kernel outputs. Because kernel responses are treated as trusted, a malicious or compromised kernel (or remote session) can send crafted output that runs script in Atom's renderer context, potentially leading to code execution or credential theft in an Electron app.

**Severity**: `critical`
**File**: `lib/components/result-view/display.tsx`

## Solution

Disable `Media.JavaScript` by default and gate active content behind an explicit trust model (per-kernel/per-document trust prompt). Sanitize HTML with a strict allowlist sanitizer (for example DOMPurify with hardened config), and render untrusted rich outputs in a sandboxed iframe without Node/Electron privileges.

## Changes

- `lib/components/result-view/display.tsx` (modified)
- `lib/components/inspector.tsx` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/lib/components/inspector.tsx
+++ b/lib/components/inspector.tsx
@@ -21,11 +21,7 @@ const Inspector = observer(({ store: { kernel } }: Props) => {
   }
   const bundle = kernel.inspector.bundle;
 
-  if (
-    !bundle["text/html"] &&
-    !bundle["text/markdown"] &&
-    !bundle["text/plain"]
-  ) {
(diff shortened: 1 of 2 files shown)


aminya (changes requested):
The fix isn't to removes features that the users relied on.

PR #2181, opened 2026-04-02
Security: Inspector renders unsanitized HTML/markdown from kernel introspection data

tuanaiseo:
## Problem

Inspector view renders `bundle` content from kernel inspection using `RichMedia` with `Media.HTML` and markdown output components. If the inspected source or kernel response is attacker-controlled, this can inject active HTML/script into the UI.

**Severity**: `high`
**File**: `lib/components/inspector.tsx`

## Solution

Treat inspection bundles as untrusted: sanitize `text/html`, disable script-capable rendering for untrusted kernels, and require explicit trust elevation before enabling rich HTML output.

## Changes

- `lib/components/inspector.tsx` (modified)
- `lib/components/result-view/display.tsx` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/lib/components/inspector.tsx
+++ b/lib/components/inspector.tsx
@@ -3,7 +3,6 @@ import { observer } from "mobx-react";
 import { RichMedia, Media } from "@nteract/outputs";
 import { INSPECTOR_URI } from "../utils";
 import type Kernel from "../kernel";
-import Markdown from "./result-view/markdown";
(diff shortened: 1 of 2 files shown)


aminya:
The fix isn't to removes features that the users relied on.

